PR #10, opened 2023-11-14
docs: remove non existing utility classes from documentation

ninamarina:
## Description
Jira ticket: https://dialpad.atlassian.net/browse/DLT-1368

Affects the utility classes docs that where using `apps/dialtone-documentation/docs/_data/spacing.json` to obtain the values, these are: `margin`, `padding`, `auto-spacing` and `gap`. 
`margin` and `padding` have the same values, but `gap` and `auto-spacing` have other ones.
Check that the shown values are the same as the generated ones in `dialtone.css`

## Pull Request Checklist

image

### DIFF
--- a/apps/dialtone-documentation/docs/_data/auto-spacing.json
+++ b/apps/dialtone-documentation/docs/_data/auto-spacing.json
@@ -1,0 +1,76 @@
+{
+  "values": [
+    {
+      "value": "0",
+      "output": "0"
+    },
+    {
+      "value": "1",
+      "output": ".1rem"
+    },
+    {
+      "value": "2",
+      "output": ".2rem"
+    },
+    {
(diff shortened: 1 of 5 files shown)


braddialpad (approved):
This change looks good, my only question is are we sure there are not more of these missing / additional utility classes in the docs in other places?

ninamarina:
> This change looks good, my only question is are we sure there are not more of these missing / additional utility classes in the docs in other places?

I honestly didn't check, I only checked for the ones that used `spacing.json`. Let me see if a can find a method to review that

ninamarina:
We are merging this as is for now since checking every utility class is a big effort. We might get more reports in the future, and maybe we can think of a better way to generate the docs.

PR #36, opened 2023-12-05
refactor(emoji-picker): code readability

iropolo:
## Description

Small PR for emoji picker component with code readability improvements.

I moved some stuff regarding keyboard navigation to make the `emoji_selector.vue` component more readable.
I also clean a little bit the `useKeyboardNavigation.js` composable to be more effective on what should do.
There is still things to improve here, but this one was sth I was left and wanted to do.

Component should work the same.


braddialpad (approved):
Because this is all composables related, I suspect we don't need to make changes to vue 2 right?

iropolo:
> Because this is all composables related, I suspect we don't need to make changes to vue 2 right?

ye.. I was thinking to update some names but... I didnt see it necessary...

PR #37, opened 2023-12-06
feat(feed-item-row): add gap instead of padding and label field

braddialpad:
## Description
Found a case where the feed items have an "external" label that wasn't represented in the component, so added support for this case and changed spacing to use gap instead of padding so it does not take up extra space if either of the threading / emoji slots don't exist. I think there is some re-thinking that needs to be done with the item layout and list item components, and also the use of dt-stack, but the scope of that is bigger than this PR, to be discussed at next grooming.

## Pull Request Checklist


iropolo:
Re-running preview due failed action. Now seems to work fine

iropolo (approved):
Looks ok, and classes seems to be working.

Thanks,

PR #41, opened 2023-12-08
feat: dialtone-vue sync script

juliodialpad:
## Description

- Added filter to run unit tests only on dialtone-vue(2/3) changes
- Added dialtone-vue2 to a11y tests filter
- Added dialtone-vue sync script, it check the diff between commits, filters the changes by path and applies the changes to the other dialtone-vue version (depending on what you selected)
- Did it in a interactive way so it should be easier to use.
- We'll need to use it and check that everything is working fine, it's a bit hard to simulate conflicts or changes that might break it.
- Please let me know if something is not working so I can fix it.

## Pull Request Checklist

 - [ ] Ensure all `gulp` scripts successfully compile.
 - [ ] Update, remove, or extend all affected documentation.
 - [ ] Ensure no private Dialpad links or info are in the code or pull request description (Dialtone is a public repo!).


scripts/dialtone-vue-sync.sh
```diff
@@ -0,0 +1,65 @@
...
+    echo "Applying patch"
+    echo "$detected_diff" | git apply -p3 --directory packages/"$dest_version" --3way;
+  else
+    echo "No changes found on packages/$src_version from $3 -> $4"
+  fi
+}
+
+echo "Welcome to dialtone-vue sync script."
+echo "This script will copy your changes from one dialtone-vue version to the other."
+echo "By default the script will copy over the changes from the last commit, if you want to copy changes between"
```
braddialpad, line 39: Could we possibly have the default be to copy over all the commits unique to this branch when compared with staging? Maybe we don't have to do this now, but could be a future update.

braddialpad (approved):

scripts/dialtone-vue-sync.sh
juliodialpad: That sounds fair, the only thing I'm afraid of is that it could possibly create more merge conflicts, don't you think?

braddialpad: it's possible yeah, but also the same way our previous script worked.

juliodialpad: Cool, let's test it and we can update it if needed.

PR #43, opened 2023-12-11
chore(emoji-picker): keyboard navigation updates

iropolo:
## Description
Few updates to emoji picker keyboard navigation:

Fix:

- The emojis are getting hidden behind the category header

Changes:

- From emoji skin selector with `shift+tab` key would jump to previous focus. (first emoji of last label on emoji selector).
- From tabset with `shift+tab` key would jump to emoji skin selector

Modified packages:

- Dialtone
- Dialtone-vue2
- Dialtone-vue3


braddialpad (approved):
Seems to work as described.

PR #53, opened 2023-12-13
refactor: dialtone tokens release process

juliodialpad:
## Description

- Updated dialtone-tokens release process to release ios and android packages
- Changed to affected instead of run-many to improve the performance on local release.
- Cleaned up packages/dialtone-tokens/.github folder

## Pull Request Checklist


.github/workflows/release.yml
```diff
@@ -13,37 +13,55 @@ env:
...
-      - uses: actions/checkout@v3
-      - uses: pnpm/action-setup@v2
+      # Will prevent the rest of the steps from running on fail
+      - name: Check if user is a dialpad member
+        uses: octokit/request-action@v2.x
         with:
-          version: 8
+          route: GET /orgs/dialpad/members/${{ github.actor }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.DIALTONE_CI_TOKEN }}
```
braddialpad, line 33: Don't think we actually need this anymore right? These should be limited to only run by those with admin / maintainer permissions

.github/workflows/release.yml
```diff
@@ -13,37 +13,55 @@ env:
...
-          route: GET /orgs/dialpad/members/${{ github.actor }}
-        env:
-          GITHUB_TOKEN: ${{ secrets.DIALTONE_CI_TOKEN }}
+          path: ${{ env.STORE_PATH }}
+          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-pnpm-store-
 
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
```
braddialpad, line 67: Do we want to skip the install if we got a cache hit? or it doesn't matter that much?

.github/workflows/release.yml
```diff
@@ -52,30 +70,75 @@ jobs:
...
 
-      - name: Set registry to github packages
+      - name: Set registry to GitHub packages
         run: npm config set @dialpad:registry https://npm.pkg.github.com
 
-      - name: Set github auth token
+      - name: Set GitHub auth token
         run: npm config set //npm.pkg.github.com/:_authToken=${{ secrets.GITHUB_TOKEN }}
 
       # note that pnpm run build is executed before publishing via prepublishOnly
```
braddialpad, line 87: Hmm seems there is a `prepublishOnly: "pnpm run build"` in each individual package. Does this still make sense? For one doesn't this run the entire `pnpm run build` for each package? Should it be moved to the root package, and then also have a single command to publish in the root package rather than using run-many? If not, should we be running the build for each package individually instead of the whole build?

.github/workflows/release.yml
juliodialpad: Really, we're we set that? I don't remember setting that up.

.github/workflows/release.yml
juliodialpad: Doesn't matter much, without cache pnpm takes like 2 minutes and with cache 8 seconds. 
That's the recommended implementation from the pnpm documentation.

.github/workflows/release.yml
braddialpad: Hmm I think it's maybe just running the local release command that doesn't let you do it if you're not admin. Anyways don't worry about it for now I will look into it some more.

.github/workflows/release.yml
braddialpad: k sounds good.

.github/workflows/release.yml
juliodialpad: It makes sense to run the build before publishing but we might want to move that config to a more centralized config, I'll move it to `nx.json` on targetDefaults to let NX now that release depends on building dependencies so it can optimize it.

We can do it in two ways:
- Using NX run-many 
- Using pnpm -r publish

Using NX run-many should be the way to go as NX would use cache and optimize the build and release order.

braddialpad: Sure the nx run-many sounds like the way to go then.

I find it a bit confusing when we should use nx commands vs pnpm commands.

juliodialpad: Yes, it's a bit confusing, I'm reading more about to know exactly when we need to use pnpm, but NX should be the way to go in most cases AFAIK for now.

braddialpad (approved):

PR #58, opened 2023-12-14
feat(hovercard): hovercard stub

iropolo:
## Description
Hovercard stub [DLT-1321](https://dialpad.atlassian.net/browse/DLT-1321)

Stub made on vue3 version, will added it to vue2 after this PR (since is the same code)

Component not added on 
packages/dialtone-vue3/common/components_list.cjs 
or
packages/dialtone-vue2/common/components_list.cjs

to avoid public hovercard stub

----

As you can see I modified a little bit popover to handle hovercard functionality.

----

The idea of this stub is to check the initial structure of hovercard component.
All other stuff (test, doc, design) will be added on implementation task [DLT-616](https://dialpad.atlassian.net/browse/DLT-616)

[DLT-616]: https://dialpad.atlassian.net/browse/DLT-616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

----

Some description:

Component has this props shared through popover:

    id
    placement
    content-class
    fallback-placements
    padding
    transition
    offset
    header-class
    footer-class
    append-to

Popover now will have a new `hovercard` prop, to provide correct functionality to hovercard component.

Slots are the sames as popover provide.

[DLT-1321]: https://dialpad.atlassian.net/browse/DLT-1321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dialtone-vue3/components/hovercard/hovercard_pop.vue
ninamarina, line 1: This file is not used, is it?

ninamarina:
The behavior looks good! The only thing is that it also opens on click, and I don't think that's what we want, because the anchor might have their own onClick interaction.

packages/dialtone-vue3/components/hovercard/hovercard_pop.vue
iropolo: huh, ye, it is a leftover, removed, thanks

iropolo:
> The behavior looks good! The only thing is that it also opens on click, and I don't think that's what we want, because the anchor might have their own onClick interaction.

hmmm yes, should not have that behaviour.. will check it out, I will probably disable it if the `hovercard` prop is true

francisrupert (changes requested):
Loving it, so excited for this one. A few interaction nuances:

* `placement`: default to `top-start`
* `offset`: default to `[ var(--dt-size-0), var(--dt-size-500) ]`
  * if can't use CSS vars, then `0, 16`
* `transition`: boolean, default `false`. 
  * Easing: `transition-timing-function: var(--ttf-out-quint)` and `transition-duration: var(--td200)`
* `placement` and `padding`: can this be a list of options, such that it appears as a `<select>` in Storybook? as [Popover](https://dialtone.dialpad.com/vue/index.html?path=/story/components-popover--default) does.

packages/dialtone-vue3/components/popover/popover.vue
```diff
@@ -953,6 +968,48 @@ export default {
...
+
+    //  ============================================================================
+    //  $ HOVERCARD
+    //  ----------------------------------------------------------------------------
+
+    setInTimer () {
+      this.inTimer = setTimeout(() => {
+        this.isOpen = true;
+      }, TOOLTIP_DELAY_MS);
+    },
```
braddialpad, line 982: Should we also only set timers if hovercard is true like the other functions? Currently on a popover with `hovercard = false` it will set a timer that never ends because all of the clearTimeout functions have `if (!this.hovercard) return;`

packages/dialtone-vue3/components/popover/popover.vue
```diff
@@ -507,6 +512,14 @@ export default {
             (appendTo instanceof HTMLElement);
       },
     },
+
+    /**
+     * Set this prop to true and popover component will support hovercard behaviour
```
braddialpad, line 517: explain specifically what "hovercard behaviour" means. In this case it causes it to trigger mouseenter/mouseleave events on a timer delay of 300ms.

packages/dialtone-vue3/components/popover/popover.vue
```diff
@@ -953,6 +968,48 @@ export default {
...
+    },
+
+    onLeaveContent (e) {
+      if (!this.hovercard) return;
+      this.setOutTimer(e);
+    },
+
+    //  ============================================================================
+    //  $ HOVERCARD
+    //  ----------------------------------------------------------------------------
```
braddialpad, line 1014: extra?

packages/dialtone-vue3/components/popover/popover.vue
```diff
@@ -953,6 +968,48 @@ export default {
...
+
+    onEnterContent () {
+      if (!this.hovercard) return;
+      clearTimeout(this.outTimer);
+    },
+
+    onLeaveContent (e) {
+      if (!this.hovercard) return;
+      this.setOutTimer(e);
+    },
```
braddialpad, line 1010: Does it make sense to do this timing logic here? Or does it make more sense for this popover to simply emit a "mouseover", "mouseleave" event and all the timing stuff is handled within the hovercard component?

If this were the case we could probably remove the "hovercard" prop from this component which doesn't quite feel right.

packages/dialtone-vue3/components/popover/popover.vue
iropolo: Hmm... currently on popover with hovercard = false it will not set any timer, because those functions are called when hovercard is true.
When is false `inTimer: null,` and `outTimer: null,`.

Where `it will set a timer that never ends` ?

packages/dialtone-vue3/components/popover/popover.vue
iropolo: K, updated to:
```
/**
     * Set this prop to true and popover component will support hovercard behaviour
     * It will open on mouseenter and close on mouseleave with timer delay of 300ms
     */
```

packages/dialtone-vue3/components/popover/popover.vue
braddialpad: oh my bad I thought this was a event handler lol. it's all good.

packages/dialtone-vue3/components/popover/popover.vue
iropolo: Since this block is exclusively for hovercard behaviour I decided to use that comment section.
If this component was created in vue3 syntax, I would like to move it out to a composable like emoji picker modules.

I would like to keep that comment.

braddialpad: k, wasn't sure if it was intended, sounds good.

packages/dialtone-vue3/components/popover/popover.vue
iropolo: Yes, it can be done on hovercard side if we use "mouseover", "mouseleave" events.

But,
Since we are reusing popover component to hovercard I would prefer to have the logic in there.
Then hovercard component does not have any logic, it is just a "popover-like variant", thats the idea to have the `hovercard` prop on popover.

Regarding emitting events, if you are not using hovercard on popover, you will emit useless events to the consumer.

braddialpad: Yeah it's a bit of a tradeoff, fine to do it this way I think.

braddialpad (approved):

iropolo:
> The behavior looks good! The only thing is that it also opens on click, and I don't think that's what we want, because the anchor might have their own onClick interaction.

Added 
`if (this.hovercard) { return; }`
on
`defaultToggleOpen` method 
to avoid `open` on click.

iropolo:
> Loving it, so excited for this one. A few interaction nuances:
> 
> * `placement`: default to `top-start`
> * `offset`: default to `[ var(--dt-size-0), var(--dt-size-500) ]`
>   
>   * if can't use CSS vars, then `0, 16`
> * `transition`: boolean, default `false`.
>   
>   * Easing: `transition-timing-function: var(--ttf-out-quint)` and `transition-duration: var(--td200)`
> * `placement` and `padding`: can this be a list of options, such that it appears as a `<select>` in Storybook? as [Popover](https://dialtone.dialpad.com/vue/index.html?path=/story/components-popover--default) does.

Updated: `placement`: default to `top-start`
Updated: `offset`: default to `[0, 16]`
Updated: `transition`: boolean, default false , now hovercard will only have `fade` transition.
Updated: `placement` and `padding` to be a select.


Could you explain a little bit this please
`Easing: transition-timing-function: var(--ttf-out-quint) and transition-duration: var(--td200)`

francisrupert:
@iropolo I guess the [deploy preview](https://dialtone.dialpad.com/vue3/deploy-previews/pr-58/) isn't up.

> Could you explain a little bit this please...

Never mind on that, I assume it's using the Popover's `fade` values, so let's stick to that. Later I may reconsider what the timing and easing of Popover should be, and that's part of a future effort to better formalize our motion-oriented design language.

braddialpad:
hmmm

> Error: google-github-actions/upload-cloud-storage failed with: the offset is lower than the number of bytes written. The server has 0 bytes and while 10 bytes has been uploaded - thus 10 bytes are missing. Stopping as this could result in data loss. Initiate a new upload to continue.

@juliodialpad you seen that before?

juliodialpad:
> hmmm
> 
> 
> 
> > Error: google-github-actions/upload-cloud-storage failed with: the offset is lower than the number of bytes written. The server has 0 bytes and while 10 bytes has been uploaded - thus 10 bytes are missing. Stopping as this could result in data loss. Initiate a new upload to continue.
> 
> 
> 
> @juliodialpad you seen that before?

No, never before, could it be related with the permissions stuff that Nick sent yesterday? 🤔

iropolo:
PR ready for merge, I will re-trigger deploy preview to check if it is working correctly

iropolo:
Deploy preview action works as expected, the previous error seems to be sth flaky not related to the PR

francisrupert (approved):
I'm still seeing `placement` default as `bottom-end`, instead of the intended `top-start`.

iropolo:
> top-start

huh, I forget to update it in storybook, I updated just component 👍🏼

iropolo:
Okay now placement looks good in storybook 

image

PR #59, opened 2023-12-14
chore: various vscode workspace related fixes

braddialpad:
## Description
Fixed issues with the vscode workspace configuration

- Wasn't running eslint on dot folders (.vuepress) since changing to the monorepo.
- Added some additional useful options that should be set by default.
- Configured vitest workspace, this is required for the vitest vscode extension to run on monorepo.
- Added extension recommmendations, should assist with onboarding.
- We had two separate markdownlint configurations. Consolidated to one.
- vitest updated to 1.0! (just released last week)

## Pull Request Checklist


dialtone.code-workspace
```diff
@@ -1,73 +1,87 @@
...
     },
 
-		// SEARCH
+    // SEARCH
     // Note that gitignored files are automatically excluded from search. Do not add files
     // here that are already gitignored.
     "search.exclude": {
-      "**/package-lock.json": true,
-      "**/locale": true
+      "**/package-lock.json": true
```
juliodialpad: package-lock.json files shouldn't longer exists on the monorepo as we're using pnpm

packages/dialtone-vue2/package.json
```diff
@@ -92,7 +92,7 @@
     "typedoc-plugin-markdown": "^3.16.0",
     "vite": "^5.0.0",
     "vite-plugin-vue2-svg": "^0.4.0",
-    "vitest": "^0.34.5",
+    "vitest": "^1.0.4",
```
juliodialpad: You can remove the dependency here if it's already on root's package.json

packages/dialtone-vue3/package.json
```diff
@@ -91,7 +91,7 @@
     "typedoc-plugin-markdown": "^3.16.0",
     "vite": "^5.0.0",
     "vite-svg-loader": "^4.0.0",
-    "vitest": "0.33.0",
+    "vitest": "1.0.4",
```
juliodialpad: Same here

package.json
```diff
@@ -11,7 +11,8 @@
     "start:dialtone-vue3": "nx run-many --target=start --projects=dialtone,dialtone-vue3",
     "start:dialtone-vue2": "nx run-many --target=start --projects=dialtone,dialtone-vue2",
     "prepare": "husky install",
-    "release": "./scripts/release.sh"
+    "release": "./scripts/release.sh",
+    "test:vitest": "vitest"
```
juliodialpad, line 15: Nice, it's going to be easier to maintain just one instance of vitest

juliodialpad (changes requested):
Nice work! Couple changes

dialtone.code-workspace
braddialpad: right, I will ignore the pnpm lock instead

packages/dialtone-vue2/package.json
braddialpad: Yeah I was wondering how that worked... do all sub packages inherit the parent packages?

braddialpad:
Fixes are in

iropolo:
Would we cool to have @ninamarina review here since she is using vscode and this would be helpful.

iropolo (approved):

packages/dialtone-vue2/package.json
juliodialpad: I think it's not inheriting, as you're running `vitest` from the root package, the other packages don't really need to have that dependency.

PR #71, opened 2023-12-22
fix: vue2 storybook components and recipe

iropolo:
## Description
Fixed all code previews of components and recipes on storybook doc.

Removed `createTemplateFromVueFile` and created `createRenderConfig` function on `storybook_utils.js` file

Before:
image

After:
image

### DIFF
--- a/packages/dialtone-vue2/common/storybook_utils.js
+++ b/packages/dialtone-vue2/common/storybook_utils.js
@@ -5,18 +5,21 @@ import iconNames from '@dialpad/dialtone-icons/dist/icons.json';
  * This is useful for more complex components that are hard to work with and
  * get messy when rendered via a template string. Will pass args and argTypes
  * into the component so it is able to be live edited with storybook controls addon.
- * @param {object} args storybook control args
- * @param {object} argTypes storybook control argument settings
- * @param {component} templateComponent vue component template for rendering to storybook.
- *                                      Note this should not be the component itself,
- *                                      but rather the usage of that component.
+ * @param component this will get the component name
+ * @param defaultTemplate we will mount in this component
+ * @param argsData storybook control args
  * @returns {component} the template component with props and args added.
  */
-export const createTemplateFromVueFile = (args, argTypes, templateComponent) => ({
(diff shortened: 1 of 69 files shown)


packages/dialtone-vue2/common/storybook_utils.js
```diff
@@ -5,18 +5,21 @@ import iconNames from '@dialpad/dialtone-icons/dist/icons.json';
...
+
+export function createRenderConfig (component, defaultTemplate, argsData) {
+  return {
+    components: { [component.name]: defaultTemplate },
+    setup () {
+      return { argsData };
+    },
+    template: `<${component.name} v-bind="argsData"></${component.name}>`,
+  };
+}
```
juliodialpad, line 22: If we're just using the name of the component, it's not easier and clearer to pass the name instead of the component?

juliodialpad (commented):
Are you going to open another PR with the Vue 3 changes?

juliodialpad:
I'm pretty sure we want to visual test this, basically you changed all the stories so we want to make sure that everything looks ok.

iropolo:
> Are you going to open another PR with the Vue 3 changes? 

yes, working on it because it will be slightly different.. since `v-bind` is not working properly...

image


🫠🫠🫠🫠🫠

packages/dialtone-vue2/common/storybook_utils.js
iropolo: As how we are using the function e.g:
`render: (argsData) => createRenderConfig(DtBanner, BannerDefault, argsData),`

I see it better than manually pass the name of the component, kind of unstable and sensitive to typo.

iropolo:
Visual tests looks ok, should be ok to go with this to fix in vue2

packages/dialtone-vue2/common/storybook_utils.js
juliodialpad: makes sense, thanks for the clarification

juliodialpad (approved):
Vue 2 changes look good, thanks!

PR #72, opened 2023-12-22
feat(hovercard): add hovercard implementation

ninamarina:
# feat(hovercard): add hovercard implementation

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DLT-616

## :book: Description
Adds story following figma example.
Implements delay for when multiple hovercards are present. To do this, a singleton is used to create a global timer that keeps track of the id of the current hovercard being displayed, and changes it according to the logic that was defined in https://dialpad.atlassian.net/browse/DLT-1144
Adds tests.

## :pencil: Checklist

For all PRs:

For all Vue changes:

For all CSS changes:

If new component:

- I am exporting any new components or constants:
  - [x] from the index.js in the component directory.
  - [x] from the index.js in the root (either `packages/dialtone-vue2` or `packages/dialtone-vue3`).
- [ ] I have added the styles for the new component to the `packages/dialtone-css` package.
- [ ] I have created a page for the new component on the documentation site in `apps/dialtone-documentation`.
- [x] I have added the new component to `common/components_list.cjs` in `packages/dialtone-vue2` and `packages/dialtone-vue3`
- [x] I have created a component story in storybook
- [x] I have created story / stories for any relevant component variants in storybook
- [x] I have created a docs page for the component in storybook.
- [x] I have checked that changing all props/slots via the UI in storybook works as expected.

hovercards


packages/dialtone-vue2/components/hovercard/hovercard.stories.js
```diff
@@ -0,0 +1,159 @@
...
+  Elementum fames nullam elementum velit proin vitae aliquet.
+  Platea nulla consectetur consequat sagittis nullam et ultricies nisl rhoncus
+  aliquet elementum venenatis quisque.</p>`,
+  },
+  {
+    name: 'Purdie Afra', time: '5:16 PM', default: `<p class="d-fs-200 d-lh-400 d-wmx90ch">
+  Elementum fames nullam elementum velit proin vitae aliquet.
+  Platea nulla consectetur consequat sagittis nullam et ultricies nisl rhoncus
+  aliquet elementum venenatis quisque.</p>`,
+  }],
```
braddialpad: Because this data is not part of the component interface itself we don't want it to show up in the storybook controls. We can either manually hide it from the controls via argTypesData, or rather than passing it through args put it directly in the `.story.vue`. (or in a constants file if used in multiple places).

packages/dialtone-vue3/components/popover/popover.test.js
```diff
@@ -92,7 +92,7 @@ describe.skip('DtPopover Tests', () => {
         expect(popoverWindow.exists()).toBe(true);
       });
 
-      it('should render the main content', () => {
+      it.skip('should render the main content', () => {
```
braddialpad, line 95: I think perhaps the same problem you're having with your hovercard tests, is the same reason these tests are skipped.

packages/dialtone-vue3/components/popover/popover.vue
```diff
@@ -521,6 +522,29 @@ export default {
...
+    },
+
+    /**
+     * A prefix for the data-qa tags in the elements inside the popover.
+     * If null, `dt-popover` will be used.
+     */
+    qa: {
+      type: String,
+      default: null,
+    },
```
braddialpad: Can we just pass a standard `data-qa="value"` from the hovercard and then do something like 
```
:data-qa="$attrs['data-qa'] ? `${ $attrs['data-qa'] }-content` : 'dt-popover-content'"
``` 
and so on? The prop feels a little awkward to me but I'm not totally against it if you  want to do it that way.

braddialpad (commented):
Couple quick comments, will look more thoroughly tomorrow.

Examples look great! Delay feels great IMO.

packages/dialtone-vue3/components/popover/popover.vue
ninamarina: I tried that, but the attrs where not passed to the child component. Will investigate more if that can be possible.

packages/dialtone-vue3/components/popover/popover.test.js
ninamarina: Might be, although in vue 2 they are not skipped. Might have something to do with the fact that the hovercard adds another wrapper to all this?

packages/dialtone-vue3/components/popover/popover.vue
ninamarina: That was fixed in the latest changes :)

braddialpad:
Fixes look good thanks! All that's left are these tests...

packages/dialtone-vue2/components/hovercard/hovercard.mdx
```diff
@@ -0,0 +1,51 @@
...
+<Canvas>
+  <Story of={HovercardStories.Default} />
+</Canvas>
+
+## When there are many hovercards
+
+Initially, there is a 300ms delay when numerous hovercards are present.
+However, when transitioning from one anchor to another, the delay is determined by the mouse movement time.
+This implies that the closer the anchors are, the faster the hovercard appears.
+Move the mouse over the avatars to reveal the hovercards in the example.
```
iropolo: Should we explain here that is dependant of a global timer that need to be set it up?

packages/dialtone-vue2/components/hovercard/hovercard.mdx
```diff
@@ -0,0 +1,51 @@
...
+</Canvas>
+
+## Slots, Props & Events
+
+<Controls />
+
+## Accessibility
+
+
+## Usage
```
iropolo: Delete empty sections or add it information

packages/dialtone-vue2/components/hovercard/hovercard.test.js
```diff
@@ -0,0 +1,147 @@
...
+
+const baseScopedSlots = {
+  anchor: '<template #anchor="{ attrs }">' +
+          '<button data-qa="dt-button" v-bind="attrs">Hover me</button>' +
+          '</template>',
+};
+
+let mockProps = {};
+let mockAttrs = {};
+let mockSlots = {};
```
iropolo: these are not being used, it could be removed, up to you

packages/dialtone-vue2/components/hovercard/hovercard.vue
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+   * The direction the popover displays relative to the anchor.
+   * <a
+   *   class="d-link"
+   *   href="https://atomiks.github.io/tippyjs/v6/all-props/#placement"
+   *   target="_blank"
+   * >
+   *   Tippy.js docs
+   * </a>
```
iropolo: This is a styled doc for our storybook, I will suggest to move all this links to the `.mdx` file and have here the short description for the component.

packages/dialtone-vue2/components/hovercard/hovercard.vue
```diff
@@ -0,0 +1,201 @@
...
+   * @values top, top-start, top-end,
+   * right, right-start, right-end,
+   * left, left-start, left-end,
+   * bottom, bottom-start, bottom-end,
+   * auto, auto-start, auto-end
+   */
+    placement: {
+      type: String,
+      default: 'top-start',
+    },
```
iropolo: Since we have validator on padding prop, we could have a validator here, up to you

packages/dialtone-vue2/components/hovercard/hovercard_constants.js
iropolo, line 1: If it is not necessary to have this file we could remove it.

packages/dialtone-vue2/components/hovercard/hovercard_default.story.vue
```diff
@@ -0,0 +1,122 @@
...
+</template>
+
+<script setup>
+import DtHovercard from './hovercard.vue';
+import DtRecipeContactRow from '@/recipes/leftbar/contact_row/contact_row.vue';
+import defaultImage from '@/common/assets/avatar2.png';
+import DtStack from '../stack/stack.vue';
+import DtIcon from '../icon/icon.vue';
+import DtButton from '../button/button.vue';
+import DtAvatar from '../avatar/avatar.vue';
```
iropolo, line 121:
```
import { DtStack } from '@/components/stack';
import { DtIcon } from '@/components/icon';
import { DtButton } from '@/components/button';
import { DtAvatar } from '@/components/avatar';
```

packages/dialtone-vue2/components/hovercard/hovercard_many.story.vue
```diff
@@ -0,0 +1,165 @@
...
+<script setup>
+import DtHovercard from './hovercard.vue';
+import DtRecipeFeedItemRow from '@/recipes/conversation_view/feed_item_row/feed_item_row.vue';
+import DtRecipeEmojiRow from '@/recipes/conversation_view/emoji_row/emoji_row.vue';
+import avatar1 from '@/common/assets/avatar1.png';
+import avatar2 from '@/common/assets/avatar2.png';
+import DtStack from '../stack/stack.vue';
+import DtIcon from '../icon/icon.vue';
+import DtButton from '../button/button.vue';
+import DtAvatar from '../avatar/avatar.vue';
```
iropolo, line 145:
```
import { DtStack } from '@/components/stack';
import { DtIcon } from '@/components/icon';
import { DtButton } from '@/components/button';
import { DtAvatar } from '@/components/avatar';
```

packages/dialtone-vue2/components/hovercard/hovercard.vue
ninamarina: Do you have an example of a component where this is done? I see the docs are filled automatically from these comments

packages/dialtone-vue2/components/popover/popover.vue
```diff
@@ -917,6 +962,56 @@ export default {
...
+    },
+
+    setOutTimer () {
+      this.outTimer = setTimeout(() => {
+        this.isOpen = false;
+      }, TOOLTIP_DELAY_MS);
+    },
+
+    onEnterAnchor () {
+      if (!this.hovercard) return;
```
iropolo, line 984: I would suggest to have a `isHovercard` computed property to improve readability, instead to use `this.hovercard`

packages/dialtone-vue2/components/hovercard/hovercard.test.js
```diff
@@ -0,0 +1,147 @@
...
+
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.runAllTimers();
+    updateWrapper();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+    vi.clearAllTimers();
```
iropolo, line 60: Not so sure about using timers in this tests....
If you remove all of them the tests still passes, It would be necessary to debug this a little bit to see if it indeed affecting something. 
If it is not, maybe it is trowing a fake check.

packages/dialtone-vue2/components/hovercard/hovercard.vue
iropolo: I mean, we are affecting the component prop section for storybook purpose.

Thats the point of adding stuff like this:
 ```
* <a
   *   class="d-link"
   *   href="https://popper.js.org/docs/v2/modifiers/flip/#fallbackplacements"
   *   target="_blank"
   * >
   *   Popper.js docs
   * </a>
   * */
```
Screenshot 2024-01-11 at 15 18 35

As you can see I just point it out there is a link where the consumer can check whatever they want related the prop usage. But if this addition also does not add value I will just skip it. (For example if it is does not explain something else that we are not explaining in our doc).

What I suggest is to do something like this:

image


Then we avoid to create noise in component prop description.

It is just a suggestion it is not necessary to be done on this way, but just heads up we are interfering component props doc for storybook purposes.

File: packages/dialtone-vue2/components/lazy_show/lazy_show.vue [hunk 13]
```diff
@@ -11,7 +11,7 @@
       v-on="$listeners"
     >
       <!-- @slot Slot for main content -->
-      <slot v-if="initialized" />
```
braddialpad: Removing this makes the component not perform lazy show anymore. the slot content will be rendered immediately rather than rendering on first time it is displayed. That said there is probably a better way of doing this. In vue 3 maybe we can use suspense? https://vuejs.org/guide/built-ins/suspense.html

packages/dialtone-vue2/components/hovercard/hovercard.vue
braddialpad: I'm not sure we've been super strict on this historically, but it does make sense to make our in component comments pure jsdoc.. maybe something like this

```
/**
@see https://atomiks.github.io/tippyjs/v6/all-props/#placement
@values top, top-start, ...
```

and then in the stories.js override the description with the markup shown here in argTypes

packages/dialtone-vue2/components/hovercard/hovercard.vue
braddialpad: Tooltip has this validator, but popover doesn't for some reason... can probably just copy it from tooltip.

packages/dialtone-vue2/components/hovercard/hovercard.test.js
ninamarina: This is quite weird, without the timers, the tests do pass, but the popover `isOpen` is never set. So it's giving a false positive.
With the timers, this log does show:
image

Will investigate, thanks for catching this!

packages/dialtone-vue3/components/hovercard/timer.js
```diff
@@ -0,0 +1,61 @@
+import { ref } from 'vue';
+import { TOOLTIP_DELAY_MS } from '@/components/tooltip/index.js';
+
+export default (function () {
+  const current = ref(null);
+  let timerLeave, timerEnter;
+  let start = null;
+  let prevCard = null;
+
+  function enter (id) {
```
braddialpad, line 15: Maybe some short comment description, it's a bit hard to tell what is going on here without going through it line by line.

packages/dialtone-vue3/components/hovercard/timer.js
```diff
@@ -0,0 +1,61 @@
...
+      }
+    } else {
+      timerEnter = setTimeout(() => {
+        current.value = id;
+        prevCard = id;
+      }, TOOLTIP_DELAY_MS);
+    }
+  }
+
+  function leave () {
```
braddialpad, line 48: Same here with comment

braddialpad (changes requested):

packages/dialtone-vue2/components/lazy_show/lazy_show.vue
ninamarina: oh, that makes sense. Will probably have to find another ways of testing this, unless suspense works better with the automation. Thanks!

ninamarina: I've been checking this out and even without the "initialized" logic, the component is not rendered the first time, because the show prop is still false. So I don't think this is actually making any performance difference.
Without the `v-if="initialized"` I can see that the DOM is only updated when the "show" property is true:
init-vue3

The slot from lazy-show is inside the div we see appearing and disappearing from the DOM in the gif:
image

I also tried this is vue 2 and it works the same way.

For what I can see, Suspense is for managing lazy loading of async dependencies, which is not what lazy show seems to be doing.

packages/dialtone-vue2/components/hovercard/hovercard.mdx
ninamarina: Yes, will do

packages/dialtone-vue2/components/lazy_show/lazy_show.vue
ninamarina: Disregard my previous comment, the div is actually being rendered (with `display: none`), until the first time the anchor is hovered. It is inside the popover:
image

When doing the first hover, tippy moves this to the body element. So lazy-show is doing a render optimization.

The problem with the test was that I was expecting the elements to be inside the wrapper, while they are on the body.
This explains also what @iropolo pointed out about the tests, they are passing even without the timers because the div is being rendered only with `display: none`.

I will revert the changes to `lazy_show` and test that the element is present in the body element.

braddialpad: Added ticket to look into the lazy-show further. https://dialpad.atlassian.net/browse/DLT-1477

once we get tests passing here, we're good to move forward.

packages/dialtone-vue3/components/hovercard/timer.js
ninamarina: Fixed in latest changes, let me know if it's clear enough

packages/dialtone-vue3/components/hovercard/timer.js
ninamarina: Fixed in latest changes

packages/dialtone-vue2/components/hovercard/hovercard.mdx
ninamarina: Fixed in latest changes

packages/dialtone-vue2/components/hovercard/hovercard.mdx
ninamarina: Fixed in latest changes

packages/dialtone-vue2/components/hovercard/hovercard.test.js
ninamarina: Fixed in latest changes

packages/dialtone-vue2/components/hovercard/hovercard.vue
ninamarina: Fixed in latest changes

packages/dialtone-vue2/components/hovercard/hovercard.vue
ninamarina: Fixed in latest changes

packages/dialtone-vue2/components/hovercard/hovercard_constants.js
ninamarina: Fixed in latest changes

packages/dialtone-vue2/components/hovercard/hovercard.test.js
ninamarina: Fixed in latest changes

packages/dialtone-vue2/components/lazy_show/lazy_show.vue
ninamarina: Fixed in latest changes

packages/dialtone-vue3/components/hovercard/timer.js
braddialpad: good description thanks.

braddialpad (approved):
Looks good by me, thanks!

iropolo:
In vue2 version the hovercard close and open on click, this does not happen in vue3 version.
This could lead to an issue if you close the hovercard and mouseover again it quickly it does not appear again.

packages/dialtone-vue2/components/hovercard/hovercard.vue
iropolo: looks great

iropolo:
Will suggest to move the template to the middle of the screen, like what we have on popover, in that way we can better see the `placement` prop working well.

ninamarina:
> In vue2 version the hovercard close and open on click, this does not happen in vue3 version. This could lead to an issue if you close the hovercard and mouseover again it quickly it does not appear again.

Nice catch! Fixed in latest commit

iropolo (approved):
Looks great, nice job!!

PR #91, opened 2024-01-09
docs: fix links starting with /

braddialpad:
Noticed in our deploy-previews any links starting with / would link directly to dialtone.dialpad.com/whateverlink rather than dialtone.dialpad.com/deploy-previews/pr-xx/whateverlink. It seems to do this when a link has a preceding slash... Seems like a bit of a vuepress bug because I don't think it should work like that when baseUrl is set correctly (it is). Anyways the fix without making changes to vuepress is just to remove preceding slash from internal links.


iropolo (approved):
Thanks

PR #102, opened 2024-01-15
fix: emoji_picker box-sizing

iropolo:
# fix: emoji_picker box-sizing

In `with_popover` story.
Seems like popover is using `box-sizing: border-box` style, this is malfunctioning the emoji_pikcer style.

I added `box-sizing: content-box !important;` to override it with the default value.

Im not sure why or what could trigger this issue now.

Fix for Vue2 and Vue3 version.

BEFORE:
Screenshot 2024-01-15 at 13 12 33

AFTER:
Screenshot 2024-01-15 at 13 12 53

### DIFF
--- a/packages/dialtone-css/lib/build/less/components/emoji-picker.less
+++ b/packages/dialtone-css/lib/build/less/components/emoji-picker.less
@@ -38,6 +38,8 @@
     width: auto;
     max-width: calc(var(--dt-size-925) + var(--dt-size-400));
     margin: 0 var(--dt-space-500);
+    // !important to default value to override popover dialog box-sizing: border-box style
+    box-sizing: content-box !important;
   }
 
   &--footer {


iropolo:
Visual test passes, does not detect any change, as expected

packages/dialtone-css/lib/build/less/components/emoji-picker.less
```diff
@@ -38,6 +38,8 @@
     width: auto;
     max-width: calc(var(--dt-size-925) + var(--dt-size-400));
     margin: 0 var(--dt-space-500);
+    // !important to default value to override popover dialog box-sizing: border-box style
+    box-sizing: content-box !important;
```
braddialpad, line 42: I am questioning whether it's better to use important, or nested css to increase specificity. I think this is okay in this case.

braddialpad (approved):

packages/dialtone-css/lib/build/less/components/emoji-picker.less
iropolo: This selector `.d-popover__dialog *` it's strictly invasive, due to `*` it is being applied everywhere.

I prefer also the explicit `!important!` rule here. To expose this popover directive.

braddialpad:
Gonna do a release, so merging this.

PR #107, opened 2024-01-18
fix(select-menu): corrected selectmenu's icon color in dark mode

francisrupert:
# PR Title

Corrected selectmenu's icon color in dark mode

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :pencil: Checklist

For all CSS changes:

image

### DIFF
--- a/packages/dialtone-css/lib/build/less/components/selects.less
+++ b/packages/dialtone-css/lib/build/less/components/selects.less
@@ -41,6 +41,10 @@
     transform: translateY(-50%);
     content: '';
     pointer-events: none;
+
+    .dialtone-theme-dark & {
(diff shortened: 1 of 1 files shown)


francisrupert:
NOTE: this still isn't an ideal method. Ultimately a DT Icon should be placed inline instead of the current embedded base64 image. And using `.dialtone-theme-dark * {}` may not scale well in the long-run.

braddialpad:
> using `.dialtone-theme-dark * {}` may not scale well in the long-run.

Yeah we will likely have to change this fairly soon for web components, not to mention multidimensional themes.

> this still isn't an ideal method. Ultimately a DT Icon should be placed inline instead of the current embedded base64 image

Can we just do this now? we have the icon: "chevrons-up-down"

francisrupert:
> > using `.dialtone-theme-dark * {}` may not scale well in the long-run.
> 
> Yeah we will likely have to change this fairly soon for web components, not to mention multidimensional themes.
> 
> > this still isn't an ideal method. Ultimately a DT Icon should be placed inline instead of the current embedded base64 image
> 
> Can we just do this now? we have the icon: "chevrons-up-down"

Breaking change right? Just searching DP and DPM codebases there quite a few instances of using the raw HTML of `.d-select` and those would need to be updated.

Updating the Vue component to use DTIcon is of course easy enough. Perhaps do so after this temporary quick update and put in backlog?

braddialpad:
> Breaking change right? Just searching DP and DPM codebases there quite a few instances of using the raw HTML of `.d-select` and those would need to be updated.

Yeah right, I was just thinking we could do this in the vue component and it wouldn't be a breaking change.

> Perhaps do so after this temporary quick update and put in backlog?

Yep sounds good

braddialpad (approved):

PR #108, opened 2024-01-18
fix(popover): set text color within popover dialog

braddialpad:
# fix(popover): set text color within popover dialog

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DP-88061

## :book: Description

Adding a default text color into d-popover__dialog

## :bulb: Context

Kind of a strange issue reported by a customer.. Seems like the tippy class is being overridden by some sort of local browser extension and making the text white. Can't confirm this fixes it since nobody can repro this locally, but worth giving a try.

## :pencil: Checklist

For all PRs:

For all CSS changes:

## :crystal_ball: Next Steps

Update in product after release

### DIFF
--- a/packages/dialtone-css/lib/build/less/components/popover.less
+++ b/packages/dialtone-css/lib/build/less/components/popover.less
@@ -40,6 +40,7 @@
     display: grid;
     grid-template-rows: min-content 1fr min-content;
     overflow: auto;
+    color: var(--dt-color-foreground-primary);
     background-color: var(--popover-color-background);
     background-clip: padding-box;
     border: var(--popover-border-width) solid var(--popover-color-border);


francisrupert (approved):
[DP-88061](https://dialpad.atlassian.net/browse/DP-88061) is a wild bug and I agree there's very likely something overriding; border color too.

This is a reasonable addition, even if redundant since the same color is inherited from `<body>`, so declaring the foreground color here is safe. 

Meanwhile, the Dropdown-like screenshot isn't strictly a DT Dropdown, though it is using DT Popover. If it were a DT Dropdown, each list item has default/hover/active colors and possibly wouldn't have this bug.

PR #118, opened 2024-01-19
feat: datepicker full language support

iropolo:
# Datepicker full language support

Datepicker now has full language support.

Previously month didnt has the `locale` prop working, now to make it work we are using the ECMAScript Internationalization API.

Also all the utils functions are working with the given `locale`.

You can test it going to the deploy preview and updating the `locale` prop in controls sections with something like `ja` `es` `de`

BEFORE:
image

AFTER:
image


iropolo:
No visual test required.

braddialpad (approved):

PR #124, opened 2024-01-24
fix: datepicker keyboard navigation fix vue2 only

iropolo:
# Datepicker keyboard navigation fix vue2 only

Some background:

We received a bug fix report from [DP-86647](https://dialpad.atlassian.net/browse/DP-86647).

Steps to reproduce it are in the ticket comment section.

Seems like keyboard navigation does not work well in the product.

It works ok in storybook, in both variants, default and with popover.

-----

Doing research seems like refs are not being assigned.

This methodology was failiing in product `el => { if (el) setDayRef(el, day) }`.

Then it is necessary to rethink how to write this dynamic ref.

This issue is only vue2, since in vue3 `template refs` works in slightly different way. (Also this report comes from vue2, not vue3, I would not touch vue3).

-----

I did the necessary refactor to make it work and in my local tests it is working as expected in product.

-----

I didnt dig on why it is happening on the product side.
Because storybook is vue2 and its works well.

Would be cool to get and tackle the origin of this issue.

There is a chance were we get this same issue on our emoji picker in the vue2 product side, or may not.

## How you can test this fix:
On our design-system repo folder
Go to this branch `git checkout datepicker-navigation-fix`
Run `npm run build`

On firespotter repo folder
Go to `firespotter/ubervoice/static/node_modules/@dialpad/dialtone/dist` and remove vue2 folder
Run `npm run clean`
Copy the `vue2` folder on our build of our design-system repo `design-system/dist/vue2`
Paste it in `firespotter/ubervoice/static/node_modules/@dialpad/dialtone/dist`
Run `npm run dev`

Go to `https://localhost-dot-sh-dot-uv-beta.appspot.com/callcenters`
_(You may need have some permissions to go to that section, if thats the case ping me)_
Click on `Add holiday` button and test the datepicker 
More details of steps to reproduce in [DP-86647](https://dialpad.atlassian.net/browse/DP-86647) comment section.

### BEFORE FIX:
https://github.com/dialpad/dialtone/assets/89984179/b74dd2f9-7f17-4d1b-aa73-f61cf9114801

### AFTER FIX:
https://github.com/dialpad/dialtone/assets/89984179/a5a11801-017f-4e67-a105-4692f8b33f44

[DP-86647]: https://dialpad.atlassian.net/browse/DP-86647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iropolo:
No visual tests are required.

packages/dialtone-vue2/components/datepicker/modules/calendar.vue
```diff
@@ -30,7 +30,7 @@
           class="d-datepicker__cell"
         >
           <dt-button
-            :ref="el => { if (el) setDayRef(el, day) }"
```
braddialpad, line 33: hmm yeah not sure this function within ref syntax is supported in vue 2.. still strange it works fine in storybook and not product though. Maybe it's a vue 2.7 vs vue 2.6 thing.

braddialpad (approved):
Bit of an odd bug. fix seems reasonable let's see if it works in product.

packages/dialtone-vue2/components/datepicker/modules/calendar.vue
iropolo: Huh... I didnt think that product has a different vue 2 version that the one we are delivering solutions and components...

braddialpad: Yep they are still on 2.6. Maybe we should revert dialtone back to 2.6 because this could definitely cause problems.

PR #130, opened 2024-01-26
fix: svg id fix - vue2

iropolo:
# fix: svg id fix

What Im doing:

- Import svg as a raw file to manipulate it as string.
- Find and replace all id with unique ids.
- Mount the svg as v-html.
- Removing the parent div and copy all the attrs to the svg.
- Created an observer to trigger the update on component changes (v-show, class, attributes, etc)
- Final result: Reactive / inline svg / independent rendered.

------

- This is a vue2 fix, vue3 fix will come after this POC approval.
- Unit tests probably fails cos I didn't updated it.
- Probably some improvements comes after this POC approval.
- Play with it to try to break or find something to improve.
- Focus on the proposal fix, all dummy code will be removed in the definitive PR

You can try the fix here:
https://dialtone.dialpad.com/vue/deploy-previews/pr-130/?path=/story/components-icon--default


iropolo:
Im preparing for this PR a few more proposal, like using the https://github.com/iconfu/svg-inject (if it works) and another one using props, then we will have more options to talk about.

braddialpad:
I think this will work but man is it ever a lot of code just to get unique ID's on svgs. I understand it is actually a pretty complex problem. We should see if we can get svg-inject or some other library based solution working, if not we can go with this.

iropolo:
Okay.. I added two more plugins to test it:

### [SVGInject](https://github.com/iconfu/svg-inject): 
With this one seems like it lost reactivity when it is doing the 
`inject` function (replacing the `` tag with `<svg>`
 in `packages/dialtone-vue2/node_modules/@iconfu/svg-inject/src/svg-inject.js`

```
 // Replace img element with new element. This is the actual injection.
        parentNode.replaceChild(injectElem, imgElem);
```

I faced this issue also in my fix, but I figure it out with an `observe` function.
So without this reactivity, it does not change the icon and also it is not reacting to the `v-show`.

### [svgLoader](https://github.com/jpkleemans/vite-svg-loader): 
Not sure if this plugin is working properly but ye... didnt make it work.

-----
I cannot deal with the img src to get the svg and use it there.
If you check out the branch you will see the icon and how the plugin is not working.

iropolo:
Some other ideas:

- we can propose is tell the consumer to use `v-if` instead `v-show`.
- this same component fix that I create, but to avoid the observe we can use a prop like isVisible.
So.. the user uses this prop instead the v-show but.... you know...., it should work and it is expected to work with v-show

----

The good new about doing this on our way is it is ours... we give support to this and it will depend just on us.
Also, this is probably compatible with vue3 without any major change (hopefully)

iropolo:
Another solution is this same component fix that I create, but to avoid the `observe` we can use a prop like `isVisible`.
So.. the user uses this prop instead the `v-show` but.... you know...., it should work and it is expected to work with `v-show`

iropolo:
The good new about doing this on our way is it is ours... we give support to this and it will depend just on us.
Also, this is probably compatible with vue3 without any major change (hopefully)

braddialpad:
>   So.. the user uses this prop instead the v-show but.... you know...., it should work and it is expected to work with v-show

Yeah I don't think this is a good solution, it will be too confusing for developers.

Maybe it's best we do use your custom solution here, but maybe we abstract it into a more generic form since this could possibly be needed for any SVGs not just icons.

iropolo:
> > but maybe we abstract it into a more generic form since this could possibly be needed for any SVGs not just icons.


k, lemme clean the PR and get a final solution using the initial propsed fix.

iropolo:
Final svg id fix ready to go.

iropolo:
Refactoring tests since this is a breaking change on how to select the icon element.
It is needed also to update it in product.

iropolo:
Fix will go in #145

PR #131, opened 2024-01-29
fix(tooltip): stop showing tootip for touch based devices

kdsanskar:
# fix(tooltip): stop showing tootip for touch based devices

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1271

## :book: Description

Hide the tooltip for touch based device.

## :bulb: Context

 
Tooltip should not showup for touch based devices as there is no hover event, yet it is showing up when user clicks the anchor element which is not even hover.

## :pencil: Checklist

For all PRs:

For all Vue changes:


packages/dialtone-vue3/components/tooltip/tooltip.vue
```diff
@@ -364,6 +370,8 @@ export default {
     this.externalAnchor && this.addExternalAnchorEventListeners();
     this.tip = createTippy(this.anchor, this.initOptions());
 
+    this.isTouchDevice = (('ontouchstart' in window) || (navigator.maxTouchPoints > 0) || (navigator.msMaxTouchPoints > 0));
```
braddialpad: This is not a great way to detect touch as this only checks if they are using a touch enabled device. What if someone was on a laptop with a touchscreen but they were using the mouse? In this case tooltips would not be shown.

Instead lets add the "touchstart" event to the wrapper span element `@touchstart="onTouchStart"`. When this event is triggered it means the user has actually "touched" the element and in this case we should not show the tooltip. This should be much easier to test as well.

braddialpad (changes requested):

packages/dialtone-vue3/components/tooltip/tooltip.vue
kdsanskar: @braddialpad I have made changes based on your comment, and it did helped in writing unit tests.

braddialpad (approved):
Looks good thanks!

packages/dialtone-vue2/components/tooltip/tooltip.vue
```diff
@@ -271,13 +272,15 @@ export default {
...
+
+      isTouchDevice: false,
     };
   },
 
   computed: {
     // whether the tooltip is visible or not.
     isVisible () {
-      return this.isShown && this.enabled && (!!this.message.trim() || !!this.$slots.default);
+      return this.isShown && this.enabled && (!!this.message.trim() || !!this.$slots.default) && !this.isTouchDevice;
```
iropolo: I would like to improve this computed property, could be something like this?

  ```
isVisible() {
    const hasMessage = !!this.message.trim();
    const hasDefaultSlot = !!this.$slots.default;
    const isDeviceCompatible = !this.isTouchDevice; 

    const shouldBeVisible = this.isShown && this.enabled && (hasMessage || hasDefaultSlot);

    return shouldBeVisible && isDeviceCompatible;
  },
```

I guess it will work as this way, double check it please to be sure, and make this update also in the vue3 version.

iropolo (approved):
Looks good, thanks

ninamarina:
The code looks good, though it seems it still shows the tooltip every once in a while when tapping the anchor. 
I tested it on my phone, every ten taps the tooltip shows:
image

It's still much better that what we have now, where the tooltip is shown in every tap.

francisrupert (approved):

packages/dialtone-vue2/components/tooltip/tooltip.vue
kdsanskar: Done

kdsanskar:
> The code looks good, though it seems it still shows the tooltip every once in a while when tapping the anchor. I tested it on my phone, every ten taps the tooltip shows: 
image

@ninamarina This issue is cuz `mouseenter` event is also getting triggered after `touchstart` event, I have tried to stop that using `relatedTarget` property but still even after that in chrome browser once in while what is happening is that property `relatedTarget` is getting not null for touch event which it should not, meanwhile I have tested in firefox it worked fine.

@ninamarina @braddialpad Please let know if there is something more we can do fix it, otherwise it looks like a browser issue in chrome.

refer: https://copyprogramming.com/howto/jquery-on-click-triggers-mouseover-on-touch-device

ninamarina (approved):

PR #133, opened 2024-01-29
chore: rename markdownlint config

braddialpad:
VSCode extension didn't like the name `.markdownlintrc` config file had to be renamed `.markdownlint.json` now it works.


juliodialpad (approved):
Thanks for fixing this, I couldn't notice because I'm not using VSCode

PR #139, opened 2024-01-31
fix(feed-item-row): set inset focus ring on feed item

braddialpad:
Set inset focus ring on feed-item so the focus ring doesn't go behind the borders of the feed list.


yorbi-dp (approved):
LGTM!

PR #145, opened 2024-02-07
feat: universal dialtone-icons

juliodialpad:
# Feat: Universal dialtone-icons

Obligatory GIF

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket

No JIRA

## :book: Description

Vue-demi didn't work well, made a bunch of POC and every single one failed in a different way that I couldn't fix for a version of Vue, so ended up creating a common file structure from where we can build both Vue versions.

- Created two folders inside dialtone-icons (vue2, vue3) which only contain configuration to build dialtone-icons into vue 2 and vue 3
- Created a script which reads and converts every SVG icon into a vue component that can be renderer on both vue 2 and vue 3
- Updated dialtone-vue 2 and dialtone-vue 3 to use the new dialtone-icons schema.
- Added a global .eslintrc.cjs file to fix errors while committing files from folders that don't have eslint configuration.
- Added global common/utils, we can use this for function like `getUniqueString` to avoid duplicating it on vue2 and vue3 (migration to this global common function is pending)

## :bulb: Context

We were having issues with duplicated id's on product, we needed to found a way to create a custom ID for every icon.

## :pencil: Checklist

For all PRs:

For all Vue changes:


packages/dialtone-vue3/components/icon/icon.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <component
     :is="icon"
-    v-show="name"
+    v-if="icon"
```
iropolo, line 4: Why it was different as the vue2 code?

common/utils.js
```diff
@@ -0,0 +1,4 @@
+let UNIQUE_ID_COUNTER = 0;
+export function getUniqueString (prefix = 'dt') {
```
iropolo, line 2: huh, I would prefer to have this like `dt-unique`. Just because maybe `dt1` or `dt2` is more probably to be used.

Or even better, `dt-unique-counter-`, so you can easily know that is coming from this util.

packages/dialtone-icons/transformSVGtoVue.cjs
```diff
@@ -0,0 +1,27 @@
...
+
+    fileNames.forEach(function (fileName) {
+      const svgContent = fs.readFileSync(`./dist/svg/${fileName}`, 'utf8');
+      const template = fs.readFileSync('./src/IconTemplate.vue', 'utf8');
+      const iconName = 'Icon' + _.startCase(_.camelCase(fileName.split('.')[0])).replace(/ /g, '');
+
+      let result = template.replace('__SVG_CONTENT__', svgContent);
+      result = result.replace('__ICON_NAME__', iconName)
+      result = result.replace(/clip-path="url\([^)]+\)"/g, ':clip-path="`url(#${uniqueID})`"')
+      result = result.replace(/clipPath id="[^"]+"/g, 'clipPath :id="uniqueID"')
```
iropolo: Here is where we introuduce uniqueId to the svg to avoid id conflicts?

packages/dialtone-vue3/components/icon/icon.vue
juliodialpad: Maybe due to some rendering differences on vue 2 and vue 3 trying to import the SVG's directly, but that shouldn't be a problem anymore

common/utils.js
juliodialpad: That's just the default prefix, for the icons I'm using `dt-icon` prefix and we're currently just using this util on icons, once we move the local usage of dialtone to this everything would be ok, I don't see any problem.

packages/dialtone-icons/transformSVGtoVue.cjs
juliodialpad: Yes, It's a little obscure as the actual function is on the IconTemplate file, but this works perfectly.

iropolo: There is some more ids to replace, e.g in whatsapp are
`linearGradient id="paint0_linear_3534_78"`
then its being used in `fill="url(#paint0_linear_3534_78)` inside `path` 

Im guessing you need to replace these too, to make it work and avoid the conflict

packages/dialtone-icons/package.json
```diff
@@ -3,18 +3,30 @@
   "version": "4.8.1",
   "description": "Dialtone icon library",
   "scripts": {
-    "build": "gulp"
+    "build": "gulp && nx run-many --target=build --projects=dialtone-icons-* --parallel=2 --skip-nx-cache"
```
braddialpad: skip-nx-cache is needed here?

packages/dialtone-icons/src/icons/index.js
```diff
@@ -0,0 +1 @@
+export const icons = import.meta.glob('./*.vue', { import: 'default', eager: true });
```
braddialpad, line 1: In theory if we turned off eager: true, this would dynamically load icons as they are used instead of loading them all up front? Obviously we'd have to test this, maybe a future upgrade though.

packages/dialtone-vue2/package.json
```diff
@@ -111,7 +111,7 @@
   },
   "peerDependencies": {
     "@dialpad/dialtone-css": "^7.30.0 || ^8.21.2",
-    "@dialpad/dialtone-icons": "^4.3.0",
+    "@dialpad/dialtone-icons": "latest",
```
braddialpad: We had a lot of problems with latest before is it safe to use here?

packages/dialtone-icons/transformSVGtoVue.cjs
iropolo: I have this code in the other PR maybe it useful to have it here

It find the ids and replace each.

https://github.com/dialpad/dialtone/pull/130/files#diff-b33ea21eb11057157f9398fd1ed9386e9a33089f18eda2727c99e2058fe22af8

 ```
modifySvgIds (svgContent) {
      // Find all 'url(#id)' and 'id="id"' instances
      const idRegex = /url\(#([^)]+)\)|id="([^"]+)"/g;
      // Generate a unique ID
      const uniqueId = (originalId) => `${this._uid}-${originalId}`;
      // Replace all 'url(#id)' and 'id="id"' instances with 'url(#uniqueId)' and 'id="uniqueId"
      return svgContent.replace(idRegex, (match, urlId, normalId) => {
        const idToReplace = urlId || normalId;
        if (idToReplace) {
          return match.replace(idToReplace, uniqueId(idToReplace));
        }
        return match;
      });
```

juliodialpad: Thanks! replaced

packages/dialtone-icons/package.json
juliodialpad: I might need to touch which files nx needs to cache, It wasn't detecting changes on icons so it was using cache always, I'll try to fix it to not have skip cache

juliodialpad: Fixed on https://github.com/dialpad/dialtone/pull/145/commits/f1a7e33a7b296313dcde2bab28a175bb282d5155

packages/dialtone-icons/src/icons/index.js
juliodialpad: Tried it, but Vue 3 was having issues to render the icons correctly, surely agree that it can be a future update.

packages/dialtone-vue2/package.json
juliodialpad: I think it's safe as we're no longer using `@dialpad/dialtone-vue` package on product and this will only affect that specific package.

I think it might benefit anyone having the dialtone-vue package as it'd always have the latest icons, right?

packages/dialtone-icons/src/icons/index.js
braddialpad: Yeah I think it might not be that simple, thanks for checking

packages/dialtone-vue2/package.json
braddialpad: I'm thinking about the implications of setting this here.. I think it might cause some issues with deployment stability

If we set the peer dependency to latest, that means it will automatically update to latest upon npm install. This means if they do a beta cut for the week and then we do a new release of dialtone-icons, it will either start erroring out here saying the peer dependency is incorrect, or it will automatically update dialtone-icons to the latest next time they do a build on beta. This would be bad because we could potentially introduce a bug in beta without actually updating dependencies

braddialpad: > I think it's safe as we're no longer using `@dialpad/dialtone-vue` package on product and this will only affect that specific package.

Doesn't our monopackage also reference package jsons from other packages? It would have to right?

juliodialpad: Not really, it if was that way we wouldn't need to manually define exports.

The dependencies, devDependencies and peerDependencies are independent and isolated on each package. I'm pretty sure that peer dependency it's not going to affect the mono package.

But it makes sense to change that to `workspace:*` instead, that way it'd be the latest version number we released on purpose, are you ok with that approach?

juliodialpad:
Updated dependencies workspace version, changed from `workspace:^` to `workspace:*` the difference is the version translated when published.

For example if dialtone-css version 8.24.4, translations would be:

- `workspace:^` = ^8.24.4
- `workspace:*` = 8.24.4

Changing this, gives us more control on which dependencies are being installed on each individual package.

@braddialpad @iropolo @ninamarina

juliodialpad:
Had to change code on `build-sd-transforms.js` because now it's linted by Eslint and currently it doesn't support type assertions https://github.com/eslint/eslint/discussions/15305#discussioncomment-1634740

@braddialpad

iropolo (approved):
So instead of doing the svg id update on the component side, we are doing it in the gulp build.
Then we are delivering already modified svg to our library.

Looks good to me and it seems to work as expected.

I would describe this behaviour in the dialtone icon repository readme.

juliodialpad:
> So instead of doing the svg id update on the component side, we are doing it in the gulp build. Then we are delivering already modified svg to our library.
> 
> Looks good to me and it seems to work as expected.
> 
> I would describe this behaviour in the dialtone icon repository readme.

No, we are delivering vue component library instead of SVG's, the library has the capacity to auto-prefix icons with unique ID's.

I'll update the documentation to note this behavior.

PR #149, opened 2024-02-09
fix(feed-item-row): change typography to compact

braddialpad:
# fix(feed-item-row): change typography to compact

- [x] Fix

## :book: Description

Change the typography to body compact as requested by design. 

Render emoji's within the default example so we can validate what it looks like.


francisrupert (approved):
I approve at least from a design perspective in the interest of unblocking this change. I defer to you or other reviewers for approval from Vue perspective.

PR #152, opened 2024-02-13
fix: load archivo font

braddialpad:
# fix: load archivo font

- [x] Fix

## :book: Description

Archivo font was not being loaded even though it was provided in the package and used in utility classes. Added into dialtone-globals.


juliodialpad (approved):
Thanks!

PR #154, opened 2024-02-13
feat: transform Figma variables to sd format

juliodialpad:
# Feat: transform Figma variables to sd format

Obligatory GIF

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket

No Jira Ticket

## :book: Description

- Added custom type (`sizing`, `spacing`, etc) after getting tokens from Figma to match our current implementation
- Added a transformation script that reads base and root files on `figma_tokens` directory, iterates through it and replaces the values accordingly.
- Currently that script is very focused on just reading those two files as it's the only I have currently, but can be adapted in the future to read any number of files.

## :bulb: Context

- We're moving forward into using Figma variables API, with this, we can actually start syncing sizing and spacing variables from Figma
- Syncing from the code to Figma it's possible now but we still need to standardize a way to do this to avoid confusions on what files need to be modified.

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

- Standardize and document a way to sync tokens to Figma


francisrupert:
@juliodialpad I've added @yorbi-dp as a reviewer. Since he wouldn't directly PR review the code, the two of you should figure out a way that most makes sense for him to review and confirm the **resulting** CSS variables are unchanged.

yorbi-dp (changes requested):
This looks great Tico! Nice work 😃 everything it's matching the current tokens and these new ones created from the Figma Variables.
- Just the token  `--dt-size-radius-circle` has a different value. Left a comment in the [evidence google sheet.](https://docs.google.com/spreadsheets/d/1H1pBc9m_cfIWbKgEEtaqxKiwTes5sBcjLAFX19rXVM8/edit#gid=0)

The rest it's great! Thanks.

yorbi-dp (approved):
LGTM! Thank you!! 🙌

packages/dialtone-tokens/.eslintrc.cjs
braddialpad, line 1: We need typescript?

packages/dialtone-tokens/dialtone-transforms.js
```diff
@@ -34,31 +34,30 @@ const pxToRemTransformer = (token, options) => {
   }
 
   if (floatVal === 0) {
-    return '0';
+    return '0rem';
```
braddialpad, line 37: didn't you say you wanted this to be 0 instead of 0rem?

packages/dialtone-tokens/sync-scripts/token_export.ts
```diff
@@ -15,6 +15,27 @@ function tokenTypeFromVariable(variable: Variable) {
...
+          return 'borderWidth';
+        case 'radius':
+          return 'borderRadius';
+        default:
+          return 'sizing';
+      }
+    default:
+      return undefined;
+  }
+}
```
braddialpad: Not that big of deal, but would an object map work better?

```
customTypeMap = {
  'space': 'spacing'
  'size': {
    border: 'borderWidth'
    ....
  }
}

// use like this
const variableNames = variable.name.split('/');
customTypeMap[variableNames[0]][variableNames[1]]
```

Your call if you like it better or not.

braddialpad (approved):
Looks pretty good. Couple comments but approving, this is good to go.

packages/dialtone-tokens/.eslintrc.cjs
juliodialpad: Just wanted to match the Figma provided scripts, I was having some weird errors about importing JS on a TS file so decided to change the script to TS (not fully migrated, just to enough to get rid of errors).

packages/dialtone-tokens/dialtone-transforms.js
juliodialpad: The other way around, the staging css output had `0rem` and after my changes the output was just `0` so added that to always have `0rem` on sizing and spacing.

packages/dialtone-tokens/sync-scripts/token_export.ts
juliodialpad: For sure I prefer the object map, I'll update it

braddialpad (approved):

PR #158, opened 2024-02-14
chore: fix release

juliodialpad:
# Fix release command

Obligatory GIF

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [x] Other (chore)

## :book: Jira Ticket

No Jira

## :book: Description

Changed the name of NX command from "release" to "publish" to avoid confusions with package.json "release".
The documentation doesn't need to be updated.

## :bulb: Context

Couldn't release because NX and pnpm commands had the same naming.

## :pencil: Checklist

For all PRs:


iropolo (approved):
~~`release` was a reserved word or it has some conflict?~~

~~... why we are changing from `release` to `publish` 🤔~~

Edit: oh I read the description now 🤪

Seems like it is still confusing to have `publish` and `release`.

What would be the explanation to have both?

NX `publish` will be the internal process, but the one in package (`release`) is what we want.

juliodialpad:
> ~`release` was a reserved word or it has some conflict?~
> 
> ~... why we are changing from `release` to `publish` 🤔~
> 
> Edit: oh I read the description now 🤪
> 
> Seems like it is still confusing to have `publish` and `release`.
> 
> What would be the explanation to have both?
> 
> NX `publish` will be the internal process, but the one in package (`release`) is what we want.

It's still confusing haha

`Release` is the process of running `semantic-release` (checking commits, updating versions and pushing commits)
`Publish` is actually publishing the packages with `pnpm publish`

The confusion was because we had in package.json scripts (release) and at the same time we had in project.json target (release).

And both were meant to do different things, so when you were trying to release (semantic-release) you were actually running `pnpm publish` which cannot be done locally but no error will pop-up, so you'll get stuck there.

The naming change is just to make it more clear `publish -> pnpm publish` `release -> semantic-release`

PR #160, opened 2024-02-19
feat(item-layout): add unstyled prop

juliodialpad:
# Item layout: add unstyled prop

Obligatory GIF

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1449

## :book: Description

- Added `unstyled` prop to item-layout
- Migrated item-layout from flexbox to grid

## :bulb: Context

- Currently item-layout includes custom stylings that we need to override if we need something custom, this change will enable us to use the item-layout without the custom styling so we can customize the styling easily. 

## :pencil: Checklist

For all PRs:

For all Vue changes:

For all CSS changes:

## :crystal_ball: Next Steps

Update components depending on item-layout to use the custom version and update styling accordingly.

- Default
Screenshot 2024-02-19 at 1 24 37 p m

- Custom
Screenshot 2024-02-19 at 1 26 18 p m


packages/dialtone-css/lib/build/less/components/item-layout.less
```diff
@@ -13,41 +13,79 @@
...
+        margin-top: var(--dt-space-200-negative);
+      }
+    }
+
+    .dt-item-layout--selected {
+      align-self: center;
+    }
+
+    .dt-item-layout--left {
+      padding-right: var(--dt-space-400);
```
ninamarina: Would it make sense to use gap instead of padding? The only cell that doesn't have this padding is selected and maybe it should have, for what I see in the use in list-item they are adding a margin:
image

iropolo:
In vue2, nothing happen when you update `type` to `custom` value.
In vue3 you can visualize the update:

Screenshot 2024-02-20 at 11 40 30

Screenshot 2024-02-20 at 11 41 32

packages/dialtone-vue2/components/item_layout/item_layout.mdx
```diff
@@ -25,3 +25,20 @@ import * as DtItemLayoutStories from './item_layout.stories.js';
...
+By default, item layout includes custom styling, like paddings, sizes, colors, etc.
+
+```html
+<dt-item-layout />
+```
+
+### With custom type
+
+Setting the `type` property to `custom`, removes all the custom styling maintaining the slots positions.
+This way you can utilize the layout and customize your own styling.
```
ninamarina: Maybe explain further how can the consumer use their own styling.
For what I see, this component is mostly used as a recipe, so is the idea that the consumer sets the paddings on the inner components? How could the padding for `dt-item-layout` be set? Or is the idea to use a wrapper if you need a padding for the whole thing?

packages/dialtone-css/lib/build/less/components/item-layout.less
```diff
@@ -13,41 +13,79 @@
...
+  display: grid;
+  grid-template-areas:
+    'left content right selected';
+  grid-template-columns: repeat(4, minmax(0, max-content));
 
-  &--content {
-    flex-grow: 1;
-    min-width: var(--dt-size-0);
-  }
+  &:not(&--custom) {
```
iropolo: This is the key change.
I would mention it in the component explanation (`.mdx`)

packages/dialtone-css/lib/build/less/components/item-layout.less
juliodialpad: Yes, makes sense to use gap instead of paddings, just migrated the old stylings, forgot to update them accordingly.

juliodialpad:
> In vue2, nothing happen when you update `type` to `custom` value. In vue3 you can visualize the update:
> 
> Screenshot 2024-02-20 at 11 40 30 Screenshot 2024-02-20 at 11 41 32

Nice catch! fixed on [abfd64e](https://github.com/dialpad/dialtone/pull/160/commits/abfd64ef43cc0013b13bdc5c80a168ecb6f772fe)

packages/dialtone-vue2/components/item_layout/item_layout.mdx
juliodialpad: `dt-item-layout` is basically just a wrapper component with slots, so If no styling is set on the slots you can pass whatever styling you want.

I think the explanation is enough, or do you have an example of what do you want to add here?

packages/dialtone-css/lib/build/less/components/item-layout.less
juliodialpad: This is mentioned [here](https://github.com/dialpad/dialtone/pull/160/files#diff-09cfc539bcbe93dff729532458a0a55b49605cc166a70ceab01c16c00b4bb1deR37-R44) or what else would you add to this?

braddialpad:
Hmm why are there visual changes in percy? should be the same as before since they aren't using the new prop right?

packages/dialtone-vue2/components/item_layout/item_layout.mdx
braddialpad: hmm does it make more sense to say "spacing" instead of styling? styling could also refer to typography and such.

packages/dialtone-css/lib/build/less/components/item-layout.less
```diff
@@ -13,41 +13,79 @@
...
-  display: flex;
-  align-items: center;
-  min-height: calc(var(--dt-size-550) + var(--dt-size-300));
-  padding: var(--dt-space-300) var(--dt-space-400);
-  font-size: var(--dt-font-size-200);
-  line-height: var(--dt-font-line-height-300);
+  display: grid;
+  grid-template-areas:
+    'left content right selected';
+  grid-template-columns: repeat(4, minmax(0, max-content));
```
braddialpad: nice, super happy we changed this to grid!

juliodialpad:
> Hmm why are there visual changes in percy? should be the same as before since they aren't using the new prop right?

Looks like the spacing changed, will verify that I migrated everything correctly.

packages/dialtone-vue2/components/item_layout/item_layout.mdx
ninamarina: @juliodialpad can you give an example of how can I set the gap to `var(--dt-space-300)`?

packages/dialtone-css/lib/build/less/components/item-layout.less
iropolo: Setting the `unstyled` property will add `dt-item-layout--custom` class that it removes all the custom styling maintaining the slots positions.

packages/dialtone-vue2/components/item_layout/item_layout.mdx
juliodialpad: > hmm does it make more sense to say "spacing" instead of styling? styling could also refer to typography and such.

It also remove some colors on the subtitle slot so it's not just spacing, I this styling its better than spacing in that case.

juliodialpad: > @juliodialpad can you give an example of how can I set the gap to `var(--dt-space-300)`?

You will not set the gap of the layout, you'll need to use unstyled version of the layout and use margin/padding on the slot, e.g:

```html
<dt-item-layout unstyled>
  <template #left>
    <span class="d-mr8">
      some content
    </span>
  </template>
  <span>
    other content
  </span>
</dt-item-layout>
```

I can add a prop to the layout to be able to set the gap, but I don't think that's what we want to do with this component.

packages/dialtone-css/lib/build/less/components/item-layout.less
juliodialpad: I don't really think that it's necessary to mention the "how" here, mentioning that adding the prop will remove the styling should be enough, but I don't have problem adding it, so I'll add this to the docs.

iropolo: We are adding a class to remove styles from another class, but the other class is still there.
Normally you add a class to add styles, in this case, the class is removing styles and _not directly_(!).

You can have the `unstyle` prop true, but if you manage to remove the class `dt-item-layout--custom` (which is not adding any styles) you will add styles (from `dt-item-layout` which is not adding the same styles initially cos `unstyle` was `true`)

This behaviour can be confusing, so to make it more clear Im suggesting this minor update.
Expose that we are adding a `--custom` class that it need to be there to trigger different styles from `dt-item-layout` class.

Hope this can help to understand a little bit more my suggestion.

packages/dialtone-vue2/components/item_layout/item_layout.mdx
ninamarina: Great, thanks for clarifying that. 
> I can add a prop to the layout to be able to set the gap, but I don't think that's what we want to do with this component.

I don't think so either, maybe when this component is used as a recipe we want to add a gap prop to the parent component.

braddialpad (approved):
Looks good now. No visual changes 👍

iropolo (approved):

PR #165, opened 2024-02-21
fix(popover): reading from undefined ref

ninamarina:
# fix(popover): reading from undefined ref

Issue in Sentry: https://dialpad.sentry.io/issues/4974421022/?project=5269274&referrer=jira_integration

https://media.giphy.com/media/iibH5ymW6LFvSIVyUc/giphy.gif?cid=790b7611vnvfweg99ws8n48qb29ixn6fsdv4xw3q7hzl3o51&ep=v1_gifs_search&rid=giphy.gif&ct=g

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DLT-1559

## :book: Description
I could not repro this in the app or in Storybook, but looking in the code, we were accessing `this.$refs.content.$el?` and `this.$refs.content` might be undefined. Changed this to `this.$refs.content?.$el?` so that the error would not happen anymore.

### DIFF
--- a/packages/dialtone-vue2/components/popover/popover.vue
+++ b/packages/dialtone-vue2/components/popover/popover.vue
@@ -678,7 +678,7 @@ export default {
       ? this.$refs.anchor.getRootNode().querySelector(`#${this.externalAnchor}`)
       : null;
     this.anchorEl = externalAnchorEl ?? this.$refs.anchor.children[0];
-    this.popoverContentEl = this.$refs.content.$el;
+    this.popoverContentEl = this.$refs.content?.$el;
 
     if (this.isOpen) {
       this.initTippyInstance();
--- a/packages/dialtone-vue2/components/popover/popover.vue
+++ b/packages/dialtone-vue2/components/popover/popover.vue
@@ -879,7 +879,7 @@ export default {
 
     focusInitialElement () {
       if (this.initialFocusElement === 'dialog') {
-        this.$refs.content.$el.focus();
+        this.$refs.content?.$el.focus();
       }
       // find by ID
       if (this.initialFocusElement.startsWith('#')) {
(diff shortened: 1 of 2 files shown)


packages/dialtone-vue2/components/popover/popover.vue
```diff
@@ -879,7 +879,7 @@ export default {
 
     focusInitialElement () {
       if (this.initialFocusElement === 'dialog') {
-        this.$refs.content.$el.focus();
+        this.$refs.content?.$el.focus();
```
braddialpad: Should we put it on $el too to be safe 

`this.$refs.content?.$el?.focus();`

packages/dialtone-vue2/components/popover/popover.vue
braddialpad, line 1: Same on vue 3?

braddialpad (commented):
I think the real problem here is a timing issue, it's trying to focus the content before it's available in the dom. It could even be an issue with the way they are using the component in product, but this should at least fix the error.

packages/dialtone-vue2/components/popover/popover.vue
ninamarina: Fixed in [34a01f6](https://github.com/dialpad/dialtone/pull/165/commits/34a01f69cc069d42f25953994df8443f5699f3a0)

packages/dialtone-vue2/components/popover/popover.vue
ninamarina: Also fixed in [34a01f6](https://github.com/dialpad/dialtone/pull/165/commits/34a01f69cc069d42f25953994df8443f5699f3a0)

braddialpad (approved):

PR #169, opened 2024-02-22
fix(tooltip): remove validators from directive

braddialpad:
# fix(tooltip): remove validators from directive

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1557

## :book: Description

Removed the custom validators on the tooltip directive, as they are handled by the prop validators on the component

## :bulb: Context

This was causing issues in product with warnings triggering when tooltip message is undefined, which is actually a valid case it should just not show a tooltip. I think we did this initially because the validator was throwing an error instead of a warning on incorrect prop validation, but that is actually what it's supposed to do. If you pass an incorrect placement the tooltip should not render and it should be an error.

## :pencil: Checklist

For all PRs:

For all Vue changes:


iropolo (approved):
Cool, warnings gone.

PR #179, opened 2024-02-28
fix: long dt chip

iropolo:
# Fix long dt chip

Jira issue: https://dialpad.atlassian.net/browse/DP-77745

Issue: 
- `dt-chip` did not handled correctly long texts.
- `combobox` did not handled correctly long items

Fix: 
Since these elements were in position absolute and it loses the parent width, we set the `max-width` property to 99% to catch it again.

`99%` was the value because the ellipsis space and `x` button, with `100%` it will be a little bit off:

image

Why `max-width` and not `width`, because `width` can be instrusive on the style and generate unexpected behaviours.

Then in the `button` element, we add `max-width: 100%` to recapture the 100% of the 99% of our parent positioned absolut element. 

To fix the combobox, it just was needed to add the `flex-shrink: 0` to the right element to ensure that do not shrink or grow, maintaining their min-width without being pushed out.

-------

PREVIOUS:
image

image

AFTER:
image

image


iropolo:
Adding @modialpad to review, since he capture this bug.

juliodialpad (dismissed):
I'll say this is a good and quick solution, but we're loosing the spacing on the right, I know it can be difficult to achieve that as I had to deal with this component previously.

Please try to match the padding on the left and right.

image

francisrupert:
Hmm, this is quite an edge-case, though it's bound to occur. Chips are meant to be terse, rarely more than a few words — but naturally that can't always be controlled.

Maybe have a reasonable max-width and truncate, paired with a Tooltip to reveal its entire label. https://dialtone.dialpad.com/components/chip.html#truncated

iropolo:
Huh ye, about losing space on the right, IMO not relevant unless design goes strict with it since this is a not normal case.

I would preffer to go with a fixed `max-with`, sth like 30% maybe? and then truncate it and use a tooltip to show the entire text as @francisrupert says.

I will update it and lets see if it is okay.

iropolo:
Hm... having some troubles to add tooltip, since truncate comes from CSS, so.. hard to catch it.
fixed `d-chip__text` class in `dt-chip`.
Im using `v-dt-tooltip` directive to show it. Maybe we can have a prop in `combobox_multi_select.vue` that allow the user to show a tooltip on the chip, sth like "showTooltip" `true`... it will add tooltip to all the elements, not just the truncated ones.


image


Another solution that would be cool to have is not doing this cut with ellipsis and directly shows all the text:

image

But it will not work cos since it is positionated absoluted with the input, it lost the correct height, so we should deal with the input.

image

------
Definetly need a research to handle this situations.

Since this is an exceptional case that it is not bombing our bug triage, I would go with the current solution if it fits for @modialpad and move forward creating an investigate ticket to tackle this deeply.

braddialpad:
> Another solution that would be cool to have is not doing this cut with ellipsis and directly shows all the text

I think the current implementation is fine.. we probably never want to show items that huge.

packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
```diff
@@ -620,13 +620,15 @@ export default {
   margin-left: var(--dt-space-200);
   margin-right: var(--dt-space-200);
   padding-left: var(--dt-space-100);
+  max-width: 99%;
```
braddialpad: max-width: calc(var(--dt-size-100-percent) - whatever_needed_padding)

packages/dialtone-css/lib/build/less/components/chip.less
```diff
@@ -41,6 +41,7 @@
   transition-timing-function: var(--ttf-in-out);
   transition-duration: var(--td200);
   transition-property: background-color;
+  max-width: 100%;
```
braddialpad: var(--dt-size-100-percent)

packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
```diff
@@ -620,13 +620,15 @@ export default {
...
 }
 
 .combobox__chip {
   margin-top: var(--dt-space-300);
   margin-left: var(--dt-space-200);
   margin-right: var(--dt-space-200);
   z-index: var(--zi-base1);
+  max-width: 100%;
```
braddialpad: var(--dt-size-100-percent)

braddialpad (approved):

PR #181, opened 2024-02-28
fix: avatar image alt repeated

juliodialpad:
# Fix Avatar's repeated alt text

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1590

## :book: Description

- Updated avatar's image alt attribute to remove setting its default value to fullName
- Added aria-label to avatar icon
- Removed button aria-label, it's not actually required as we're using one of these values:
  - the text on the avatar
  - the icon aria-label
  - the image alt text
- Added avatarAlt props to components that need to pass it to avatar.
- Updated tests
- Updated Vue 3 version of the avatar, it was missing some watchers

## :bulb: Context

DtAvatar component used fullName as a fallback value to imageAlt, this was causing a11y issues as the name
was repeated as alt attribute and text.

## :pencil: Checklist

For all PRs:

For all Vue changes:

## :crystal_ball: Next Steps

This will require a migration on product as new required props has been added


packages/dialtone-vue2/components/avatar/avatar.vue
```diff
@@ -335,15 +333,13 @@ export default {
...
-      if (!this.clickable) return undefined;
-
-      return this.fullName || this.imageAlt || this.$attrs['aria-label'];
-    },
   },
 
   watch: {
+    clickable () {
+      this.validateProps();
+    },
```
braddialpad: Does this need immediate?

packages/dialtone-vue2/components/avatar/avatar.vue
```diff
@@ -416,8 +412,8 @@ export default {
     },
 
     validateProps () {
-      if (this.imageSrc && !(this.fullName || this.imageAlt)) {
-        throw new Error('full-name or image-alt must be set if image-src is provided');
+      if (this.imageSrc && !this.imageAlt) {
+        throw new Error('image-alt must be set if image-src is provided');
```
braddialpad: Is this actually what we want? The case we're trying to fix here is that the image-alt is repeated as text, like in contact row: https://dialtone.dialpad.com/vue/index.html?path=/story/recipes-leftbar-contact-row--default 

It already states the name in text, meaning it should have no image-alt even when image-src is provided.

packages/dialtone-vue2/components/avatar/avatar.vue
juliodialpad: No, the component validates the props on mounted so we're ok there

packages/dialtone-vue2/components/avatar/avatar.vue
juliodialpad: Hmm, I cannot reproduce the issue, maybe I'm misunderstanding what the issue is here.

Previously we were defaulting the avatar image-alt to the name, so I removed that. Now, you need to provide an image alt that should actually be different from the name, right? as you're describing the image, not the actual name of the contact.

braddialpad: It should just have no image-alt at all if it's already described by the name text.

packages/dialtone-vue2/recipes/leftbar/contact_row/contact_row.vue
```diff
@@ -16,7 +16,7 @@
       <dt-avatar
         :full-name="name"
         :image-src="avatarSrc"
-        :image-alt="avatarAlt"
+        image-alt=""
```
braddialpad, line 19: So we are saying that `image-alt=""` and `:image-alt="undefined"` are two different things? We set `image-alt=""` when the alt text is unnecessary because it's set set by another description?

juliodialpad: Yes, in fact it is.

If you set alt="undefined" or don't set it, it will raise an a11y warning.

If set to empty string, and there's text around, it will not raise any warning.

juliodialpad: https://help.siteimprove.com/support/solutions/articles/80000863904-accessibility-image-alt-text-best-practices

braddialpad: Great, thanks for the article. I'm okay with this solution, but let's write in the comment for image-alt the difference between blank and undefined because you'd never know that if you didn't already read it.

braddialpad (approved):

packages/dialtone-vue2/recipes/leftbar/contact_row/contact_row.vue
juliodialpad: Comment updated

PR #183, opened 2024-02-29
fix(feed-item-row): accessibility error

ninamarina:
# fix(feed-item-row): accessibility error

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DLT-1592

## :book: Description
Fixes the a11y violation `aria-allowed-attr` that was disabled in Storybook.
When enabling the rule, I could repro in vue 3.
The error was happening when the role is `listitem`, since it's assigning an `aria-selected` which is not allowed for that role. Even if the value is false, the attribute is being assigned in vue 3 anyways, so to fix it I am setting it to `undefined` if the rol doesn't allow it, considering the roles listed here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected

## Before
image

## Now
image

## :pencil: Checklist

For all PRs:

For all Vue changes:

### DIFF
--- a/packages/dialtone-vue2/components/list_item/list_item.test.js
+++ b/packages/dialtone-vue2/components/list_item/list_item.test.js
@@ -7,7 +7,7 @@ const MOCK_CLICK_STUB = vi.fn();
 const baseProps = {
   id: 'dt-item',
   navigationType: LIST_ITEM_NAVIGATION_TYPES.ARROW_KEYS,
-  role: 'tab',
+  role: 'option',
 };
 const baseListeners = {
   click: MOCK_CLICK_STUB,
--- a/packages/dialtone-vue2/components/list_item/list_item.test.js
+++ b/packages/dialtone-vue2/components/list_item/list_item.test.js
@@ -7,6 +7,7 @@ const MOCK_CLICK_STUB = vi.fn();
 const baseProps = {
   id: 'dt-item',
   navigationType: LIST_ITEM_NAVIGATION_TYPES.ARROW_KEYS,
+  role: 'tab',
 };
 const baseListeners = {
   click: MOCK_CLICK_STUB,
--- a/packages/dialtone-vue2/components/list_item/list_item.vue
+++ b/packages/dialtone-vue2/components/list_item/list_item.vue
@@ -9,7 +9,7 @@
     }]"
     :tabindex="isFocusable ? 0 : -1"
     :role="role"
-    :aria-selected="isHighlighted"
+    :aria-selected="hasAriaSelected"
     v-on="listItemListeners"
   >
     <component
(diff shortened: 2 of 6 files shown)


packages/dialtone-vue2/components/list_item/list_item.test.js
```diff
@@ -7,6 +7,7 @@ const MOCK_CLICK_STUB = vi.fn();
 const baseProps = {
   id: 'dt-item',
   navigationType: LIST_ITEM_NAVIGATION_TYPES.ARROW_KEYS,
+  role: 'tab',
```
braddialpad: Why role tab?

packages/dialtone-vue2/components/list_item/list_item.vue
```diff
@@ -48,6 +48,8 @@ import utils from '@/common/utils';
 import { DtIcon } from '@/components/icon';
 import { DtItemLayout } from '@/components/item_layout';
 
+const ARIA_SELECTED_ALLOWED_ROLES = ['gridcell', 'option', 'row', 'tab'];
```
braddialpad: While these are all the valid roles for aria-selected, I don't really think we should allow the list item component to be anything other than roles `listitem` (non-interactive) or `option` (interactive)

packages/dialtone-vue2/components/list_item/list_item.vue
```diff
@@ -9,7 +9,7 @@
     }]"
     :tabindex="isFocusable ? 0 : -1"
     :role="role"
-    :aria-selected="isHighlighted"
+    :aria-selected="hasAriaSelected"
```
braddialpad: maybe can be just `:aria-selected="role === 'option' ? isHighlighted : undefined"` or put in computed, your call

packages/dialtone-vue2/components/list_item/list_item.test.js
ninamarina: It is so that the tests that assert over the `aria-selected` attribute make sense and pass. I can use `option` is that's better

packages/dialtone-vue2/components/list_item/list_item.vue
ninamarina: Great, in that case I can also add a validator

packages/dialtone-vue2/components/list_item/list_item.test.js
braddialpad: yeah use option

braddialpad (approved):
Thanks!

PR #186, opened 2024-03-04
feat: dialtone stylelint plugin - no mixins rule

braddialpad:
# feat: dialtone stylelint plugin - no mixins rule

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1260

## :book: Description

- Added new custom 'dialtone-stylelint-plugin' package which we can add custom dialtone specific stylelint rules to if desired.
- Added initial rule "no-mixins" which detects LESS mixins and warns developers not to use / change them

## :bulb: Context

We wanted a linter rule that could detect LESS mixin usage in the product. Not really anything in existing stylelint rules that could do this, so added a custom rule, and a new package for future custom dialtone rules.

This is a new package and it has been intentionally structured very similar to our eslint package. Most of the configuration was taken from that package but there were some changes that had to be made.

Run the script "test" on the new package to run tests.

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

Add the rule to Dialpad and test that it works as expected.

## :link: Sources

https://stylelint.io/developer-guide/plugins


packages/stylelint-plugin-dialtone/lib/rules/no-mixins.js
```diff
@@ -0,0 +1,52 @@
...
+    });
+
+    if (!validOptions) return;
+
+    // This iterates through one selector at a time, so you don't have to worry about checking for nested selectors.
+    root.walkRules((ruleNode) => {
+      const { source } = ruleNode;
+
+      // any line that starts with a period and ends with a semicolon is a LESS mixin.
+      const regex = /\.[\S]+;/gm;
```
juliodialpad, line 32: Looks like there's a bunch of different ways to create and call mixins...

I guess we only want to detect dialtone mixins only so the regex here will work for .mixin; and .mixin(); 

Don't you think it's too generic that it could potentially detect other stuff as mixins?

https://lesscss.org/features/#mixins-feature

juliodialpad (approved):
Looking good! just a small comment related to the regex being too generic maybe.

packages/stylelint-plugin-dialtone/lib/rules/no-mixins.js
braddialpad: most of those are for defining mixins, we just want to find mixin usage, which I think this will cover. Maybe we should add `d-` into the regex though so it only finds dialtone mixins.

juliodialpad: No, I think we want to remove all mixins usage, so I'll prefer to find all of them instead of just dialtone's

braddialpad: that is true haha

braddialpad: only one I see there that might not work is the `#outer.inner();` format, but I don't think I've ever even seen one used this way in product.

PR #188, opened 2024-03-05
docs: add search to tokens page

ninamarina:
# docs: add search to tokens page

Obligatory GIF

- [ ] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [x] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DLT-1505

## :book: Description
Adds search to the tokens page.
You can search by name, category or description.
Adds a dynamic table of contents, using vuepress toc component.

## :pencil: Checklist

For all PRs:

For all CSS changes:

## :crystal_ball: Next Steps
Search by keywords instead of description: https://dialpad.atlassian.net/browse/DLT-1607

search-tokens

### DIFF
--- a/apps/dialtone-documentation/docs/.vuepress/baseComponents/AllTokens.vue
+++ b/apps/dialtone-documentation/docs/.vuepress/baseComponents/AllTokens.vue
@@ -2,7 +2,6 @@
   <div class="tokens-bar">
     <dt-input
       id="search-input"
-      ref="searchRef"
       v-model="searchInput"
       aria-label="Search tokens"
       placeholder="Search Tokens / Value / Keyword"
@@ -45,24 +44,36 @@
       @change="setTheme"
     />
   </div>
-  <div v-for="category in categories" :key="category">
+  <div v-for="category in Object.keys(filteredTokens)" :key="category">
     <slot :name="category">
       <!-- slot for anchor header -->
     </slot>
-    <token-table :category="category" :format="format" :theme="theme" :search-criteria="searchCriteria" />
+    <token-table :category="category" :tokens="filteredTokens[category]" />
+  </div>
+  <div
+    v-if="!hasSearchResults"
+    class="d-d-flex d-fl-center d-p16 d-gg4 d-fc-tertiary d-fs-300"
+  >
+    <span>No results found for</span>
+    <strong class="d-fw-semibold">
+      &OpenCurlyDoubleQuote;{{ searchCriteria }}&CloseCurlyDoubleQuote;
+    </strong>
   </div>
 </template>
 
 <script setup>
+import * as tokensJson from '@dialpad/dialtone-tokens/dist/doc.json';
+import { getComposedTypographyTokens, getComposedShadowTokens } from '../common/token-utilities';
 import TokenTable, { CATEGORY_MAP, FORMAT_MAP, THEMES } from '@baseComponents/TokenTable.vue';
-import { computed, ref } from 'vue';
+import { computed, ref, onBeforeMount } from 'vue';
 import { debounce } from '../common/utilities';
 
 const format = ref('CSS');
 const theme = ref('light');
-const categories = Object.keys(CATEGORY_MAP);
 const searchInput = ref(null);
 const searchCriteria = ref(null);
+const processedTokens = {}; // is set beforeMount and never changes
+const filteredTokens = ref({}); // same as processedTokens but filtered by format, theme and search
 
 const formatSelectMenuOptions = computed(() => {
   return Object.keys(FORMAT_MAP).map((item) => {
@@ -78,7 +89,25 @@ const setTheme = (newTheme) => {
   theme.value = newTheme;
 };
 
-const setSearchCriteria = () => (searchCriteria.value = searchInput.value);
+const setSearchCriteria = () => {
+  searchCriteria.value = searchInput.value;
+
+  if (!searchCriteria.value) {
+    filteredTokens.value = structuredClone(processedTokens[format.value][theme.value]);
+    return;
+  }
+  const newTokens = {};
+  const regex = new RegExp(searchCriteria.value, 'i');
+  Object.keys(CATEGORY_MAP).forEach((category) => {
+    const results = processedTokens[format.value][theme.value][category].filter((token) => {
+      if (regex.test(category)) return true;
+      const { name, tokenValue, description } = token;
+      return regex.test(name) || regex.test(tokenValue) || regex.test(description);
+    });
+    if (results.length) newTokens[category] = results;
+  });
+  filteredTokens.value = structuredClone(newTokens);
+};
 
 const searchToken = () => {
   debounce(() => setSearchCriteria());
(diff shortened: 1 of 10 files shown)


yorbi-dp:
Super stoked for this! 🔥

yorbi-dp (changes requested):
Thank you, Nina! This is looking great. Just a couple of comments related to the expected results when searching:

- When searching for multiple word token names, I expect to get the token in the results. For example, searching `foreground primary` should bring everything that has `foreground primary` in that order, such as `var(--dt-color-foreground-primary)`.
- When searching for full token names, I'm not getting any results. See the example.
Screenshot 2024-03-05 at 4 00 15 PM
- I'm still not 100% convinced about using the description to search because it's bringing results I'm not expecting. For example, searching surface also brings up foreground colors because it's in the description. Probably not for this ticket, but can we add keywords instead of using the description for each token?

These are all my comments for now. Thank you Nina!

ninamarina:
> Thank you, Nina! This is looking great. Just a couple of comments related to the expected results when searching:
> 
> * When searching for multiple word token names, I expect to get the token in the results. For example, searching `foreground primary` should bring everything that has `foreground primary` in that order, such as `var(--dt-color-foreground-primary)`.

Fixed in last commit:
image

Notice that is also brings `var(--dt-color-foreground-tertiary)` as a result because it has "primary" in the description. That would be fixed when we look for keywords instead of description

> * When searching for full token names, I'm not getting any results. See the example.

Fixed in last commit:
image

> - I'm still not 100% convinced about using the description to search because it's bringing results I'm not expecting. For example, searching surface also brings up foreground colors because it's in the description. Probably not for this ticket, but can we add keywords instead of using the description for each token?

I totally agree this would be better. Will look into where can we add this keywords.

yorbi-dp (changes requested):
This is looking awesome, Nina! Thanks.

One improvement: when searching for styles that are in the Figma structure, e.g., `typography/body/compact/small`, I would expect to get the token that is represented by that style, `var(--dt-typography-body-compact-small)`.
image
Expected result:
image

yorbi-dp (approved):
This is looking great Nina! Thank you.

apps/dialtone-documentation/docs/.vuepress/baseComponents/AllTokens.vue
```diff
@@ -1,5 +1,43 @@
...
+      type="text"
+      autocomplete="off"
+      @keyup="searchToken"
+    >
+      <template #leftIcon>
+        <dt-icon name="search" size="300" />
+      </template>
+      <template #rightIcon>
+        <dt-button
+          v-if="!isSearchEmpty"
```
iropolo: `isSearchEmpty` is being used once and using `!`

Could we call it sth like:
`isSearchNotEmpty`
`const isSearchNotEmpty = computed(() => searchInput.value && searchInput.value.trim().length > 0);`

apps/dialtone-documentation/docs/.vuepress/baseComponents/AllTokens.vue
```diff
@@ -14,50 +52,243 @@
...
       @change="setTheme"
     />
-  </dt-stack>
-  <div v-for="category in categories" :key="category">
-    <slot :name="category">
-      <!-- slot for anchor header -->
-    </slot>
-    <token-table :category="category" :format="format" :theme="theme" />
+  </div>
+  <div v-for="category in Object.keys(filteredTokens)" :key="category">
```
iropolo: Seems like we are using `Object.keys(filteredTokens)` in many places.

Could we create a 

`const filteredTokensKeys = computed(() => Object.keys(filteredTokens.value));`

and replace them.

apps/dialtone-documentation/docs/.vuepress/baseComponents/AllTokens.vue
braddialpad: Maybe just `hasSearchTerm`. I usually try to avoid variable names with 'Not' in them because it's confusing.

apps/dialtone-documentation/docs/.vuepress/baseComponents/AllTokens.vue
```diff
@@ -14,50 +52,243 @@
...
+      tabindex="-1"
+    >
+      <a
+        :href="`#${category}`"
+        class="header-anchor d-link d-docsite--link d-link"
+      >#</a>
+      {{ category }}
+    </h2>
+    <p class="dialtone-intro">
+      {{ categoryIntroAndDesc[category]?.intro }}
```
iropolo: huh, I would like to void use inline expression in templates.
Instead of this what if we use a method like:

```
const getCategoryIntro = (category) => {
  return categoryIntroAndDesc[category]?.intro;
};
```

```
<p
      v-if="getCategoryIntro(category)"
      class="dialtone-intro">
      {{ getCategoryIntro(category) }}
    </p>
```

Then we avoid to create an empty <p> element.

apps/dialtone-documentation/docs/.vuepress/baseComponents/AllTokens.vue
```diff
@@ -14,50 +52,243 @@
...
+    </p>
+    <div v-html="categoryIntroAndDesc[category]?.description" />
```
iropolo: Here also is creating an empty `<div>` and it has an inline expression, we can replicate what we did previously with a method.

We are calling these method twice, we can avoid it with a computed property but since this is not reactive I dont see it necessary, and the method is pretty straightforward.

apps/dialtone-documentation/docs/.vuepress/baseComponents/AllTokens.vue
```diff
@@ -14,50 +52,243 @@
...
+    <token-table :category="category" :tokens="filteredTokens[category]" />
+  </div>
+  <div
+    v-if="!hasSearchResults"
```
iropolo: Here also we are using `hasSearchResults` just once and with a `!`.
I will suggest to update the name to sth like:

`noSearchResults`
`const noSearchResults = computed(() => Object.keys(filteredTokens.value).length === 0);`

apps/dialtone-documentation/docs/.vuepress/baseComponents/AllTokens.vue
```diff
@@ -14,50 +52,243 @@
...
+    <span>No results found for</span>
+    <strong class="d-fw-semibold">
+      &OpenCurlyDoubleQuote;{{ searchCriteria }}&CloseCurlyDoubleQuote;
+    </strong>
   </div>
 </template>
 
-<script>
+<script setup>
+import * as tokensJson from '@dialpad/dialtone-tokens/dist/doc.json';
```
iropolo: `import tokensJson from '@dialpad/dialtone-tokens/dist/doc.json';`
This should be the same.

apps/dialtone-documentation/docs/.vuepress/baseComponents/AllTokens.vue
```diff
@@ -14,50 +52,243 @@
...
+import { getComposedTypographyTokens, getComposedShadowTokens } from '../common/token-utilities';
 import TokenTable, { CATEGORY_MAP, FORMAT_MAP, THEMES } from '@baseComponents/TokenTable.vue';
```
iropolo: Huh, my idle is blaming a little bit this import. 
`import TokenTable, { CATEGORY_MAP, FORMAT_MAP, THEMES } from '../baseComponents/TokenTable.vue';`
it works well like this and I can go with the definitions.

Beyond that, those exported constant should be separated from the file. 
There is one `apps/dialtone-documentation/docs/.vuepress/common/constants.js`.
But definetly not part of the scope of the ticket.

apps/dialtone-documentation/docs/tokens/index.md
```diff
@@ -3,52 +3,11 @@ title: Tokens
 description: Multi-platform values that make up Dialtone's design language
 ---
 
-## Introduction
-
 Design tokens are all the defined values that are part of the design system, such as color, spacing, typography, and more. Design tokens are published in many different forms so they can be consumed on all platforms (CSS, LESS, Android, iOS, etc).
 
 See the [dialtone-tokens repository](https://github.com/dialpad/dialtone-tokens) for usage info.
```
iropolo: mmm.. should be this I guess `https://github.com/dialpad/dialtone/pkgs/npm/dialtone-tokens` or.. directly to `https://github.com/dialpad/dialtone` monorepo.

apps/dialtone-documentation/docs/.vuepress/theme/components/Page.vue
```diff
@@ -101,4 +102,9 @@ const includeToc = computed(() => {
   const headers = usePageData().value.headers;
   return headers?.length > 0;
 });
+
+const tokensPageClass = computed(() => {
+  if (route.path.includes('tokens')) return 'tokens-page';
+  return '';
```
iropolo, line 108: Im guessing this will be escalable soon.

We will think to have this dynamically sth like `pageDynamicallyClasses`
Then it will handle these classes.

iropolo (approved):
Looks good, works well, happy to have this integration!!, nice job

apps/dialtone-documentation/docs/.vuepress/baseComponents/AllTokens.vue
iropolo: Agree, sure, it can be confusing sometimes.

apps/dialtone-documentation/docs/.vuepress/theme/components/Page.vue
ninamarina: We can think about that in the case that some other page needs a custom layout like this one

apps/dialtone-documentation/docs/tokens/index.md
braddialpad: The subrepo for tokens still has it's own readme

https://github.com/dialpad/dialtone/tree/staging/packages/dialtone-tokens

ninamarina:
Thanks @iropolo, just implemented your suggestions

braddialpad (approved):

PR #195, opened 2024-03-08
feat(tooltip): add appendTo prop

braddialpad:
# feat(tooltip): add appendTo prop

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1542

## :book: Description

Added appendTo prop to tooltip

## :bulb: Context

My prior addition of the appendTo prop had a bunch of complicated code in it to make it reactive. This is likely what caused the bug in ubervoice. This is a much simpler method which only sets appendTo on initial render and should be much safer than my previous PR. Additionally I don't really think there is much of a valid use case to have appendTo be reactive anyways.

## :pencil: Checklist

For all PRs:

For all Vue changes:

## :crystal_ball: Next Steps

Test in product to make sure the prior error is no longer happening.

## :link: Sources

https://atomiks.github.io/tippyjs/v6/all-props/#appendto

### DIFF
--- a/packages/dialtone-vue2/components/tooltip/tooltip.stories.js
+++ b/packages/dialtone-vue2/components/tooltip/tooltip.stories.js
@@ -66,6 +66,12 @@ export const argTypesData = {
     },
   },
 
+  appendTo: {
+    control: {
+      type: null,
+    },
+  },
+
   enabled: {
     type: 'boolean',
   },
--- a/packages/dialtone-vue2/components/tooltip/tooltip.vue
+++ b/packages/dialtone-vue2/components/tooltip/tooltip.vue
@@ -55,6 +55,9 @@ import {
   TOOLTIP_STICKY_VALUES,
   TOOLTIP_DELAY_MS,
 } from './tooltip_constants.js';
+import {
+  POPOVER_APPEND_TO_VALUES,
+} from '../popover/popover_constants';
 import { getUniqueString } from '@/common/utils';
 import { DtLazyShow } from '@/components/lazy_show';
 import {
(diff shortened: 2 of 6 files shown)


iropolo (approved):
Great, thanks

packages/dialtone-vue2/components/tooltip/tooltip.stories.js
```diff
@@ -66,6 +66,12 @@ export const argTypesData = {
     },
   },
 
+  appendTo: {
+    control: {
+      type: null,
```
ninamarina, line 71: Why null? Don't we want to test this in storybook?

packages/dialtone-vue2/components/tooltip/tooltip.vue
```diff
@@ -479,6 +497,7 @@ export default {
       return {
         contentElement: this.$refs.content.$el,
         allowHTML: true,
+        appendTo: this.appendTo === 'body' ? this.anchor?.getRootNode()?.querySelector('body') : this.appendTo,
```
ninamarina, line 500: Why is the default set to `this.anchor?.getRootNode()?.querySelector('body')` instead of `document.body`?

packages/dialtone-vue2/components/tooltip/tooltip.stories.js
braddialpad: yeah because it's non-reactive changing it live in storybook will do nothing.

packages/dialtone-vue2/components/tooltip/tooltip.vue
braddialpad: it's for shadow dom reasons. When within a shadow dom we want it to get the shadow dom body, rather than the root body.

packages/dialtone-vue2/components/tooltip/tooltip.stories.js
ninamarina: ooh I see

ninamarina (approved):

PR #199, opened 2024-03-11
fix: update missing props

iropolo:
# Update missing props

Jira ticket: https://dialpad.atlassian.net/browse/DLT-1595

Updating props in components:

- Banner
- Toast
- Notice

With following props:

- hideIcon
- hideAction
- hideClose

--------

Screenshot 2024-03-11 at 12 58 10

image

### DIFF
--- a/packages/dialtone-vue2/components/banner/banner.vue
+++ b/packages/dialtone-vue2/components/banner/banner.vue
@@ -35,6 +35,7 @@
         <slot />
       </dt-notice-content>
       <dt-notice-action
+        :hide-action="hideAction"
         :hide-close="hideClose"
         :close-button-props="closeButtonProps"
         :visually-hidden-close="visuallyHiddenClose"
@@ -139,6 +140,7 @@ export default {
 
     /**
      * Hides the close button from the notice
+     * @values true, false
      */
     hideClose: {
       type: Boolean,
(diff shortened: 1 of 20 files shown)


packages/dialtone-vue2/components/banner/banner.vue
braddialpad, line 1: Noticed the height of the banner changes when hiding the action. That probably shouldn't happen.



Might want some design guidance on the correct way to fix this.

iropolo: Huh, let's maintain the previous height.... cc. @yorbi-dp

yorbi-dp: The height can adjust depending on the content of the banner because it doesn't change after being rendered; it doesn't push content underneath as it renders with all its content. But a quick update can be made: in the design decision, the actions/buttons are sized as 'sm' (small), but in the code, it's currently set to 'md' (medium). Thanks!

iropolo: Ye, it goes with the default value, that is `md`, I will update it to `sm` and it will be fix in doc, but,
if someone pass a button `md` or `xl` then this will be the expected behaviour.

I guess it should be ok.

iropolo:
Adding visual test to update screenshots.
We update button action size from banner component from `md` to `sm`.

braddialpad (approved):

yorbi-dp (approved):
LGTM!

PR #201, opened 2024-03-11
fix: message input select media

juliodialpad:
# Fix select-media event on DtRecipeMessageInput

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

No Jira

## :book: Description

Changed the value emitted by `select-media` event to emit the files object instead of just the file names.

## :bulb: Context

@bianca-artola-dialpad reported that message input component was only emitting file name and they need more information, emitting the whole object should fix their issues.

## :pencil: Checklist

For all PRs:

For all Vue changes:

## :link: Sources

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file

### DIFF
--- a/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
+++ b/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
@@ -597,8 +597,8 @@ export default {
       this.$refs.messageInputImageUpload.$refs.input.click();
     },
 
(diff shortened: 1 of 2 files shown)


braddialpad (approved):

bianca-artola-dialpad:
can we do the same when dropping images?  we already have add-media but is just returning the path. I think we will also need to add something to paste images

PR #205, opened 2024-03-13
docs: add composition tokens tooltips

ninamarina:
# docs: add composition tokens tooltips

Obligatory GIF

- [x] Documentation

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1506

## :book: Description
Adds tooltips with the value for each separate tokens for composition tokens.
To test it, look for the first tokens in the list for Typography and Shadow.

## :pencil: Checklist

For all PRs:

tokens-tooltips

### DIFF
--- a/apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
+++ b/apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
@@ -55,7 +55,23 @@
         <td
           class="d-code--sm"
         >
-          <div class="d-wmx264">
+          <div v-if="isCompositionToken(tokenValue)" class="d-wmx264">
+            <span v-for="value in tokenValue" :key="value">
+              <span v-if="valueIsDivided(value)">
+                <span v-dt-tooltip="getTokenValue(getFirstValue(value))">
+                  {{ getFirstValue(value) }}
+                </span>
+                /
+                <span v-dt-tooltip="getTokenValue(getSecondValue(value))">
+                  {{ getSecondValue(value) }}&nbsp;
+                </span>
+              </span>
+              <span v-else v-dt-tooltip="getTokenValue(value)">
+                {{ value }}&nbsp;
+              </span>
+            </span>
+          </div>
+          <div v-else class="d-wmx264">
             {{ tokenValue }}
           </div>
         </td>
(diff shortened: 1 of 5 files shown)


packages/dialtone-css/lib/build/less/components/tooltip.less
```diff
@@ -54,6 +54,7 @@
     line-height: var(--tooltip-line-height);
     letter-spacing: calc(var(--dt-space-100) * 0.25);
     text-align: center;
+    word-wrap: break-word;
```
braddialpad: hmm should we be doing this for all tooltips in the library or should it just be specific to the displaying tokens case?

apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
braddialpad, line 1: For some reason nothing is happening in the deploy preview when I highlight a composition token

packages/dialtone-css/lib/build/less/components/tooltip.less
ninamarina: This doesn't affect tooltips with normal long text, but it does prevent the text being cut when there's a long text without spaces, so I think it's ok to add it to all tooltips.

ninamarina: I think it's safe doing it for all, because it doesn't affect normal long text with spaces. I cuts the word only when the text reaches it's max width with a long word without spaces, instead of overflowing. And I think that's what we want in all cases

apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
ninamarina: The deploy seems to be using an old version of Dialtone, because it doesn't include the last changes from Francis either. I just merged the last staging, but the preview is still old 🤔

ninamarina: Nvm, I am seeing the latest changes now, but still the tooltip not showing. Looking into it

braddialpad: It might have broke due to the recent changes we made to composition tokens

ninamarina: I think it's related to this: https://dialpad.atlassian.net/browse/DLT-1567
The directive is not showing in vue 3, it does locally, but not in the published version.

ninamarina:
Moving this one to draft since I am having issues with the tooltip directive

apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
ninamarina: Fixed with the commit from Tico!

packages/dialtone-css/lib/build/less/components/tooltip.less
ninamarina: I finally removed this because I added the spaces to the token value, so I no longer need it

braddialpad (approved):
Nice looks good after the fix. I'll merge this in for you since I'm about to do a release.

PR #206, opened 2024-03-14
fix: add box-sizing, padding, margin to reset

braddialpad:
# fix: add box-sizing, padding, margin to reset

Obligatory GIF

- [x] Fix

## :book: Description

https://dialpad.atlassian.net/browse/DLT-1633

Added box-sizing: border-box, padding 0 and margin 0 to our reset

## :bulb: Context

Finding a lot of spacing inconsistencies between DP1, DP2, and even our storybook. Found that DP1 has margin / padding 0 set on all elements which overrides the default user agent padding/margin. Our storybook and DP2 don't have this set, causing the spacing issues.

Set box-sizing: border-box as well since that's another one we should probably have.

## :pencil: Checklist

For all PRs:

### DIFF
--- a/apps/dialtone-documentation/docs/.vuepress/theme/assets/less/dialtone-docs.less
+++ b/apps/dialtone-documentation/docs/.vuepress/theme/assets/less/dialtone-docs.less
@@ -25,14 +25,6 @@
 //      [1]     Universally set border-box
 //  ----------------------------------------------------------------------------
 
-* {
-  &,
-  &::before,
-  &::after {
-    box-sizing: border-box; // [1]
-  }
-}
-
 html {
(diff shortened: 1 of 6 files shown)


francisrupert (approved):
🤦‍♂️ Good catch. 

Maybe a backlog item to find current in-product use of `class="d-box-border` and determine if they're redundant or still appropriate after this change. 

Meanwhile, I know I've seen `class="d-box-content"` and `box-sizing: content-box` in use, but never investigated why. That's pretty specific, so they're probably intentional — but I do wonder.

braddialpad:
seems to be one spacing issue with the feed item pill after reviewing the visual tests.. looking into.

PR #208, opened 2024-03-14
fix: fix styling of d-link--mention

braddialpad:
# fix: fix styling of d-link--mention

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1509

## :book: Description

Moved `d-link--mention` to dialtone-css instead of vue, and corrected some of the styling

## :bulb: Context

Mention styling was incorrect as shown in the above mentioned jira ticket. 

## :pencil: Checklist

For all PRs:

For all CSS changes:

## :crystal_ball: Next Steps

Bring into product to complete story

### DIFF
--- a/packages/dialtone-css/lib/build/less/components/link.less
+++ b/packages/dialtone-css/lib/build/less/components/link.less
@@ -17,6 +17,8 @@
     --link-color-default: var(--dt-color-link-primary);
     --link-color-default-hover: var(--dt-color-link-primary-hover);
     --link-text-decoration: underline;
+    --link-padding: 0;
+    --link-background-color: transparent;
 
     position: relative;
     display: inline-flex;
@@ -25,13 +27,13 @@
     justify-content: center;
     box-sizing: border-box;
     margin: 0;
-    padding: 0;
+    padding: var(--link-padding);
     color: var(--link-color-default);
     font: inherit;
     text-decoration: var(--link-text-decoration);
     text-underline-offset: calc(var(--dt-size-border-300) - var(--dt-size-border-100));
     text-decoration-thickness: var(--dt-size-border-50);
-    background-color: transparent;
+    background-color: var(--link-background-color);
     border: 0;
     transition-timing-function: var(--ttf-out-quint);
     transition-duration: var(--td200);
@@ -133,16 +135,19 @@
     // Styling specific to mentions such as @brad.paugh. The underline highlighting
     // is reversed compared to a regular link, and it has a light background.
     &--mention {
-        color: var(--dt-color-link-primary);
+        --link-text-decoration: none;
+        --link-padding: 0 var(--dt-space-200);
+        --link-background-color: hsl(var(--dt-color-purple-400-hsl) / 0.1);
+
         line-height: var(--dt-font-line-height-200);
-        text-decoration: none;
-        background-color: hsl(var(--dt-color-purple-400-hsl) / 0.1);
         border-radius: var(--dt-size-radius-200);
 
         &:hover {
-            color: var(--dt-color-link-primary);
-            text-decoration: underline;
-            background-color: hsl(var(--dt-color-purple-400-hsl) / 0.2);
+            --link-text-decoration: underline;
+            --link-background-color: hsl(var(--dt-color-purple-400-hsl) / 0.2);
+
+            text-underline-offset: calc(var(--dt-size-border-300) - var(--dt-size-border-100));
+            text-decoration-thickness: var(--dt-size-border-50);
         }
     }
 }
--- a/packages/dialtone-css/lib/build/less/components/link.less
+++ b/packages/dialtone-css/lib/build/less/components/link.less
@@ -127,4 +127,22 @@
             cursor: not-allowed;
         }
     }
+
+    //  $$  MENTION
+    //  ----------------------------------------------------------------------------
+    // Styling specific to mentions such as @brad.paugh. The underline highlighting
+    // is reversed compared to a regular link, and it has a light background.
+    &--mention {
+        color: var(--dt-color-link-primary);
+        line-height: var(--dt-font-line-height-200);
+        text-decoration: none;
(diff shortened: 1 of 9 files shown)


packages/dialtone-css/lib/build/less/components/link.less
```diff
@@ -127,4 +127,22 @@
...
+
+    //  $$  MENTION
+    //  ----------------------------------------------------------------------------
+    // Styling specific to mentions such as @brad.paugh. The underline highlighting
+    // is reversed compared to a regular link, and it has a light background.
+    &--mention {
+        color: var(--dt-color-link-primary);
+        line-height: var(--dt-font-line-height-200);
+        text-decoration: none;
+        background-color: hsl(var(--dt-color-purple-400-hsl) / 0.1);
```
braddialpad: For some reason `--dt-color-link-primary` doesnt seem to be generating an hsl... using purple 400 for now.

packages/dialtone-vue2/components/rich_text_editor/extensions/mentions/MentionComponent.vue
```diff
@@ -5,7 +5,7 @@
   >
     <dt-link
       kind="inverted"
-      class="dt-link--mention"
+      class="d-link--mention"
```
francisrupert: Will need to be `class="d-link d-link--mention"`, unless we'd prefer to allow `class="d-link--mention` available standalone (slight deviation we sometimes do from "true" BEM), which means you'd have to ensure the base `d-link` styles fold into `--mention` too.

packages/dialtone-css/lib/build/less/components/link.less
```diff
@@ -127,4 +127,22 @@
             cursor: not-allowed;
         }
     }
+
+    //  $$  MENTION
+    //  ----------------------------------------------------------------------------
+    // Styling specific to mentions such as @brad.paugh. The underline highlighting
+    // is reversed compared to a regular link, and it has a light background.
+    &--mention {
+        color: var(--dt-color-link-primary);
```
francisrupert: Add `padding: 0 var(--dt-space-200);` for touch less claustrophic.

packages/dialtone-css/lib/build/less/components/link.less
```diff
@@ -127,4 +127,22 @@
...
+        line-height: var(--dt-font-line-height-200);
+        text-decoration: none;
+        background-color: hsl(var(--dt-color-purple-400-hsl) / 0.1);
+        border-radius: var(--dt-size-radius-200);
+
+        &:hover {
+            color: var(--dt-color-link-primary);
+            text-decoration: underline;
```
francisrupert: Add
```
text-underline-offset: calc(var(--dt-size-border-300) - var(--dt-size-border-100));
text-decoration-thickness: var(--dt-size-border-50);
```

These properties are used for all `<a>`-presenting items. You can find them elsewhere. Should probably be at least _local_ CSS Variables, or maybe Design Tokens later.

packages/dialtone-vue2/components/rich_text_editor/extensions/mentions/MentionComponent.vue
```diff
@@ -5,7 +5,7 @@
   >
     <dt-link
       kind="inverted"
```
francisrupert: Not sure why `inverted` is here. I'd view this `--mention` variant to be a new `kind` value since there wouldn't be a mention variant of `danger`, `warning`, etc.

packages/dialtone-vue2/components/rich_text_editor/extensions/mentions/MentionComponent.vue
francisrupert, line 1: Is **Mention** its own component then?

francisrupert (changes requested):
Documentation backlog item for `.d-link` and `<dt-link>`, I suppose? Or fold it into this PR?

packages/dialtone-vue2/components/rich_text_editor/extensions/mentions/MentionComponent.vue
braddialpad: ah yep thanks, definitely should have d-link

packages/dialtone-vue2/components/rich_text_editor/extensions/mentions/MentionComponent.vue
braddialpad: yeah.. I think that makes a lot more sense.

packages/dialtone-vue2/components/rich_text_editor/extensions/mentions/MentionComponent.vue
francisrupert: And duh, that's what we do for other variants. 
image

packages/dialtone-css/lib/build/less/components/link.less
braddialpad: Looks like this is already set on d-link actually.

francisrupert: However... because it's inverted (mention isn't underlined by default) you have be explicit again. I think...

packages/dialtone-vue2/components/rich_text_editor/extensions/mentions/MentionComponent.vue
braddialpad: nah this component's only purpose is to be passed in to TipTap. It's not meant to be a standalone component.

braddialpad:
Fixes are in, thanks for the review. new type added to dt-link as mentioned.

francisrupert (approved):
Yep, looking great! I see it looking as expected in Vue3 preview. I defer to you if you add its raw HTML example to [doc site](https://dialtone.dialpad.com/deploy-previews/pr-208/components/link.html) as part of this PR or later.

braddialpad:
Docsite updated.

PR #215, opened 2024-03-19
fix: use mixin property to detect mixin

braddialpad:
# fix: use mixin property to detect mixin

Obligatory GIF

- [x] Fix

## :book: Description

Apparently there is a "mixin" property directly on the node in stylelint when it is a mixin, so this is a much better way to detect rather than regex 😄 

## :bulb: Context

A bunch of false positives were happening with the prior regex

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

Attempt product implementation

### DIFF
--- a/packages/stylelint-plugin-dialtone/lib/rules/no-mixins.js
+++ b/packages/stylelint-plugin-dialtone/lib/rules/no-mixins.js
@@ -26,6 +26,7 @@ const ruleFunction = (primary) => {
 
     // This iterates through one selector at a time, so you don't have to worry about checking for nested selectors.
     root.walkAtRules((ruleNode) => {
+      // docs for mixin property used here: https://github.com/shellscape/postcss-less?tab=readme-ov-file#mixins
       if (!ruleNode.mixin) return;
       report({
         result,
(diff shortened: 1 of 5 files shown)


packages/stylelint-plugin-dialtone/lib/rules/no-mixins.js
```diff
@@ -25,20 +25,15 @@ const ruleFunction = (primary) => {
...
-    root.walkRules((ruleNode) => {
-      const { source } = ruleNode;
-
-      // any line that starts with a period and ends with a semicolon is a LESS mixin.
-      const regex = /\.[\S]+;/gm;
-
-      const matches = [...source.input.css.matchAll(regex)];
-      for (const match of matches) {
+    root.walkAtRules((ruleNode) => {
+      if (ruleNode.mixin) {
```
juliodialpad: Nice, there's a check to verify if a node is a mixin! can you please share the docs to read more about it please?

Also, It'd be better to use the early return pattern here, to improve readability.

```suggestion
      if (!ruleNode.mixin) return;
```

juliodialpad (approved):
Thanks! hopefully this works as expected, please share the docs to read more about it and fix the small nit-pick.

packages/stylelint-plugin-dialtone/lib/rules/no-mixins.js
braddialpad: good idea, thanks.

braddialpad: docs here: https://github.com/shellscape/postcss-less?tab=readme-ov-file#mixins

PR #218, opened 2024-03-20
docs: fix search for tokens

ninamarina:
# docs: fix search for tokens

We discovered that for Android and iOS searching for tokens in figma format would not brung any results.
For example, searching for typography/body/md:

**Before**
image

**Now**
image

Obligatory GIF

- [ ] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [x] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket
No Jira

### DIFF
--- a/apps/dialtone-documentation/docs/.vuepress/baseComponents/AllTokens.vue
+++ b/apps/dialtone-documentation/docs/.vuepress/baseComponents/AllTokens.vue
@@ -163,11 +163,10 @@ const filterTokens = () => {
   }
   const newTokens = {};
 
-  // Replace '/' characters for ' ' to allow to search by the name shown in Figma:
-  // typography/body/sm/compact should match var(--dt-typography-body-sm-compact)
-  // or dtTypographyBodySmCompact
-  const searchValues = searchCriteria.value.replaceAll('/', ' ').split(' ');
-  console.log('searchValues: ', searchValues);
+  // Replace '/' and '-' characters for ' ' to allow to search by the name shown in Figma:
+  // typography/label/md-plain should match var(--dt-typography-label-plain)
+  // or dtTypographyLabelMdPlain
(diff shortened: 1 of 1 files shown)


yorbi-dp (approved):
This looks great! Thanks a bunch for the quick fix.

iropolo:
Okay so...

type of searches:
1. typography body md
2. typography/body/md
3. typography-body-md

Before:
CSS: 1,2,3
Android: 1
IOs: 1

Now:
CSS: 1,2,3
Android: 1,2
IOs: 1,2

-------

Seems like there is no `/` or `-` tokens in Android and IOs, but there is `-` tokens on CSS.
What I suggest is remove `/` and just let `1` and `3` method.

CSS: 1,3
Android: 1,3
IOs: 1,3


I see it more naturally. Or, add `-` to Andoroid and IOs too to be consistent.

But this is a design decision probably, I'm not sure why we decide to use `/` to separate words.

There is also a noticed lag on switching to Token page, we should investigate it to improve. Technical improvement or Design Decision by splitting into sections.

yorbi-dp:
> But this is a design decision probably, I'm not sure why we decide to use / to separate words.
Screenshot 2024-03-20 at 12 07 49 PM

This is how Figma currently displays the style names. Hopefully, we will get an update soon to show the actual token name in the Figma UI.

iropolo (approved):

ninamarina:
> There is also a noticed lag on switching to Token page, we should investigate it to improve

Yes, this is because the list is really long. The performance issue is in the render phase, so it's tricky to improve it, but we talked about if we can do some "lazy rendering" and add a loading state. I don't think pagination is possible since we have the table of contents with the anchor links.

PR #221, opened 2024-03-21
fix: warn console error cleanup

sanjeevdialpad:
# Fix Console warn regarding button type

The send button had importance="primary" and type="default" with circle: true, this was causing warning on console due to some invalid combination. Used dialtone classes to achieve the same and mute those console warnings.

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [x] Style (code style changes, not css changes)
- [ ] Test
- [ ] Other (chore)

### DIFF
--- a/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
+++ b/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
@@ -174,12 +174,14 @@
             <dt-button
               data-qa="dt-message-input-send-btn"
               size="sm"
-              :kind="sendButtonKind"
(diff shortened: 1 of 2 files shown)


ninamarina:
I wonder what is the reason for making that an invalid combination

braddialpad:
> I wonder what is the reason for making that an invalid combination

Designers said they didn't want anyone using this particular combination, except in this case 😆

braddialpad (approved):
Thanks!

PR #226, opened 2024-03-21
perf: dialtone-icons tree shaking

juliodialpad:
# Dialtone-icons tree shaking

Obligatory GIF

- [x] Performance Improvement

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1654

## :book: Description

- Added tree-shaking and dynamic import capabilities to dialtone-icons.
- Removed support for CJS, do we still need to support this? I don't think we're using that on any of our apps.
- Moved files to a workspace common directory.
- Added an automated index.js file (gitignored) to dialtone-icons vue2 and vue3 version to function as an entry point.

## :bulb: Context

We're improving our libraries to make them lightweight and tree-shakeable

## :pencil: Checklist

For all PRs:


braddialpad:
> Removed support for CJS, do we still need to support this? I don't think we're using that on any of our apps.

You mean the dialtone cjs dist output right? not internal cjs usage?

Not totally sure if anyone is using it.. I'll check some of of our applications, but yeah it would be nice to remove.

packages/dialtone-vue2/components/icon/icon.vue
```diff
@@ -62,13 +61,25 @@ export default {
...
 
-    icon () {
-      return icons[`./${this.name}.vue`];
+  watch: {
+    name: {
+      immediate: true,
+      async handler (name) {
+        const icon = await import(`/node_modules/@dialpad/dialtone-icons/vue2/dist/components/${name}.js`);
+        this.icon = icon.default;
+      },
```
braddialpad: I remember trying to do a dynamic import within our library before and it errored out when they tried to use it on the application side. Our infrastructure is much different than back then though, so maybe it works now. Just something to look out for when we try this out.

packages/dialtone-icons/package.json
```diff
@@ -6,18 +6,17 @@
...
-      "import": "./dist/dialtone-icon@3.js",
-      "require": "./dist/dialtone-icon@3.cjs"
-    },
-    "./*": "./dist/*"
+    "./vue2": "./vue2/dist/dialtone-icons.js",
+    "./vue3": "./vue3/dist/dialtone-icons.js",
+    "./vue2/*": "./vue2/dist/components/*.js",
+    "./vue3/*": "./vue3/dist/components/*.js",
+    "./svg/*": "./dist/svg/*",
+    "./*.json": "./dist/*.json"
```
braddialpad: Should we still allow `"./*": "./dist/*"` ?

packages/dialtone-icons/package.json
```diff
@@ -65,5 +64,6 @@
     "Icons",
     "Icon Library"
   ],
-  "license": "MIT"
+  "license": "MIT",
+  "sideEffects": false
```
braddialpad, line 71: Can we explain or document what sideEffects are?

packages/dialtone-icons/package.json
```diff
@@ -6,18 +6,17 @@
...
-    },
-    "./vue3": {
-      "import": "./dist/dialtone-icon@3.js",
-      "require": "./dist/dialtone-icon@3.cjs"
-    },
-    "./*": "./dist/*"
+    "./vue2": "./vue2/dist/dialtone-icons.js",
+    "./vue3": "./vue3/dist/dialtone-icons.js",
+    "./vue2/*": "./vue2/dist/components/*.js",
+    "./vue3/*": "./vue3/dist/components/*.js",
```
braddialpad: Hmm so does this mean consumers have to import components like `import DtIcon from '@dialpad/dialtone/vue2/icon'` or the code splitting doesn't work? Maybe will require testing product side, could depend on how the bundler is configured?

packages/dialtone-icons/vue2/vite.config.js
```diff
@@ -1,26 +1,32 @@
...
+const iconEntries = glob.sync('../src/icons/*.vue').reduce((entries, path) => {
+  const entryName = path.replace(/^\.\.\/src\/icons\/(.*)\.vue$/, 'components/$1');
+  entries[entryName] = path;
+  return entries;
+}, {});
 
 export default defineConfig({
   plugins: [vue()],
   build: {
+    outDir: 'dist',
```
braddialpad: dist is the default value isn't it?

juliodialpad:
> > Removed support for CJS, do we still need to support this? I don't think we're using that on any of our apps.
> 
> You mean the dialtone cjs dist output right? not internal cjs usage?
> 
> Not totally sure if anyone is using it.. I'll check some of of our applications, but yeah it would be nice to remove.

Yes, just the cjs output, or maybe there's a way to just generate dialtone-icons.cjs, not all the exports, I'll check if it's possible, if not, I think it's ok to just remove cjs output.

packages/dialtone-vue2/components/icon/icon.vue
juliodialpad: Tested on multiple sample applications and working well so far so I don't expect any mayor issue while testing on product.

packages/dialtone-icons/package.json
juliodialpad: No, `"./svg/*"` and `"./*.json"` are covering the exports on dist and `"./vue2/*"` and `"./vue3/*"` are covering the tree-shake on vue 2 and vue 3 icon versions.

packages/dialtone-icons/package.json
juliodialpad: This is a left-over for testing tree-shake on webpack.

I'm not sure this is working as expected, it's not well documented so I'll remove it as it does nothing right now.

packages/dialtone-icons/package.json
juliodialpad: It's not enforced, consumers can still import from `'@dialpad/dialtone/vue2'`.

Importing from `DtIcon from '@dialpad/dialtone/vue2/icon'` instead of `{ DtIcon } from '@dialpad/dialtone/vue2'` it's a way to enforce the tree-shake behavior regardless of the consumer bundler config.

But that's on the dialtone-vue side, here we're making available to import every icon individually, like `import IconAccessibility from '@dialpad/dialtone-icons/vue2/accesibility'`

packages/dialtone-icons/vue2/vite.config.js
juliodialpad: Yeah, I'll remove it

packages/dialtone-icons/package.json
braddialpad: Okay sweet that's what I was hoping.

packages/dialtone-icons/package.json
braddialpad: "A "side effect" is defined as code that performs a special behavior when imported, other than exposing one or more exports. An example of this are polyfills, which affect the global scope and usually do not provide an export."

hmm don't really think we have to worry about that.

braddialpad: or maybe we do have to set it to false on all our projects or webpack will not tree shake them? kinda seems that way

braddialpad: https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

packages/dialtone-vue2/components/icon/icon.vue
braddialpad: k sounds good.

packages/dialtone-icons/package.json
braddialpad: Seems like dialtone-css needs to be sideEffects: true based on this, since it's imported globally and css-loader is used. https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

juliodialpad:
After a long investigation the whole weekend trying to async load and tree-shake icons. Here are my thoughts:

- When async loading the spacing is not reserved until the icon is fully loaded, in order to solve this, I used DtSkeleton to reserve the space until the icon fully loads. After adding this, noticed a downgrade on performance on the `icon-variants` story (maybe because it's loading all the icons at once?).
- Our current DtIcon implementation is not tree-shakeable and it will never be as long as were using a prop to determine which icon should render, we'll need to do a big migration if we want to be able to fully tree-shake our icons.
- Modified dialtone-icons output to be able to use every icon as standalone component, passing the same props as `DtIcon`, this will allow us to make the migration gradually while still supporting the usage of `dt-icon`.
- The bundle size and performance of the consumer applications can be improved amazingly if we do our library components isolated from each other as much as possible and make the right imports to avoid polluting the bundle with unused code.

Here are some screenshots of the comparison between bundle sizes on test applications using different configs.

## Webpack

### Using DtIcon component

-  Using `import { DtIcon } from '@dialpad/dialtone-vue';`

Screenshot 2024-03-25 at 4 37 31 a m
Screenshot 2024-03-25 at 4 37 56 a m

- Using `import { DtIcon } from '@dialpad/dialtone-vue/components/icon';`

Screenshot 2024-03-25 at 4 35 05 a m
Screenshot 2024-03-25 at 4 36 02 a m

### Using dialtone-icons directly (suggested migration)

- Using `import { DtIconAccessibility } from '@dialpad/dialtone-icons/vue2';`

Screenshot 2024-03-25 at 4 43 46 a m
Screenshot 2024-03-25 at 4 45 22 a m

- Using `import DtIconAccessibility from '@dialpad/dialtone-icons/vue2/accessibility';`

Screenshot 2024-03-25 at 4 47 39 a m
Screenshot 2024-03-25 at 4 48 09 a m

## Vite

### Using DtIcon component

-  Using `import { DtIcon } from '@dialpad/dialtone-vue';`

Screenshot 2024-03-25 at 4 56 15 a m
Screenshot 2024-03-25 at 4 56 53 a m

- Using `import { DtIcon } from '@dialpad/dialtone-vue/components/icon';`

Screenshot 2024-03-25 at 4 57 29 a m
Screenshot 2024-03-25 at 4 57 44 a m

### Using dialtone-icons directly (suggested migration)

- Using `import { DtIconAccessibility } from '@dialpad/dialtone-icons/vue2';`

Screenshot 2024-03-25 at 4 58 58 a m
Screenshot 2024-03-25 at 4 59 19 a m

- Using `import DtIconAccessibility from '@dialpad/dialtone-icons/vue2/accessibility';`

Screenshot 2024-03-25 at 4 59 58 a m
Screenshot 2024-03-25 at 5 00 11 a m

juliodialpad:
Visual tests are marking all the snapshots as "new" due to the base branch is `remove-multiple-exports` and not staging, this will be merged into that branch and will run the visual tests there anyway.

braddialpad:
> Visual tests are marking all the snapshots as "new" due to the base branch is `remove-multiple-exports` and not staging, this will be merged into that branch and will run the visual tests there anyway.

ah yeah that makes sense

braddialpad:
> * Our current DtIcon implementation is not tree-shakeable and it will never be as long as were using a prop to determine which icon should render, we'll need to do a big migration if we want to be able to fully tree-shake our icons.

In theory if we were to use dynamic import for this wouldn't that fix the issue? Then the icon would only be imported when used?

braddialpad (approved):

juliodialpad:
> > * Our current DtIcon implementation is not tree-shakeable and it will never be as long as were using a prop to determine which icon should render, we'll need to do a big migration if we want to be able to fully tree-shake our icons.
> 
> In theory if we were to use dynamic import for this wouldn't that fix the issue? Then the icon would only be imported when used?

No, the way bundlers work is that if you do dynamic import it'd import every possibility so you're always importing all of them in that case.

We need to move out of dt-icon to make it fully tree-shakeable, another option would be to use CSS icons, something like `font awesome` were you only apply a class `<i class="fa fa-home" />` and the icons are rendered by CSS. That way, we would only need to use `dt-icon` to add that class dynamically.

PR #234, opened 2024-03-25
docs: NO-JIRA fix example in blog post

braddialpad:
# docs: fix example in blog post

- [x] Documentation

## :book: Description

Small error in the blog post that was causing some developers confusion. The example was showing the old format of the utility class instead of the new one.


braddialpad:
going to just push this to avoid further confusion

PR #244, opened 2024-03-29
feat(focus-ring): NO-JIRA reduce focus ring size

francisrupert:
# Reduce focus ring size

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Description

Now that focus ring style has been in play in product for a while now, it's a touch overbearing. This reduces it by one miserable pixel.

## :pencil: Checklist

For all PRs:

For all CSS changes:

image

### DIFF
--- a/packages/dialtone-tokens/tokens/base/default.json
+++ b/packages/dialtone-tokens/tokens/base/default.json
@@ -492,7 +492,7 @@
           "x": "{size.0}",
           "y": "{size.0}",
           "blur": "{size.0}",
-          "spread": "{size.300}",
+          "spread": "{size.200} + {size.100}",
           "color": "{color.border.focus}",
           "type": "dropShadow"
         }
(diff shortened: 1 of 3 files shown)


packages/dialtone-tokens/tokens/base/default.json
```diff
@@ -492,7 +492,7 @@
           "x": "{size.0}",
           "y": "{size.0}",
           "blur": "{size.0}",
-          "spread": "{size.300}",
+          "spread": "{size.200} + {size.100}",
```
juliodialpad, line 495: Should this be a specific size instead of a calculation? 

I remember we were discussing about this calculations on sizes but don't remember exactly what was the final decision on that discussion.

francisrupert: This one is acceptable and is inline with how we math certain Design Tokens, and isn't necessarily a magic number. In this case it is giving the appearance of `2px` while technically being `3px`. We definitely don't want to create a Design Token for `3px` since that size is against the grain of the size/space system.

juliodialpad (approved):
Thanks!

PR #256, opened 2024-04-09
feat: DLT-1680 add ability to enable/disable extensions

bianca-artola-dialpad:
# feat: DLT-1680 add ability to enable/disable extensions

Obligatory GIF

- [x] Refactor

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DLT-1680

## :book: Description
After some discussion with @braddialpad on the ticket, we conclude that the best option for this improvement was to add one parameter per extensions so we maintain the same logic that we made on other components.
Some technical considerations:

- I set all the new parameters as true as default to avoid breaking current behavior of rich_text_editor that is being used. Adding @emariotti-dialpad as a reviewer to check that everything continue working as expected.
- I left some basics extensions as default of the component because of 2 reasons. First of all, I don't think will be useful to have the component without those ones. Second, the component needs some default extensions as default as was failing when removing them.

## :bulb: Context
On the web app we are planning to use the Message Input component. For now, this component will be only used on web (not on mobile) and there are some functionalities that we currently don't have, like bold, italic, strikethrough, etc. For that reason, we want to allow clients to have the opportunity to enable/disable the different extensions that the component provide. 

## :pencil: Checklist

For all PRs:


braddialpad:
> * I left some basics extensions as default of the component because of 2 reasons. First of all, I don't think will be useful to have the component without those ones. Second, the component needs some default extensions as default as was failing when removing them.

Definitely makes sense to have some defaults set.

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
```diff
@@ -153,6 +154,62 @@ export default {
...
+      default: true,
+    },
+
+    /**
+     * Whether the input allows for list item to be introduced in the text.
+     */
+    allowListItem: {
+      type: Boolean,
+      default: true,
+    },
```
braddialpad: Hmm does it make sense to have this separate from "allowBulletList"? Seems like this should always be enabled if allowBulletList is enabled.

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
```diff
@@ -153,6 +154,62 @@ export default {
...
+
+    /**
+     * Whether the input allows for bullet list to be introduced in the text.
+    */
+    allowBulletList: {
+      type: Boolean,
+      default: true,
+    },
```
braddialpad, line 180: Error when enabling this in storybook.

Screenshot 2024-04-09 at 1 05 07 PM

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
```diff
@@ -153,6 +154,62 @@ export default {
...
+
+    /**
+     * Whether the input allows for underline to be introduced in the text.
+     */
+    allowUnderline: {
+      type: Boolean,
+      default: true,
+    },
```
braddialpad, line 204: There is no way to actually enter an underline in markdown, so it's impossible to set this without a button or similar I guess maybe if you pasted text that contained underline? Also the reason markdown is like this is because underlining text can be confused with links on the web. Maybe we don't need to include this one at all?

bianca-artola-dialpad: For me is fine to remove this one, as we are not going to use it. @emariotti3 do you have any concern with this? can we remove it?

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
bianca-artola-dialpad: sounds fair. I will update this

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
bianca-artola-dialpad: fyi I think it can be triggered by cmd + shift + u

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
bianca-artola-dialpad: should be fixed now as I removed that prop. FYI I created a ticket some time ago because it seems that the list is not working: https://dialpad.atlassian.net/browse/DLT-1668

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
braddialpad: Yeah looks like it.. The editor component actually has a button for it in the UI and it might make more sense in the context of the editor component, so maybe it's best to just keep it. Not needed in the message input though.

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
braddialpad: ah okay thanks.. We'll look at that bug

braddialpad (approved):
I think we're good on this. Some issues with bullet list still, but that's a separate issue.

ninamarina:
For the future: we could pass an array of `disabledExtensions` (apart from the default ones), and another for `enabledExtensions`, in case a consumer wants to enable extra externsions that are not the defaults.
Also, add to the components docs what are the extensions by default.

PR #257, opened 2024-04-09
docs: DLT-1502 add token categorization

ninamarina:
# docs: add token categorization

Obligatory GIF

- [x] Documentation

## :book: Jira Ticket
[[DLT-1502]](https://dialpad.atlassian.net/browse/DLT-1502)

## :book: Description
Adds the following structure to the tokens page:

• Colors
 - Foreground
 - Surface
 - Border
 - Link
 - Neutral
 - Brand -> Including Gradient here.
 - Opacity
 - Components
 - Base

• Typography
 - Style: hides base tokens for the composition tokens
     Body
     Headline
     Label
     Helper
     Code
 - Family
 - Font weight
 - Textcase
 - Line height
 - Font size
 - Components

• Shadows
 Hides base tokens for the composition tokens

• Size
 "Base" size tokens without subcategory
 - Negative
 - Percentage
 - Radius
 - Border
 - Components
     icon/size/border is hidden

• Space
 "Base" space tokens without subcategory
 - Negative
 - Percentage

The filtering and the rendering is now done recursively since the structure is now more complex and to allow adding subcategories.

## :pencil: Checklist

For all PRs:

tokens-categories

[DLT-1502]: https://dialpad.atlassian.net/browse/DLT-1502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


francisrupert:
Awesome, looking forward to this! I know it's Draft PR at the moment, but while you're in here can you add `autofocus` attribute on the `<input>` fields for [Tokens](https://dialtone.dialpad.com/deploy-previews/pr-257/tokens/) and  [Icons](https://dialtone.dialpad.com/deploy-previews/pr-257/design/icons/)?

francisrupert:
Let's fill in the empty left nav with _something_ for now, even if speculative. Ask @yorbi-dp how that should be, e.g. something like the attached. Note the "Planned" badge, as seen on [Design Language](https://dialtone.dialpad.com/deploy-previews/pr-257/design/).

image

ninamarina:
> Let's fill in the empty left nav with _something_ for now, even if speculative. Ask @yorbi-dp how that should be, e.g. something like the attached. Note the "Planned" badge, as seen on [Design Language](https://dialtone.dialpad.com/deploy-previews/pr-257/design/).

This will be done here: https://dialpad.atlassian.net/browse/DLT-1698

ninamarina:
> add `autofocus` attribute on the `<input>` fields for [Tokens](https://dialtone.dialpad.com/deploy-previews/pr-257/tokens/) and [Icons](https://dialtone.dialpad.com/deploy-previews/pr-257/design/icons/)?

Adding this to the same ticket as above, because we don't have that property currently in DtInput

apps/dialtone-documentation/docs/.vuepress/common/token-utilities.js
```diff
@@ -45,3 +46,138 @@ export function getComposedShadowTokens (theme) {
     });
   return tokens;
 }
+
+// eslint-disable-next-line complexity
```
ninamarina: I don't love to ignore the complexity linter rule, but there was no way of categorizing the tokens with a general criteria (they don't always have the category / subcategory name in the same place of their name). So I decided to do a bunch of ifs that will at least be clear to the reader and avoid cumbersome regular expressions.

iropolo:
Huh this is growing a lot, kind of hard to follow it.

Could we generate `processedTokens` on built process?, then we avoid to construct it in component.

We should start to split this component and move all related stuff to these. (composables)

In `AllTokens` we should split:

- Aside

`search`
`table generator` (token tree)

iropolo:
Huh this is growing a lot, kind of hard to follow it.

Could we generate `processedTokens` on built process?, then we avoid to construct it in component.

We should start to split this component and move all related stuff to these. (composables)

In `AllTokens` we should split:

- Aside

`search`
`table generator` (token tree)

iropolo:
Huh this is growing a lot, kind of hard to follow it.

Could we generate `processedTokens` on built process?, then we avoid to construct it in component.

We should start to split this component and move all related stuff to these. (composables)

In `AllTokens` we should split:

- Aside

`search`
`table generator` (token tree)

iropolo:
Huh this is growing a lot, kind of hard to follow it.

Could we generate processedTokens on built process?, then we avoid to construct it in component.


We should start to split this component and move all related stuff to these. (composables)
`apps/dialtone-documentation/docs/.vuepress/common/token-utilities.js` these functions should be inside our table generator (token tree) composable.
If it is in `common` I suppose it is shareable through the app, seems like its not, it is specific for this component.


In `AllTokens` we should split:

- Aside
filteredHeaders
tocOptions
......

- Search
searchInput
searchToken
hasSearchTerm
resetSearch
formatSelectMenuOptions
setFormat
setTheme
noSearchResults
....


- table generator (token tree)
filteredTokens
updateHeaders
updateHeadersRecursively
filterTokens
addTokens
addComposedTokens
....

Definitely not scope of this ticket, but this is close to go out of hands.

apps/dialtone-documentation/docs/.vuepress/common/constants.js
```diff
@@ -24,9 +24,100 @@ export const CATEGORY_MAP = {
...
+  component: ['topbar', 'sidebar', 'presence', 'mention', 'avatar', 'badge', 'checkbox', 'icon', 'inputs', 'action'],
 };
 
+export const SUBCATEGORY_MAP = {
+  color: ['foreground', 'surface', 'border', 'link', 'neutral'],
+  font: ['family', 'weight', 'size'],
+  size: ['radius', 'border'],
+};
+
+export const getTokensStructure = () => ({
```
iropolo: Not sure why this is a function and not just a `constant`.

apps/dialtone-documentation/docs/.vuepress/baseComponents/AllTokens.vue
```diff
@@ -244,23 +246,14 @@ onBeforeMount(() => {
...
-          });
-        if (!processedTokens[format]) processedTokens[format] = {};
-        if (!processedTokens[format][theme.value]) processedTokens[format][theme.value] = {};
-        processedTokens[format][theme.value][category] = [...tokens];
+      // initialize processedTokens with desired structure
+      const structure = getTokensStructure();
+      Object.entries(tokensJson[theme.value]).forEach((token) => {
+        addTokenToStructure(token, format, structure);
       });
+      processedTokens[format][theme.value] = structuredClone(structure);
```
iropolo: if `getTokensStructure` were just a const we could easily use it here, without creating a new const and avoid using `structuredClone`, not sure why we are using it anyways, maybe Im missing some edge case here.

apps/dialtone-documentation/docs/.vuepress/common/token-utilities.js
braddialpad: I think it is fine in the case of processing files such as this to ignore complexity rule. The purpose of this function is basically to take a large amount of complexity and structure/simplify it, so it's understandable it's high complexity.

apps/dialtone-documentation/docs/.vuepress/common/constants.js
ninamarina: Is not a constant to avoid having to do a deep clone of the object

apps/dialtone-documentation/docs/.vuepress/common/constants.js
```diff
@@ -24,9 +24,100 @@ export const CATEGORY_MAP = {
...
+  color: {
+    foreground: {
+      _children: [],
```
braddialpad: Why _children instead of just making foreground an array itself?

apps/dialtone-documentation/docs/.vuepress/baseComponents/AllTokens.vue
ninamarina: I suppose I could assign `getTokensStructure ` directly to `processedTokens[format][theme.value]` and populate that one to avoid a `structuredClone`

apps/dialtone-documentation/docs/.vuepress/common/constants.js
ninamarina: I need it for the case of `size` and `space`, that have "orphan tokens", i.e tokens that are not under any subcategory

braddialpad:
> Adding this to the same ticket as above, because we don't have that property currently in DtInput

Because autofocus is a native HTML attribute and we have $attrs on the `<input>` element itself in DtInput, we should be able to just set this without making any changes to the component

ninamarina:
> Huh this is growing a lot, kind of hard to follow it.
> 
> Could we generate processedTokens on built process?, then we avoid to construct it in component.

We could, I don't see much gain there, because is not something that's impacting the performance and it's used only by the public website to show the tokens in this particular categorization. I think we should add this to the build process if we wanted to use this categorization somewhere else.
Still, some better file separation should be in place though.
> 
> We should start to split this component and move all related stuff to these. (composables) `apps/dialtone-documentation/docs/.vuepress/common/token-utilities.js` these functions should be inside our table generator (token tree) composable. If it is in `common` I suppose it is shareable through the app, seems like its not, it is specific for this component.

Yes, the `common` folder does not seem like a good place because it's not common to all the app, it's only for tokens. Will find a better place for these logic.
> 
> In `AllTokens` we should split:
> 
> * Aside
>   filteredHeaders
>   tocOptions
>   ......
> * Search
>   searchInput
>   searchToken
>   hasSearchTerm
>   resetSearch
>   formatSelectMenuOptions
>   setFormat
>   setTheme
>   noSearchResults
>   ....
> * table generator (token tree)
>   filteredTokens
>   updateHeaders
>   updateHeadersRecursively
>   filterTokens
>   addTokens
>   addComposedTokens
>   ....
> 
> Definitely not scope of this ticket, but this is close to go out of hands.

Agree

apps/dialtone-documentation/docs/.vuepress/common/token-utilities.js
```diff
@@ -45,3 +46,138 @@ export function getComposedShadowTokens (theme) {
...
+export const addTokenToStructure = (token, format, structure) => {
```
iropolo: Seems like this function can be splitted in 3.
Would be cool to separate some code from it to have it more clear.

-----
```
const [key, value] = token;
  if (!value[FORMAT_MAP[format]] || !value[FORMAT_MAP.CSS]) return;
```
-----
```
const { name, value: tokenValue, description, keywords } = value[FORMAT_MAP[format]];
  const { value: exampleValue, name: exampleName } = value[FORMAT_MAP.CSS];
  const displayToken = { exampleValue, exampleName, name, tokenValue, description, keywords };
```
-----
```
const splitKeys = key.split('/');

  // COLOR
  if (key.startsWith('color') && SUBCATEGORY_MAP.color.includes(splitKeys[1])) {
....
```

Just to follow in a better way what we are doing here and what the function should build.


About the `ifs`, seems like there are some pattern to follow.
What we have:

1. color
2. color/brand
3. opacity
4. color
5. typography
6. font
7. font/text-case
8. font/lineHeight
9. CATEGORY_MAP.component.includes(splitKeys[0]) && (splitKeys[1] === '`font`' || splitKeys[1] === 'lineHeight')
10. shadow
11. size
12. size/radius
13. size/border
14. size
15. if (CATEGORY_MAP.component.includes(splitKeys[0]) && splitKeys[1] === '`size`') { if (splitKeys[2] === 'border') {
16. space && negative
17. space && percent
18. space


Ok, its a lot, and also it depends on the order.

We definetly should separet these in different function, but these functions can be less specific.
As I see we can have something like:

1. color
2. opacity
3. typography
4. font
5. shadow
6. size
7. space 

It can be sth like:
`handleColorTokens(param_1, param_2, param_3)`
...

iropolo:
> > Huh this is growing a lot, kind of hard to follow it.
> > Could we generate processedTokens on built process?, then we avoid to construct it in component.
> 
> We could, I don't see much gain there, because is not something that's impacting the performance and it's used only by the public website to show the tokens in this particular categorization. I think we should add this to the build process if we wanted to use this categorization somewhere else. Still, some better file separation should be in place though.

Anything that we can prepare previously to serve the browser and avoid building it in it will be better.
Not just for performance impact but for code healthy.

It is better for the component to just grab the data from somewhere and use it, than build it + use it.

It is similar on what it happen on the emoji_picker backside.
All the emoji data/structure is being built before and the component just grab it and use it.

We should take note of this and see if we can do it in the improve ticket.

iropolo (approved):
Its a good implementation, looks and works well, there is nothing that cannot be done in a following improvement ticket.

We should take improvement ticket ASAP and before implement more features to avoid tech debt.

I dont see any blocker for it, If its get the Ok from design, it should be ready to go.

ninamarina:
> > > Huh this is growing a lot, kind of hard to follow it.
> > > Could we generate processedTokens on built process?, then we avoid to construct it in component.
> > 
> > 
> > We could, I don't see much gain there, because is not something that's impacting the performance and it's used only by the public website to show the tokens in this particular categorization. I think we should add this to the build process if we wanted to use this categorization somewhere else. Still, some better file separation should be in place though.
> 
> Anything that we can prepare previously to serve the browser and avoid building it in it will be better. Not just for performance impact but for code healthy.
> 
> It is better for the component to just grab the data from somewhere and use it, than build it + use it.
> 
> It is similar on what it happen on the emoji_picker backside. All the emoji data/structure is being built before and the component just grab it and use it.
> 
> We should take note of this and see if we can do it in the improve ticket.

"Anything that we can prepare previously to serve the browser and avoid building it in it will be better" -> I am not so sure about this. If tomorrow we decide we want to show this information in a different way, we would have to update the build process, to make a documentation change. Everything in the build should be to make sure we have the data, but how we show it should be a responsibility of the front.

iropolo:
> "Anything that we can prepare previously to serve the browser and avoid building it in it will be better" -> I am not so sure about this. 

Maybe I didnt explain it/express myself correctly, beyond if it is being serve from back or in the browser, the idea is to delegate the functionality, regardless where its being built.

Thats why I focus on `Not just for performance impact but for code healthy.`.

If it is being built in a different process, or in a different file it is the same, just remove it from the component.

ninamarina:
> > "Anything that we can prepare previously to serve the browser and avoid building it in it will be better" -> I am not so sure about this.
> 
> Maybe I didnt explain it/express myself correctly, beyond if it is being serve from back or in the browser, the idea is to delegate the functionality, regardless where its being built.
> 
> Thats why I focus on `Not just for performance impact but for code healthy.`.
> 
> If it is being built in a different process, or in a different file it is the same, just remove it from the component.

Oh yeah, to that I agree. Will separate this file better.

francisrupert:
image

The hierarchy is a bit off for Size and Space. There's technically a subcategory above "Negative".

@yorbi-dp, What say you? How should that be presented to keep hierarchy predictable? It doesn't *strictly* have a subcategory name from the design token classification, though it's implicitly "default" or "base" or some such.

apps/dialtone-documentation/docs/.vuepress/common/token-utilities.js
ninamarina: About the splitting, it's fixed.
About adding `handleColorTokens(params)`, `handleTypographyTokens(params)`, etc. I don't think it adds much readability, and also it might add complexity because the function would have to do two things:
1- Add the tokens to the structure
2- Return if it handled the tokens, or if not, in that case we should continue looking for other categories

ninamarina:
@iropolo did a lot of splitting if you want to take a look again. Thanks!

apps/dialtone-documentation/docs/.vuepress/common/constants.js
braddialpad: got it

braddialpad:
> What say you? How should that be presented to keep hierarchy predictable? It doesn't _strictly_ have a subcategory name from the design token classification, though it's implicitly "default" or "base" or some such.

Yeah agree this is quite confusing, think we should just call it base.

yorbi-dp (changes requested):
> @yorbi-dp, What say you? How should that be presented to keep hierarchy predictable? It doesn't strictly have a subcategory name from the design token classification, though it's implicitly "default" or "base" or some such.

Yep, I agree it doesn't have a subcategory name beyond what they represent, but to maintain a predictable hierarchy, let's group them into "Base". cc @ninamarina 

This is looking great!

yorbi-dp (approved):
Looks great!! Thank you Nina.

PR #259, opened 2024-04-09
feat(breakpoints): add responsive margins

fabiangarga:
# Add responsive margins

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

### DIFF
--- a/packages/dialtone-css/gulpfile.cjs
+++ b/packages/dialtone-css/gulpfile.cjs
@@ -32,6 +32,7 @@ const breakpoints = [
 const classes = [
   /\.d-d-(flex|none|block)$/, // Display Flex, None and Block
   '.d-t0',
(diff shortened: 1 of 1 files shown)


fabiangarga:
Not sure if we should add all of them, might be useful but I just need top/bottom at the moment.

braddialpad (commented):
Wondering what is your use case for this? If it's space between items we generally try to use margins as little as possible instead using the [dt-stack](https://dialtone.dialpad.com/vue/index.html?path=/story/components-stack--default) component with gap.

fabiangarga:
@braddialpad Thanks for suggestion `dt-stack`, if I understand that component properly, it will imply using logic on the component and applying the gap based on the width, which seems like too much for the use case I was looking. Responsive margins IMO are quite a common thing, since different layouts might imply different spaces around the components. But anyways, closing this since I will use a different approach.

PR #266, opened 2024-04-12
fix: NO-JIRA update exports and peerDependencies

juliodialpad:
# Updated exports and peer dependencies

- Updated dialtone-vue 2, dialtone-vue 3 and mono-package exports
- Updated vite.config on dialtone-vue 2 and dialtone-vue 3 to divide common/* files and disable minification to help on debugging.
- Updated peer dependencies versioning to exact version to avoid installing incompatible versions.


package.json
```diff
@@ -106,36 +106,36 @@
...
+    "@tiptap/vue-2": "2.0.3",
+    "@tiptap/vue-3": "2.0.3",
+    "date-fns": "2.30.0",
+    "docopt": "0.6.2",
+    "emoji-regex": "10.2.1",
+    "emoji-toolkit": "8.0.0",
+    "requireindex": "1.2.0",
     "stylelint": "15.11.0",
-    "tippy.js": "^6.3.7"
+    "tippy.js": "6.3.7"
```
braddialpad: Can we set `save-prefix=""` in our npmrc so it always installs exact versions by default when we do `pnpm add`?

juliodialpad: Didn't knew about that, let me check.

juliodialpad: Either using `--save-prefix` or `--save-exact` flags or adding `save-exact=""` to `.npmrc` worked for me, but added anyways in case it works for someone else.

https://pnpm.io/npmrc#save-prefix
https://pnpm.io/cli/add#--save-exact--e
https://docs.npmjs.com/cli/v10/using-npm/config#save-prefix

braddialpad (approved):

braddialpad:
Some unit test errors, seems like something to do with multiple taptap versions being loaded.

juliodialpad:
> Some unit test errors, seems like something to do with multiple taptap versions being loaded.

Yes, seems like different versions for vue 2, vue 3 and root packages were being installed, now it's more important that ever that we use exact versions to avoid ending up having multiple versions on multiple packages without noticing due to pnpm

PR #271, opened 2024-04-17
docs(tokens): DLT-1637 add copy button on hover

ninamarina:
# docs(tokens): DLT-1637 add copy button on hover

Obligatory GIF

- [x] Documentation

## :book: Jira Ticket
[DLT-1637]

## :book: Description
Adds Copy Button on hover for the tokens section. Uses LazyShow component to improve render time.

## :pencil: Checklist

For all PRs:

For all CSS changes:

tokens-copy

[DLT-1637]: https://dialpad.atlassian.net/browse/DLT-1637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
+++ b/apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
@@ -32,10 +32,11 @@
       <tr
         v-for="({ exampleValue, exampleName, name, tokenValue, description }) in tokens"
         :key="name"
+        tabindex="0"
         @mouseenter="onEnterRow(name)"
-        @mouseleave="onLeaveRow(name)"
-        @focusin="onEnterRow(name)"
-        @focusout="onLeaveRow(name)"
+        @mouseleave="onLeaveRow()"
+        @focusin="onFocusInRow(name)"
+        @focusout="onFocusOutRow()"
       >
         <td>
           <token-example :category="category" :name="exampleName || name" :value="exampleValue.toString()" />
@@ -53,6 +54,7 @@
             <div class="d-w32">
               <dt-lazy-show :show="showCopyButton(name)">
                 <copy-button
+                  :ref="`copyButton${name}`"
                   :text="name"
                   aria-label="Copy Token"
                 />
@@ -165,12 +167,32 @@ export default {
       this.hoveredRow = name;
     },
 
-    onLeaveRow (name) {
+    onLeaveRow () {
       this.hoveredRow = null;
     },
 
     showCopyButton (name) {
       return this.hoveredRow === name;
+    },
+
+    onFocusOutRow () {
+      if (document.activeElement.tagName === 'TR') {
+        this.hoveredRow = null;
+      }
+    },
+
+    async onFocusInRow (name) {
+      if (this.hoveredRow === name) return;
+      this.hoveredRow = name;
+      // when focusing the next row we want to redirect the focus to
+      // the copy button
+      if (document.activeElement.tagName === 'TR') {
+        await this.$nextTick(); // wait for the copy button to render
+        const copyButtonRef = this.$refs[`copyButton${name}`];
+        if (Array.isArray(copyButtonRef)) {
+          copyButtonRef[0]?.$el?.querySelector('button')?.focus();
+        }
+      }
     },
   },
 };
--- a/apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
+++ b/apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
@@ -1,4 +1,3 @@
-<!-- eslint-disable vuejs-accessibility/mouse-events-have-key-events -->
 <template>
   <table class="d-table dialtone-doc-table">
     <thead>
@@ -35,6 +34,8 @@
         :key="name"
         @mouseenter="onEnterRow(name)"
         @mouseleave="onLeaveRow(name)"
+        @focusin="onEnterRow(name)"
+        @focusout="onLeaveRow(name)"
       >
         <td>
           <token-example :category="category" :name="exampleName || name" :value="exampleValue.toString()" />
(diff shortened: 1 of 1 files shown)


apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
```diff
@@ -1,3 +1,4 @@
+<!-- eslint-disable vuejs-accessibility/mouse-events-have-key-events -->
```
juliodialpad: > Enforce onmouseover/onmouseout are accompanied by onfocus/onblur. Coding for the keyboard is important for users with physical disabilities who cannot use a mouse, AT compatibility, and screenreader users.


I don't like to disable eslint rules, I think it's ok to add onfocus/onblur events to make sure keyboard navigation is enabled too.

https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/mouse-events-have-key-events.md

apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
```diff
@@ -148,6 +159,26 @@ export default {
     getSecondValue (value) {
       return value.split(' / ')[1];
     },
+
+    onEnterRow (name) {
+      console.log('ENTER', name);
```
juliodialpad: remove console.log

apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
```diff
@@ -148,6 +159,26 @@ export default {
...
+      this.hoveredRow = name;
+    },
+
+    onLeaveRow (name) {
+      console.log('LEAVE', name);
```
juliodialpad: remove console.log

apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
```diff
@@ -148,6 +159,26 @@ export default {
...
+    },
   },
 };
 </script>
+
+<style scoped>
+  .copy-button {
+    position: absolute;
+  }
+</style>
```
juliodialpad: why not using `class="d-ps-absolute"`?

juliodialpad (changes requested):

yorbi-dp (commented):
Clean! Thank you, Nina. 
Only one thing, is it possible to add the option to 'tab with the keyboard' to jump into each token line, or does the 'copy button' need to be visible for that?

apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
ninamarina: In this case, the focus over the row is not enabled, so I didn't think it made much difference. But I will add them in case we add keyboard navigation in the future 👍

apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
ninamarina: Oh, this was not supposed to be committed. Removed it, thank you

iropolo (approved):

yorbi-dp (approved):
Thank you Nina! Looks great.

apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
```diff
@@ -161,24 +164,36 @@ export default {
...
+      // when focusing the next row we want to redirect the focus to
+      // the copy button
+      if (document.activeElement.tagName === 'TR') {
+        await this.$nextTick(); // wait for the copy button to render
+        const copyButtonRef = this.$refs[`copyButton${name}`];
+        if (Array.isArray(copyButtonRef)) {
+          copyButtonRef[0]?.$el?.querySelector('button')?.focus();
+        }
+      }
+    },
```
juliodialpad: I think is more natural to focus the row and then the copy button, that way the screen reader can also read the value. It is important for a11y to know what you're about to copy.

juliodialpad (changes requested):

apps/dialtone-documentation/docs/.vuepress/baseComponents/TokenTable.vue
juliodialpad: Something like we're doing on the contact-row recipe https://dialtone.dialpad.com/vue3/index.html?path=/story/recipes-leftbar-contact-row--default

https://github.com/dialpad/dialtone/assets/87546543/32568999-5925-43db-b56b-499d7a71cd2c

ninamarina: Makes sense. @yorbi-dp if you agree I would go with that approach, this is how it would work:
tokens-nav

yorbi-dp: I see your point; focusing on the row first, then the copy button does seem more natural, especially for accessibility purposes to ensure the screen reader reads the value. However, since the row doesn't have any associated action or navigation, unlike the contact-row which likely opens a chat, there might not be a pressing need to focus on it. Alternatively, adding an aria-label to the button, such as 'Copy <token name> to clipboard,' could be a good solution. What do you guys think?

ninamarina: The button has an aria-label that is "copy token", the thing is that without the focus on the row you cannot really read the name, value and description. Unless there's a way of doing that with aria attributes, but I don't think it's posible doing it this way because since the button is not initially visible, I need to remove the focus from the row to the button once it's visible and that messes with the accessibility.
I would go with the focus in the row even if it doesn't lead to an action, because we want to prioritize accessibility.

yorbi-dp: Oh, got it. Awesome, let's do the focus on the row then. Thank you! 🙌

juliodialpad (approved):
Thanks Nina!

PR #274, opened 2024-04-18
fix(message-input): DLT-1714 limit file type on image picker

bianca-artola-dialpad:
# fix(message-input): DLT-1714 limit file type on image picker

Obligatory GIF

- [x] Fix

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DLT-1714

## :book: Description
This pr adds _accept_ parameter for the input on the message_input component so we can limit the file types the file input should accept. Documentation: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#accept

## :bulb: Context
Currently, on Dialpad app we only allow sending images and videos, no PDFs or other file types. Also as the icon is an Image I think is correct to limit it to only allow images and videos.

## :pencil: Checklist

For all PRs:

### DIFF
--- a/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
+++ b/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
@@ -76,6 +76,7 @@
             <dt-input
               ref="messageInputImageUpload"
               data-qa="dt-message-input-image-input"
+              accept="image/*, video/*"
               type="file"
               class="d-ps-absolute"
               multiple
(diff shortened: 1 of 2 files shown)


juliodialpad:
I guess you will still need to do a validation product-side to avoid the files being added, as this only sets the default options but that can be changed, I don't know if there's a way to limit that on this component, anyways the behavior on multiple operating systems would differ, so this will still need to be done product-side.

On this component

https://github.com/dialpad/dialtone/assets/87546543/df03dc18-43fe-4fcb-b8e9-06e3b30bd10d



Currently on product

https://github.com/dialpad/dialtone/assets/87546543/0e930cfd-f9c8-4baf-99e3-fe0331ba44a3

juliodialpad (approved):

bianca-artola-dialpad:
Yes, good catch. We have a function that makes that work: https://github.com/dialpad/firespotter/blob/64185a62eaeb56cf645f8bc66c6fd404c1b56b84/ubervoice/static/js/vue_mixins/files_mixin.js#L32-L36
Thank you!

PR #283, opened 2024-04-24
docs: DLT-1538 illustrations workflow

iropolo:
# DLT-1538 Illustrations workflow

Output:
Before:
```
dist/
├─ icons.json
├─ keywords.json
├─ svg/
│  ├─ *.svg
```

After:
```
dist/
├─ icons.json
├─ illustrations.json
├─ keywords-icons.json
├─ keywords-illustrations.json
├─ svg/
│  ├─ illustrations/
│  │  ├─ *.svg
│  ├─ icons/
│  │  ├─ *.svg

```

-----
## Breaking changes

We updated `keywords.json` to `keywords-icons.json`
svg folder now has a child folder for `icons` and `illustrations` then all the `*.svg` inside.

-----
## Images

image

image

Screenshot 2024-04-29 at 11 56 17

-----
## Questions

- Should SVG Illustrations have class `d-illustration d-illustration--${name}` like icons?
- Is the category name 'spot-illustrations' ok?, should we have it or no?. Would be cool if we have one so we already have the workflow in place like the icons.


packages/dialtone-icons/transformSVGtoVue.cjs
```diff
@@ -11,32 +11,38 @@ const _ = require('lodash');
...
-      const template = fs.readFileSync('./src/IconTemplate.vue', 'utf8');
-      const iconName = `dt-icon-${fileName.replace('.svg', '')}`.toLowerCase()
-      .split('-')
-      .map(word => word.charAt(0).toUpperCase() + word.slice(1))
-      .join('');
+      const filePath = `./dist/svg/${fileName}`;
 
-      let result = template.replace('__SVG_CONTENT__', svgContent);
-      result = result.replaceAll('__ICON_NAME__', iconName);
+      if (fs.statSync(filePath).isFile()) { // Check if the item is a file
```
iropolo, line 16: `if (fs.statSync(filePath).isFile()) {`

Did this just to by pass my error, I can remove it before merge.

packages/dialtone-icons/transformSVGtoVue.cjs
```diff
@@ -2,42 +2,88 @@ const fs = require('node:fs');
...
+
+        // // Create unique IDs
+        // result = result.replace(/(clip-path|fill)="url\(#([^)]+)\)"/g, ':$1="`url(#${uniqueID}$2)`"')
+        // result = result.replace(/(clipPath|linearGradient|radialGradient) id="([^"]+)"/g, '$1 :id="`${uniqueID}$2`"')
+        //
+        // if (!/\${uniqueID}/g.test(result)) {
+        //   // Remove created function if not needed
+        //   result = result.replace(/\n\s+this.uniqueID.*;/, '');
+        //   result = result.replace(/import \{ getUniqueString \} from '\.\.\/utils';\n+/, '');
+        // }
```
iropolo: Removed from illustrations components since it is affecting the svg.
The svg, does not work anymore after doing this. Probably something with the regex.

Since illustrations will be being used in emptyState components, I guess for now it should not be an issue the duplicate id error.

packages/dialtone-icons/gulpfile.cjs
```diff
@@ -130,6 +143,57 @@ const buildIcons = function (done) {
...
+            cleanupIDs: {
+              minify: true,
+            },
+          }
+        ],
+      }
+    }))
+    .pipe(rename({ dirname: '' }))
+  .pipe(dest(paths.illustrations.outputSvg));
+};
```
juliodialpad: Don't think illustrations need this, this is a pre-processing needed to standardize the icons colors and sizes so we can handle it correctly with CSS.

Please confirm with designers what part of this processing is needed and what is not.

packages/dialtone-icons/transformSVGtoVue.cjs
juliodialpad: This doesn't modify the SVG, only the Vue component output.

It's only reading the files, never writing the SVG's back, only the vue components.

juliodialpad (changes requested):
> Should SVG Illustrations have class d-illustration d-illustration--${name} like icons?

Yes, let's keep the process as close as possible to the icons to ease the migration later

> Is the category name 'spot-illustrations' ok?, should we have it or no?. Would be cool if we have one so we already have the workflow in place like the icons.

It's ok to have that category name on the input, the output files should be output to a plain structure (currently working like that).

Please add `packages/dialtone-icons/src/illustrations` to `.gitignore` we don't want to ship generated files to GitHub.

 Looking good so far! thanks

juliodialpad:
As we're renaming the file from `keywords.json` to `keywords-icons.json` and `keywords-illustrations.js` a migration will be needed on our documentation site and maybe on product if they're using that file somewhere.

Also you need to update the exports on `packages/dialtone-icons/package.json` and `package.json` to match the new paths.

iropolo:
- Added `packages/dialtone-icons/src/illustrations` to `.gitignore`
- Doc updates related to `keywords.json`, `keywords-icons.json` and `keywords-illustrations.json`
- Updated `exports` in our `packages/dialtone-icons/package.json`

I don't see `keywords.json` file is being used in product somewhere.

Waiting from design directive on our illustration standardize process.

packages/dialtone-icons/gulpfile.cjs
fede-dp: I think this isn't needed today, but eventually we will need to have some kind of size for the same illustration, if we want to have it for the future will be great

packages/dialtone-icons/transformSVGtoVue.cjs
iropolo: 👍🏼 , solved, was confused cos it grabs the already minifed one that is on `/dist/` folder.
We removed the minifed process from the `.pipe`.

packages/dialtone-icons/gulpfile.cjs
iropolo: k, I will just leave essential stuff.

iropolo:
Should be ok to review now.

braddialpad:
> ## Breaking changes
> We updated `keywords.json` to `keywords-icons.json` 

Probably not an issue, I doubt they are referencing this in product yet.

> svg folder now has a child folder for `icons` and `illustrations` then all the `*.svg` inside.

Does this mean anything product side directly referencing the svgs has to change it's path? what about the vue components?

apps/dialtone-documentation/docs/guides/contributing/index.md
```diff
@@ -40,7 +44,7 @@ If you need to add an icon into Dialtone, here’s how you would go about doing
 5. Open a pull request.
 6. Once approved it can be merged into staging and will go out in the next release.
 
-### For spot illustrations
+### For spot illustrations in Dialtone CSS
```
braddialpad: We can remove this entire section now can't we? spot illustrations shouldn't be in dialtone css anymore after this.

apps/dialtone-documentation/docs/guides/contributing/index.md
```diff
@@ -19,12 +19,16 @@ in our Dialtone repository.
...
+This process is similar for both icons and illustrations.
+The exported SVG file(s) should be placed in the appropriate folder category inside `packages/dialtone-icons/src/svg/illustrations/`.
+The keywords for the illustrations should be added to the `keywords-illustrations.json` file.
 
 1. Create a new branch starting with "feat/" in the name.
-2. Place the exported SVG file(s) in the appropriate folder category inside `packages/dialtone-icons/src/svg/`
+2. Place the exported SVG file(s) in the appropriate folder category inside `packages/dialtone-icons/src/svg/icons/`
 3. If you need to add keywords related to the icon.
-   - Add the icon name to `packages/dialtone-icons/src/keywords.json` into the correct category.
+   - Add the icon name to `packages/dialtone-icons/src/keywords-icons.json` into the correct category.
```
braddialpad: Split this into two separate groups of steps, one for icons and one for illustrations. I know it seems a little repetitive but when someone is adding an icon for the first time they're going to want exact steps to easily follow. Right now they kind of have to figure out on their own what to change when adding an illustration

packages/dialtone-icons/.github/CONTRIBUTING.md
```diff
@@ -47,9 +47,12 @@ Generated Vue icons are output to the `src/icons/` folder when you do `nx build
 
 ### Important Folders and Files
 
-- `src/svg`: All the source SVG icon files.
-- `keywords.json`: Contains the categories on which icons are going to be included and the keywords to make the icons more discoverable while searching on [Dialtone icons documentation](https://dialpad.design/components/icon.html).
+- `src/icons/svg`: All the source SVG icon files.
+- `src/illustrations/svg`: All the source SVG icon files.
+- `keywords-icons.json`: Contains the categories on which icons are going to be included and the keywords to make the icons more discoverable while searching on [Dialtone icons documentation](https://dialpad.design/components/icon.html).
+- `keywords-illustrations.json`: Contains the categories on which icons are going to be included and the keywords to make the icons more discoverable.
```
braddialpad: substitute 'icons' with 'illustrations' in this line.

iropolo:
> > svg folder now has a child folder for `icons` and `illustrations` then all the `*.svg` inside.
> 
> Does this mean anything product side directly referencing the svgs has to change it's path? what about the vue components?

I didnt see using it, should be already migrated to `dt-icon` component.
Which vue components? from Product? not sure what are you referring to.

apps/dialtone-documentation/docs/guides/contributing/index.md
iropolo: Huh, I guess it is still necessray, this section is this one that @fede-dp built. https://dialtone.dialpad.com/design/illustrations/

Not sure, if he can confirm it to us would be better.

apps/dialtone-documentation/docs/guides/contributing/index.md
iropolo: Ye I thought that the paragraph at the top would be enough but ok, lemme added it twice.

braddialpad:
> I didnt see using it, should be already migrated to `dt-icon` component. Which vue components? from Product? not sure what are you referring to.

a reference like this would not change right?

```
import DtIconPhoneHangUp from '@dialpad/dialtone-icons/vue3/phone-hang-up';
```

iropolo:
> > I didnt see using it, should be already migrated to `dt-icon` component. Which vue components? from Product? not sure what are you referring to.
> 
> a reference like this would not change right?
> 
> ```
> import DtIconPhoneHangUp from '@dialpad/dialtone-icons/vue3/phone-hang-up';
> ```

Hm... it shouldn't, Im not modifying anything related to the icons itself, just the SVGs folder.
@juliodialpad ping to have a second verifying on it.

braddialpad:
Also how do we import the new illustrations, would it be like this?

```
import DtIllustrationPhoneHangUp from '@dialpad/dialtone-icons/vue3/blank-space';
```

or like this?
```
import DtIllustrationPhoneHangUp from '@dialpad/dialtone-icons/vue3/illustrations/blank-space';
```

iropolo:
> Also how do we import the new illustrations, would it be like this?
> 
> ```
> import DtIllustrationPhoneHangUp from '@dialpad/dialtone-icons/vue3/blank-space';
> ```
> 
> or like this?
> 
> ```
> import DtIllustrationPhoneHangUp from '@dialpad/dialtone-icons/vue3/illustrations/blank-space';
> ```

Huh, should be through the DtIllustration component. Like icons with DtIcon component.
Not sure if it is necessary to be imported like that, but If it, we should add it then...

But, who is using that way? is that the deprecated way?

fede-dp:
To me it's makes sense to have this hierarchy, do you think we will need something else? @francisrupert

apps/dialtone-documentation/docs/guides/contributing/index.md
iropolo: After discussion, we will remove it (https://dialtone.dialpad.com/guides/contributing/#for-spot-illustrations).
It is not necessary anymore. 
@fede-dp @braddialpad

fede-dp (approved):
looks great!

iropolo:
Ping here @braddialpad @juliodialpad

braddialpad:
> Huh, should be through the DtIllustration component. Like icons with DtIcon component. Not sure if it is necessary to be imported like that, but If it, we should add it then...
> 
> But, who is using that way? is that the deprecated way?

When @juliodialpad did his research on tree shaking he found the if we dynamically import assets at runtime (like in dt-icon) it's actually impossible to tree shake them. it will always include all of them in the build.

```
import DtIllustrationBlankSpace from '@dialpad/dialtone-icons/vue3/illustrations/blank-space';
```

We can merge this PR once we can import using the above method. Using the dt-icon component is no longer recommended, it still exists because it is in use in so many places, but we'll have to migrate it at some point.

juliodialpad (changes requested):
You're missing the illustrations exports on `packages/dialtone-icons/vue2/vite.config.js` and `packages/dialtone-icons/vue3/vite.config.js`.

Check the files and add the entries just as the `iconsEntries`.

If we don't add that entries, the illustrations are not going to be tree-shakeable.

iropolo:
> You're missing the illustrations exports on `packages/dialtone-icons/vue2/vite.config.js` and `packages/dialtone-icons/vue3/vite.config.js`.
> 
> Check the files and add the entries just as the `iconsEntries`.
> 
> If we don't add that entries, the illustrations are not going to be tree-shakeable.

Added on [004aef3](https://github.com/dialpad/dialtone/pull/283/commits/004aef38b5c408fdd105c7cfe3e93d0702269e4e)

braddialpad:
It wasn't outputting to src/illustrations because the folder was not included in build outputs in project.json

added `{projectRoot}/src/illustrations` to the array and it fixed it in 72e2040. Looks good now.

braddialpad (approved):

juliodialpad (changes requested):
Previously, with this export 
```json
"exports": {
    ...
    "./*": "./dist/svg/*"
  },
``` 

we could access the icons on `@dialpad/dialtone-icons/accessibility.svg` as the icons were directly on the `dist/svg` folder, as you introduced `icons` and illustrations folders, now all those exports changed.

Maybe they're using a bunch of them on product. I'll recommend to leave icons and illustrations at the `dist/svg` level to avoid such breaking change.

juliodialpad:
> It wasn't outputting to src/illustrations because the folder was not included in build outputs in project.json
> 
> added `{projectRoot}/src/illustrations` to the array and it fixed it in [72e2040](https://github.com/dialpad/dialtone/commit/72e2040f0ab7f456daa9a74312c19cee8ecda082). Looks good now.

That `outputs` array is for NX to determine if the output matches the cache or it needs to re-run the build. Thanks for adding it! forgot about it.

braddialpad:
> Maybe they're using a bunch of them on product. I'll recommend to leave icons and illustrations at the `dist/svg` level to avoid such breaking change.

I think they actually are not, since our icon migration nothing is using svg anymore

juliodialpad:
> > Maybe they're using a bunch of them on product. I'll recommend to leave icons and illustrations at the `dist/svg` level to avoid such breaking change.
> 
> I think they actually are not, since our icon migration nothing is using svg anymore

I found some references on our documentation, seems like it's only documentation that is pointing to that old path.

https://github.com/dialpad/dialtone/blob/8905669111ab992de313821e157e076f3367755f/packages/dialtone-icons/README.md#L39

https://github.com/dialpad/dialtone/blob/3f625ad59b5e867d122328315a70f3c31f374ed4/README.md#L299

iropolo:
> I found some references on our documentation, seems like it's only documentation that is pointing to that old path.

Removed old implementation references and updated doc.

juliodialpad (approved):

PR #287, opened 2024-04-25
feat: DLT-1651 message input parse suggestion

juliodialpad:
# Message input parse suggestion

Obligatory GIF

- [x] Feature
## :book: Jira Ticket

- https://dialpad.atlassian.net/browse/DLT-1651

## :book: Description

- Updated @tiptap dependencies to 2.3.0
- Added input and paste functionality to mentions and channels
- Fixed an issue found with emoji pasting
- Updated DtRichTextEditor to parse mentions, channels and emojis when setting the content through the `value` prop.
- Added `update:value` emit to DtMessageInput to be able to use v-model correctly.

## :bulb: Context

- We needed mentions to be initialized through the message input `value` prop.
- Found some other issues along the way so fixed them too.
- Created a ticket of an issue found that would make this feat to take longer: https://dialpad.atlassian.net/browse/DLT-1731

## :pencil: Checklist

For all PRs:

For all Vue changes:

## :crystal_ball: Next Steps

Fix: https://dialpad.atlassian.net/browse/DLT-1731

https://github.com/dialpad/dialtone/assets/87546543/f4e2297a-a2ff-433c-b962-8e6cde97a872


ninamarina:
I think when typing we should wait for spacebar being pressed, if not, maybe the user wanted to choose `#dialtone-vue`, instead of `#dialtone` and thats not possible: 
channels

Slack behaves like that:
channels-slack

braddialpad:
> I think when typing we should wait for spacebar being pressed, if not, maybe the user wanted to choose `#dialtone-vue`, instead of `#dialtone` and thats not possible:

Yeah I think this is best... But also in the case of the / command `/dpm` it triggers immediately.. at least in current product.

braddialpad:
> But also in the case of the / command `/dpm` it triggers immediately.. at least in current product.

I guess we say in the ticket we'll not copy how product currently works and use the space or enter method.

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
```diff
@@ -428,9 +432,70 @@ export default {
...
+      this.insertContent();
       this.addEditorListeners();
     },
 
+    /**
+     * This function is necessary as tiptap doesn't render the content passed
+     * directly through `editor.commands.setContent` the content passed down to it
+     * should be already parsed. So We're parsing the elements into it's corresponding
+     * HTML version before setting it.
+     */
```
braddialpad, line 444: I wonder if this would work better if we used json instead of html. Anyways doesn't matter for now, it's not an option due to how our backend currently works.

braddialpad (approved):
works really nice thanks!

braddialpad:
> I think when typing we should wait for spacebar being pressed, if not, maybe the user wanted to choose `#dialtone-vue`, instead of `#dialtone` and thats not possible:

This seems to only be a problem with channels, mentions work correctly.

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
juliodialpad: I think it would work exactly the same, why do you think it'd work better with json? 🤔

juliodialpad:
> > I think when typing we should wait for spacebar being pressed, if not, maybe the user wanted to choose `#dialtone-vue`, instead of `#dialtone` and thats not possible:
> 
> This seems to only be a problem with channels, mentions work correctly.

Yeah, fixed that.

The only issue I'm having right now is similar to what Nina is having on its PR, related to the input going down instead of space.

It only happens if no text at the beginning tho.

PR #295, opened 2024-05-02
ci: NO-JIRA improve release script

juliodialpad:
# Improve release script

Obligatory GIF

- [x] CI

## :book: Description

- Added `CHANGELOG.md` to path filter change detection.
- Added `fetch-depth: 0` to fetch all the git history.
- Removed the `build` dependency of `release-local` script, it's not necessary actually and it's just making it take longer than it should.
- Improved release script to stop deleting alpha/beta, it was causing issues with git tags.

## :bulb: Context

I was testing releasing alpha/beta versions with the new `--skip-tag` flag and it was failing due to alpha/beta branches were behind the current one as it was deleted and recreated.

I think having the alpha/beta releases on staging/production should not make any damage, but please think about it and let me know if having any concern.

## :pencil: Checklist

For all PRs:

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -22,24 +22,6 @@
         "^build"
       ]
     },
-    "release": {
-      "dependsOn": [
-        "build",
-        "^build"
-      ]
-    },
-    "release-local": {
-      "dependsOn": [
-        "build",
-        "^build"
-      ]
-    },
-    "release-github": {
-      "dependsOn": [
-        "build",
-        "^build"
-      ]
-    },
     "start": {
       "dependsOn": [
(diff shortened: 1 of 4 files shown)


nx.json
```diff
@@ -22,24 +22,6 @@
         "^build"
       ]
     },
-    "release": {
-      "dependsOn": [
-        "build",
-        "^build"
-      ]
-    },
```
juliodialpad, line 30: Renamed this command from `release` to `publish` a while ago, this was a leftover

braddialpad (approved):

PR #298, opened 2024-05-03
docs: NO-JIRA github doc update

braddialpad:
# docs: NO-JIRA github doc update

Obligatory GIF

- [x] Documentation

## :book: Description

- Moved library usage instructions to the top of the readme as that is the first thing most people would want to see
- Removed more obscure steps out of the initial readme and into the readme for that specific package, for example most users will not need to import tokens directly, they will just get them from Dialtone CSS.
- Put quick start as the first section of contributing
- Added some common commands
- Moved all contributing documentation that applied to the monorepo as a whole into the root contributing.md. The content left in the package specific contributing.md's should be specific to that package only. There was a lot of duplication here before this was done.

## :bulb: Context

Did some GitHub documentation cleanup, it has gotten a bit messy and confusing since all the changes recently.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -126,7 +126,7 @@ After you have discussed your change with the Dialtone team, follow these steps 
    - 'visual-test-ready' if your PR includes visual UI changes.
    - 'no-visual-test' if no UI changes.
 
-8. If it's a Vue change, you need to update both dialtone-vue2 and dialtone-vue3 packages. You may sync your changes from Vue 2 to Vue 3 (or vice versa) using the `./scripts/dialtone-vue-sync.sh` script.
+8. If it's a Vue change, you need to update both dialtone-vue2 and dialtone-vue3 packages. You may sync your changes from Vue 2 to Vue 3 (or vice versa) using the `./scripts/dialtone-vue-sync.sh` script. Please see the section in the Dialtone Vue contributing guide: [Dialtone Vue Sync Script](../packages/dialtone-vue3/.github/CONTRIBUTING.md#dialtone-vue-sync-script) for details on usage of the script.
 9. Once your changes have been approved, you may squash your branch into staging.
 
 Once your change is in `staging` it will go live with the next Dialtone Vue release.
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -96,7 +96,7 @@ This is why it is important to evaluate the change carefully and asses its impac
 Any new components or updates to existing components require the following:
 
 - Unit tests covering the entire change.
-- Storybook documentation including a live rendered component via controls and MDX. See [the documentation](https://dialtone.dialpad.com/vue/?path=/story/docs-storybook-getting-started--page)
+- Storybook documentation including a live rendered component via controls and MDX. See [the documentation](https://dialtone.dialpad.com/vue/?path=/docs/docs-storybook-getting-started--docs)
 - Component is accessible according to requirements.
   - Navigable by keyboard.
   - Read by a screen reader.
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,8 +1,10 @@
 # Contributing
 
+Thanks for your interest in contributing to Dialtone! Please take a moment to review this document before submitting a pull request.
+
 Dialtone exists as a monorepo containing multiple packages. Libraries are stored in the packages folder, and deployable documentation sites are stored in the apps folder. Currently the Dialtone Vue storybook documentation sites are stored in the same folder as their libraries due to a limitation of storybook, however this may change in the future as it makes more logical sense for them to be stored in the apps folder.
 
-Each package has its own contributing guide, please follow the links to find the actual contributing guide you might need.
+This contributing guide contains guidelines and processes that apply to all packages of the Dialtone monorepo. There are also contributing guides specific to each package, so make sure to also read the contributing guide for the package you are contributing to in addition to this one:
 
 ## Dialtone CSS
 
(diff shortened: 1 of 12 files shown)


.github/CONTRIBUTING.md
```diff
@@ -19,3 +21,219 @@ Each package has its own contributing guide, please follow the links to find the
...
+
+- **Fix:** fixes a technical defect, documentation typo, or Figma symbol defect.
+- **Enhancement:** extends an existing style or component without changing the underlying architecture or behavior.
+- **New feature:** adds something new, like a component.
+
+### What belongs in Dialtone?
+
+There are a couple important considerations when thinking about contributing to Dialtone. The first is to remember that Dialtone strives to offer styles, components, or patterns intended to be shared by multiple teams or features. Generally, one-off or first-time elements (i.e. snowflakes) aren't a great fit, though there may be the occasional exception.
+
+The second is to check with the Dialtone team (in the #dialtone Dialpad channel) to ensure the contribution isn't already requested, planned, or even complete. You may also see our [Jira backlog](https://dialpad.atlassian.net/jira/software/c/projects/DLT/boards/548/backlog) for upcoming work, and our [Jira timeline](https://dialpad.atlassian.net/jira/software/c/projects/DLT/boards/548/timeline) for our long term quarterly roadmap.
```
juliodialpad, line 53: Can we include a link to the #dialtone channel?

.github/CONTRIBUTING.md
```diff
@@ -19,3 +21,219 @@ Each package has its own contributing guide, please follow the links to find the
...
+- **Making new additions to the component API**: for example, adding a new prop to a component that triggers a different flow of behaviour or output, or adding a new emit event to detect specific changes.
+- **Refactoring**: changing the internal code of a component without changing any external behaviour.
+
+Some updates can unintentionally cause breaking changes, or cause changes to a component's output behaviour or functionality without necessary breaking it.
+This is why it is important to evaluate the change carefully and asses its impact on existing usage.
+
+Any new components or updates to existing components require the following:
+
+- Unit tests covering the entire change.
+- Storybook documentation including a live rendered component via controls and MDX. See [the documentation](https://dialtone.dialpad.com/vue/?path=/story/docs-storybook-getting-started--page)
```
juliodialpad: Broken link.

```suggestion
- Storybook documentation including a live rendered component via controls and MDX. See [the documentation](https://dialtone.dialpad.com/vue/?path=/docs/docs-storybook-getting-started--docs)
```

.github/CONTRIBUTING.md
```diff
@@ -19,3 +21,219 @@ Each package has its own contributing guide, please follow the links to find the
...
+  - Minimum contrast ratio.
+- Changes must be made for Vue 2 as well as Vue 3, `dialtone-vue2` and `dialtone-vue3` package folders respectively
+- Unit tests are passing locally.
+  - `nx test dialtone-vue2` or `nx test dialtone-vue3`
+- Linters are passing locally.
+  - `nx lint dialtone-vue2` or `nx lint dialtone-vue3`
+- Library builds locally.
+  - `nx build dialtone-vue2` or `nx build dialtone-vue3`
+- Documentation builds locally.
+  - `nx storybook:build dialtone-vue2` or `nx storybook:build dialtone-vue3`
```
juliodialpad, line 112: Should we enforce running this commands before pushing?

.github/CONTRIBUTING.md
```diff
@@ -19,3 +21,219 @@ Each package has its own contributing guide, please follow the links to find the
...
+3. Create a personal branch to make your change off of `staging` with `git checkout -b my-change-branch`. We use kebab-case for branch names.
+4. Make and commit your changes. Note our commit message conventions in [COMMIT_CONVENTION.md].
+5. Push your branch to remote. `git push -u origin my-change-branch`.
+6. Create a pull request into the `staging` branch, reviewers will be automatically added and notified of your PR.
+7. Set the label on your PR:
+
+- 'visual-test-ready' if your PR includes visual UI changes.
+- 'no-visual-test' if no UI changes.
+
+<!-- markdownlint-disable MD029 -->
```
juliodialpad: Is this necessary? Maybe fixing the previous list indentation would be enough so we can remove this annotation.

.github/CONTRIBUTING.md
```diff
@@ -19,3 +21,219 @@ Each package has its own contributing guide, please follow the links to find the
...
+8. If it's a Vue change, you need to update both dialtone-vue2 and dialtone-vue3 packages. You may sync your changes from Vue 2 to Vue 3 (or vice versa) using the `./scripts/dialtone-vue-sync.sh` script.
```
juliodialpad: We should add a comment about some differences between vue2 and vue3 that need to be done manually. Or maybe saying that the sync script only copies over the changes but still need to make sure it's working on vue3.

README.md
```diff
@@ -83,95 +194,103 @@ and build any dependency needed before running your command.
...
-nx start:dialtone
+nx build
 ```
 
-This will start the documentation site and watch the library changes, so it is live updated.
+##### Run all Vue unit tests
 
-Access the local server at `http://localhost:4000`
+```bash
+nx test:vue
```
juliodialpad: Just running `nx test` will run the tests as it is listed on `./project.json` and also that's the right way to do it because we can define if something needs to be build before running tests. running `nx test:vue` will run the tests directly without looking into dependencies and that might cause issues.

```suggestion
nx test
```

juliodialpad (changes requested):
Really nice job Brad! 

A couple nitpicks found, thanks!

.github/CONTRIBUTING.md
braddialpad: That's a good idea, I think we can run "only changed tests" in lint-staged now that we are on vitest

it's `vitest related`

https://vitest.dev/guide/cli#vitest-related

.github/CONTRIBUTING.md
braddialpad: unfortunately I don't think there's any way to link to channels currently. You can link to threads but not channels

README.md
braddialpad: oh sweet didn't know it worked like that. Yeah thats definitely the way to go.

iropolo (approved):

.github/CONTRIBUTING.md
braddialpad: weirdly if I put 3 spaces on the prior list it fixes it.

https://stackoverflow.com/questions/50163820/how-to-write-an-ordered-list-which-contains-nested-unordered-lists-github-markd

.github/CONTRIBUTING.md
braddialpad: Added a whole new section on using the script and differences between Vue 2 and Vue 3 [74a99c1](https://github.com/dialpad/dialtone/pull/298/commits/74a99c15e2ab31f4606c13bf350fd9d3d1d48d89)

braddialpad:
Added dialtone vue css import to readmes as mentioned earlier today.

PR #305, opened 2024-05-08
feat(message-input): NO-JIRA add keydown event

bianca-artola-dialpad:
# feat(message-input): NO-JIRA add keydown event

Obligatory GIF

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket

NO-JIRA

## :book: Description

This pull request adds keydown event on message input component.

## :bulb: Context
We need to prevent the backspace key default event when using the input component on the new message flow because if not it moves to the previous input and in this case is not the expected behavior.
Example of current behavior:

https://github.com/dialpad/dialtone/assets/144126346/e0933c12-1d5a-4b76-b068-10311c27a47c

## :pencil: Checklist

For all PRs:

### DIFF
--- a/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.stories.js
+++ b/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.stories.js
@@ -145,6 +145,7 @@ export const argsData = {
   onNoticeClose: action('notice-close'),
   onSkinTone: action('skin-tone'),
   onCancel: action('cancel'),
+  onHandleKeydown: action('keydown'),
 };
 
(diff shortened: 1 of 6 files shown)


braddialpad (changes requested):
This is a problem with the new message page rather than the message input. Why is backspace deleting the contact when the message input is in focus? It should only delete the contact while the focus is on the contact input. Likely the backspace event handler is on the whole page rather than just the relevant input.

bianca-artola-dialpad:
yeap you're right. thanks for the feedback :)

PR #308, opened 2024-05-09
fix: DLT-1710 unexpected emoji on links

juliodialpad:
# Fix unexpected emoji on links

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1710

## :book: Description

- Fixed a warning pnpm related to typescript version.
- Updated the emoji regex to detect if the emoji is at the start of the line or has a space before.
- Updated components that use emojiRegex

## :bulb: Context

If a string such as a URL contained a valid emoji shortcode it'd be replaced but that's not intended.

## :pencil: Checklist

For all PRs:

For all Vue changes:

Before
Screenshot 2024-05-09 at 12 34 41 p m

After
Screenshot 2024-05-09 at 12 34 06 p m


braddialpad:
Looks like some test / deploy preview failures.

juliodialpad:
> Looks like some test / deploy preview failures.

It's due to some issue with slash commands on Vue 3, fixing

packages/dialtone-vue2/common/emoji/index.js
```diff
@@ -15,6 +15,8 @@ export let emojiFileExtensionLarge = '.png';
 
 export const emojiJson = emojiJsonLocal;
 
+export const emojiShortCodeRegex = /(^| ):\w+:/g;
```
braddialpad: It's a pretty common case to put multiple emojis beside each other without spaces like so :smile::smile::smile::smile:. Maybe we can do a special case to not escape when there is a "double colon"

braddialpad: Also if you use a unicode emoji in emoji-text-wrapper it allows you to do this, but not with shortnames

packages/dialtone-vue2/components/rich_text_editor/extensions/emoji/emoji.js
```diff
@@ -1,29 +1,29 @@
...
 import { PluginKey } from '@tiptap/pm/state';
 
 import Suggestion from '@tiptap/suggestion';
 import suggestionOptions from './suggestion';
 
 export const EmojiPluginKey = new PluginKey('emoji');
 
-const inputShortCodeRegex = /:\w+:$/;
-export const emojiShortCodeRegex = /:\w+:/g;
+const inputShortCodeRegex = /(^| )(:\w+:)$/;
```
braddialpad: Is it necessary to have this? Maybe we can just make our const a regex string `"(^| )(:\w+:)"` and then append a $ to it here? then use `new RegExp` to turn the string into a regex? Not that big of deal though.

packages/dialtone-vue2/common/emoji/index.js
juliodialpad: > It's a pretty common case to put multiple emojis beside each other without spaces like so 😄😄😄😄. Maybe we can do a special case to not escape when there is a "double colon"

That implied the use of `Positive Lookbehind` which IIRC we just had an issue with it on older iOS devices.

juliodialpad: > Also if you use a unicode emoji in emoji-text-wrapper it allows you to do this, but not with shortnames

It's using the same emojiShortCodeRegex so this should be fixed too.

packages/dialtone-vue2/components/rich_text_editor/extensions/emoji/emoji.js
juliodialpad: > Is it necessary to have this? Maybe we can just make our const a regex string `"(^| )(:\w+:)"` and then append a $ to it here? then use `new RegExp` to turn the string into a regex? Not that big of deal though.

Yes, most places use `emojiShortCodeRegex` which already includes a global flag.

That's the only place where we need `$` at the end.

I know is not ideal but seems better that having a string and having to call `new RegExp()` everywhere adding the `g` flag.

iropolo (approved):

packages/dialtone-vue2/common/emoji/index.js
braddialpad: > That implied the use of `Positive Lookbehind` which IIRC we just had an issue with it on older iOS devices.

hmm wasn't that negative lookbehind?

braddialpad (approved):

packages/dialtone-vue2/common/emoji/index.js
juliodialpad: > > That implied the use of `Positive Lookbehind` which IIRC we just had an issue with it on older iOS devices.
> 
> hmm wasn't that negative lookbehind?

Yeah, I think the issues were with negative lookbehind but they both have the same compatibility. Let's just keep an eye on this if some issues come related to it.

PR #309, opened 2024-05-10
fix(general-row): give example of contact center row using general row

braddialpad:
# fix(general-row): give example of contact center row using general row

Obligatory GIF

## :book: Description

Addon for @tiagox's PR. Reworked the general row to support a new slot called "secondaryLabel" which can contain the contact center status label, and the changes needed to the action button as well to show an "expand" button.

@tiagox You will still have to implement this into contact_centers_row using general_row, however I have provided a live example in contact_centers_row_default.story.vue that you can check out. I think this should cover all the issues you were having but please let me know if there's anything remaining, or if something I've done here is not going to work correctly with your requirements.


braddialpad:
Closing, parent PR was merged in.

PR #311, opened 2024-05-14
fix(datepicker): DP-89398 change tooltip fallback positions

braddialpad:
# fix(datepicker): change tooltip fallback positions

## Obligatory GIF (super important!)

Obligatory GIF

## :hammer_and_wrench: Type Of Change

These types will increment the version number on release:

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DP-89398

## :book: Description

Tooltip was overlapping the month name when there was not enough horizontal space for it to display on top. Added custom fallbackPlacements to remedy this.

## :bulb: Context

See Jira ticket, datepicker implementation within integrations iframe.

## :pencil: Checklist

For all PRs:

- [x] I have ensured no private Dialpad links or info are in the code or pull request description (Dialtone is a public repo!).
- [x] I have reviewed my changes.
- [x] I have added all relevant documentation.
- [x] I have considered the performance impact of my change.

For all Vue changes:

- [x] I have made my changes in Vue 2 and Vue 3. Note: you may sync your changes from Vue 2 to Vue 3 (or vice versa) using the `./scripts/dialtone-vue-sync.sh` script. Read docs here: [Dialtone Vue Sync Script](../packages/dialtone-vue3/.github/CONTRIBUTING.md#dialtone-vue-sync-script)

## :crystal_ball: Next Steps

Release and notify in jira ticket

PR #314, opened 2024-05-15
fix(tooltip): DLT-1757 round button alignment

braddialpad:
# fix(tooltip): round button alignment

Obligatory GIF

- [x] Fix

## :book: Description

- Changed tooltip arrows from custom CSS to the built in Tippy arrow SVG, which it automatically aligns correctly
- No longer rendering tooltips with a custom tippy render function. (Why were we doing this in the first place?)
- Still using the old style custom render function for popover, because that one will be a ton of work to fix.
- When rendering this way, the default offset had to change. I will have to check any custom offsets set in product because they may be incorrect after this change.
- Remove aria attrs because they are handled by tippy
- Remove lazy show because that's handled by tippy

## :bulb: Context

Noticed this issue in my prior PR https://github.com/dialpad/dialtone/pull/311

The tooltip arrow was not rendering correctly on circle buttons because we were using custom css from who knows how long ago, and it was rendering the arrow in a fixed position. Tippy renders it's arrow dynamically to be in the center so I changed it to that instead. In order to do this I had to get rid of our custom override of the tippy render function, which seemed to have no real purpose, and it disabled a ton of default tippy functionality and options that we could have been taking advantage of.

There are a ton of other things we can improve in this component and the popover one now, but I'm trying to keep the scope of this somewhat reasonable.

BEFORE

Screenshot 2024-05-15 at 1 40 03 PM

AFTER

Screenshot 2024-05-15 at 1 37 36 PM

## :pencil: Checklist

For all PRs:

For all Vue changes:

For all CSS changes:

## :crystal_ball: Next Steps

check usage of offset in product.

See Context

## :link: Sources

https://atomiks.github.io/tippyjs/v6/all-props


braddialpad:
@francisrupert the shape of the arrow changed slightly since it's now just using the built in tippy arrow. Should we make a custom SVG to make it look like the previously used arrow? For reference this is the svg for the default tippy arrow.

```
<svg width="16" height="6" xmlns="http://www.w3.org/2000/svg"><path d="M0 6s1.796-.013 4.67-3.615C5.851.9 6.93.006 8 0c1.07-.006 2.148.887 3.343 2.385C14.233 6.005 16 6 16 6H0z"></svg>
```

packages/dialtone-vue2/components/tooltip/tooltip.vue
```diff
@@ -522,19 +511,25 @@ export default {
...
   }
 }
+
+.tippy-box > .tippy-svg-arrow {
+  fill: var(--dt-color-surface-contrast);
+}
+
+.tippy-box[data-animation='fade'][data-state='hidden'] {
+  opacity: 0;
+}
```
iropolo: Could we move this to dialpad css?, since this is not an scoped style, we will be importing `@import 'tippy.js/dist/svg-arrow.css';` globally when this component is being initialized.

iropolo:
Noticed some slight difference that tooltip not being centered with the content:

BEFORE:
image

AFTER:
image

packages/dialtone-vue2/components/tooltip/tooltip.test.js
```diff
@@ -2,7 +2,6 @@ import { createLocalVue, mount } from '@vue/test-utils';
 import DtTooltip from './tooltip.vue';
 import {
   TOOLTIP_KIND_MODIFIERS,
-  TOOLTIP_DIRECTIONS,
```
iropolo: Seems like direction css will be managed by tippy.

Im thinking if somehow we can still test if it works correctly

iropolo: image

I know sometimes test this can be tricky but if we could reach it we can get the `data-placement` value and match it with ours.

Thats the generated tooltip when you hover the element.

packages/dialtone-vue2/components/tooltip/tooltip.vue
```diff
@@ -12,25 +12,19 @@
...
       <!-- @slot Slot for the anchor element -->
       <slot
         name="anchor"
       />
     </span>
-    <dt-lazy-show
+    <div
       :id="id"
       ref="content"
-      :show="isVisible"
```
iropolo, line 25: `isVisible` is a leftover should be removed.

packages/dialtone-vue2/components/tooltip/tooltip.vue
```diff
@@ -12,25 +12,19 @@
...
-      role="tooltip"
-      aria-hidden="false"
       data-qa="dt-tooltip"
-      appear
-      :transition="transition"
```
iropolo, line 30: `transition` prop is not being used anymore, seems like `fade` is the default from tippy.
Either we make it work again or definitely removed it from the component and storie.

iropolo:
Seems like there is some issues in the test.

```
it('should set default classes', () => {
  expect(tooltip.classes('d-tooltip__arrow-tippy--top')).toBe(true);
});
```
This one seems like it is not necessary anymore since we are not handling the classes anymore.

Then there is some issues related to `show` prop or `isShown` watcher.

packages/dialtone-vue2/components/tooltip/tooltip.vue
braddialpad: Yeah I think it is unnecessary... nobody ever uses anything other than the default, will remove.

packages/dialtone-vue2/components/tooltip/tooltip.vue
braddialpad: The problem is tippy.js is not a dependency of dialtone-css, so it won't be able to import it. Scoping it also won't work because the tooltip renders outside of the component at the document body. I think it's probably okay just to leave global 🤷

packages/dialtone-vue2/components/tooltip/tooltip.test.js
braddialpad: hmm yeah think we could do that.

packages/dialtone-vue2/components/tooltip/tooltip.vue
braddialpad: thanks

packages/dialtone-vue2/components/tooltip/tooltip.vue
braddialpad: nevermind, I think there is still some value in being able to set transition to false for no transition, so I will make it work like this.

iropolo (approved):
Cool and useful improvement, thanks.

packages/dialtone-vue2/components/tooltip/tooltip.test.js
braddialpad: Added placement tests back in

braddialpad:
Fixed svg shape, merging.

PR #322, opened 2024-05-21
fix(tooltip): DP-99352 correct offsets, trim empty check

braddialpad:
# fix(tooltip): correct offsets, trim empty check

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DP-99352

## :book: Description

- Fixed offsets in Dialtone after the recent tooltip changes
- Added trim to the tooltip empty check

## :bulb: Context

Default offset changed after the recent tooltip change, so manual offsets had to be fixed. This was done in product but not in dialtone itself. Fixed the few that were in here.

Empty check was not trimmed so spaces and newlines were getting marked as "not empty" when they should have been causing blank tooltips to show.

## :pencil: Checklist

For all PRs:

For all Vue changes:

## :crystal_ball: Next Steps

CP this into beta to fix reported issues.

before:
Screenshot 2024-05-21 at 1 35 46 PM

after:
Screenshot 2024-05-21 at 1 36 02 PM

### DIFF
--- a/packages/dialtone-vue2/components/tooltip/tooltip.test.js
+++ b/packages/dialtone-vue2/components/tooltip/tooltip.test.js
@@ -107,6 +107,15 @@ describe('DtTooltip tests', () => {
         expect(tooltip.textContent).toBe('Test message');
       });
 
+      describe('When tooltip content is a space character', () => {
+        it('should not render the content', async () => {
+          wrapper.destroy();
+          mockProps = { show: true, message: ' ' };
+          updateWrapper();
+          expect(tippyContent.getAttribute('data-state')).not.toBe('visible');
+        });
+      });
+
       describe('When inverted is true', () => {
         it('should have the inverted class set', async () => {
           await wrapper.setProps({ show: true, inverted: true });
--- a/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
+++ b/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
@@ -53,7 +53,7 @@
           v-if="showImagePicker"
           placement="top-start"
           :message="showImagePicker.tooltipLabel"
-          :offset="[-4, -4]"
+          :offset="[-4, 12]"
         >
           <template #anchor>
             <dt-button
@@ -141,7 +141,7 @@
           placement="top-end"
           :enabled="characterLimitTooltipEnabled"
           :message="showCharacterLimit.message"
-          :offset="[10, -8]"
+          :offset="[10, 8]"
         >
           <template #anchor>
             <p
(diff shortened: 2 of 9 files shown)


packages/dialtone-vue2/components/tooltip/tooltip.test.js
```diff
@@ -107,6 +107,15 @@ describe('DtTooltip tests', () => {
         expect(tooltip.textContent).toBe('Test message');
       });
 
+      describe('When tooltip content is a space character', () => {
+        it('should not render the content', async () => {
+          wrapper.destroy();
```
juliodialpad, line 112: why are we destroying the wrapper here?

packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
```diff
@@ -170,12 +170,12 @@
 
         <!-- Send button -->
         <dt-tooltip
-          v-if="showSend"
+          v-if="true"
```
juliodialpad: If this is always true, isn't it better to remove it?

juliodialpad (changes requested):
Couple comments

packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
braddialpad: agh thanks for noticing that, that was for testing only, will change back.

packages/dialtone-vue2/components/tooltip/tooltip.test.js
braddialpad: I think it's only an issue in this case because of the tooltip writing to the body of the JSDom instance and that will persist between tests unless you destroy. It causes a bunch of errors if I don't do it. 

There is wrapper.destroy in afterEach but that doesn't get run when there are nested describes like this.

it runs the first beforeEach:
```
mockProps = { show: true };
updateWrapper();
```

then after in the test we do:
```
wrapper.destroy();
mockProps = { show: true, message: ' ' };
updateWrapper();
```

Seems to cause problems without the wrapper.destroy() because the previously mounted updateWrapper in the beforeEach hasn't been destroyed, and the jsdom instance is still there.

braddialpad: might be because we are getting the content like so:

```
.tippy-content
```

so if there are multiple tippys rendered it will just take the first one.

packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
braddialpad: fixed.

ninamarina:
I still see this issue in the preview for the doc site:
image

But maybe that's a different problem?

The storybook looks good

braddialpad:
> I still see this issue in the preview for the doc site

Ah yeah, had to import the dialtone vue css into the docsite

iropolo (approved):

PR #329, opened 2024-05-24
fix: DLT-1731 skin tone emojis

juliodialpad:
# Detect skin tone emojis on rich text editor emoji plugin

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1731

## :book: Description

- Changed [emoji-regex](https://github.com/mathiasbynens/emoji-regex) dependency to [regex-combined-emojis](https://github.com/sweaver2112/Regex-combined-emojis) that seems to be more maintained and up-to-date.
- Updated emoji input rule to return shortcode instead of emoji, if we already found the emoji and know that it is valid, why we would need to send unicode and having to transform it again?

## :bulb: Context

Skin tone emojis were not being detected correctly.

## :pencil: Checklist

For all PRs:

For all Vue changes:

### Before

https://github.com/dialpad/dialtone/assets/87546543/7f9ddcd6-139e-42fe-9860-3b0ef5f7aa24

### After

https://github.com/dialpad/dialtone/assets/87546543/cf510018-0286-407a-b3ce-31918e9aab44

### DIFF
--- a/packages/dialtone-vue3/common/emoji/index.js
+++ b/packages/dialtone-vue3/common/emoji/index.js
@@ -1,6 +1,7 @@
-import emojiRegex from 'emoji-regex';
+import { emojiPattern } from 'regex-combined-emojis';
 import { emojisIndexed } from '@dialpad/dialtone-emojis';
 
+export const emojiRegex = new RegExp(emojiPattern, 'g');
 export const emojiVersion = '8.0';
 export const defaultEmojiAssetUrl = 'https://cdn.jsdelivr.net/joypixels/assets/' + emojiVersion + '/png/unicode/32/';
 export let customEmojiAssetUrl = null;
@@ -174,7 +175,6 @@ export function codeToEmojiData (code) {
     return shortcodeToEmojiData(code);
   } else {
     const unicodeString = unicodeToString(code);
-
     const result = emojiJson[unicodeString];
     if (result) result.key = unicodeString;
     return result;
@@ -201,7 +201,7 @@ export function filterValidShortCodes (shortcodes) {
 // removes duplicates
 // @returns {string[]}
 export function findEmojis (textContent) {
-  const matches = [...textContent.matchAll(emojiRegex())];
+  const matches = [...textContent.matchAll(emojiRegex)];
   const emojis = matches.length ? matches.map(match => match[0]) : [];
   return new Set(emojis);
 }
--- a/apps/dialtone-documentation/docs/.vuepress/baseComponents/tokens/AllTokens.vue
+++ b/apps/dialtone-documentation/docs/.vuepress/baseComponents/tokens/AllTokens.vue
@@ -14,7 +14,7 @@
       &OpenCurlyDoubleQuote;{{ searchCriteria }}&CloseCurlyDoubleQuote;
     </strong>
   </div>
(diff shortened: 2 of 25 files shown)


packages/dialtone-vue3/common/emoji/index.js
```diff
@@ -1,6 +1,7 @@
-import emojiRegex from 'emoji-regex';
+import { emojiPattern } from 'regex-combined-emojis';
```
braddialpad, line 1: I'm a bit skeptical about this library, it's 5 times larger than emoji-regex was and only has 8 weekly downloads (emoji-regex in comparison has 101,125,859 weekly downloads). Are you sure we were not just using emoji-regex incorrectly? It seems like it should be able to match multiple codepoints just looking at the readme.

juliodialpad: You're right, it doesn't have many weekly downloads and it's bigger in size.

The reason I choose that library instead of `emoji-regex` was due to this:

> Note: because the regular expression **has the global flag set**, this module
> exports a function that returns the regex rather than exporting the regular
> expression itself, **to make it impossible to** (accidentally) **mutate the
> original regular expression**.

We can use that regular expression to detect on paste, but it cannot be used to detect on user input (selecting an emoji from the emoji selector) because we need to remove the global flag an append the `$` to match the end of the string.

Also the `emoji-regex` library was published 7 moths ago.

Screenshot 2024-05-27 at 11 27 13 a m

And the `regex-combined-emojis` its being updated more regularly.

Screenshot 2024-05-27 at 11 27 35 a m

braddialpad: The emoji-regex library only needs to update when the Unicode standard for emojis updates. The most recent update to unicode was September 2023 with Unicode version 15.1, so they haven't had to publish since then.

braddialpad: > We can use that regular expression to detect on paste, but it cannot be used to detect on user input (selecting an emoji from the emoji selector) because we need to remove the global flag an append the `$` to match the end of the string.

Hmm maybe we just pull ourselves from https://www.npmjs.com/package/emoji-test-regex-pattern. This is where emoji-regex gets it's data from. Then we can modify ourselves as we need. see `emoji-test-regex-pattern/dist/latest/javascript.txt`

juliodialpad: > > We can use that regular expression to detect on paste, but it cannot be used to detect on user input (selecting an emoji from the emoji selector) because we need to remove the global flag an append the `$` to match the end of the string.
> 
> Hmm maybe we just pull ourselves from https://www.npmjs.com/package/emoji-test-regex-pattern. This is where emoji-regex gets it's data from. Then we can modify ourselves as we need. see `emoji-test-regex-pattern/dist/latest/javascript.txt`

What's the difference of using that instead of `regex-combined-emojis`? It is even bigger in size and it'd need more work to parse from `javascript.txt` file and create the regex from the output.

braddialpad: Yeah guess its fine 🤷‍♂️

braddialpad (approved):

PR #331, opened 2024-05-25
fix(rich-text-editor): DLT-1752 remove custom parsing

braddialpad:
# fix(rich-text-editor): DLT-1752 remove custom parsing

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1752

## :book: Description

- Removed the code we put in previously to parse mentions on programmatic input.
- Added an async wait into the mentions example.

## :bulb: Context

What we were doing here with the parsing is not really the way TipTap is supposed to work. TipTap fully supports doing this on it's own, however you have to pass in input that it understands which is HTML or JSON input. see: https://tiptap.dev/docs/editor/guide/output

To do this properly we have to change our backend to support TipTaps html or json output and use that to populate the input. I talked to Bianca and they have agreed to update the backend to be able to handle this. For now we will just not parse links / mentions live in the input which is actually the way it works currently in production.

## :pencil: Checklist

For all PRs:

For all Vue changes:

## :crystal_ball: Next Steps

Check if all TipTap formatting will work correctly after the server side changes have been made

## :link: Sources

https://tiptap.dev/docs/editor/guide/output
https://tiptap.dev/docs/editor/api/utilities/html#generate-html-from-json

### DIFF
--- a/packages/dialtone-vue2/components/rich_text_editor/mention_suggestion.js
+++ b/packages/dialtone-vue2/components/rich_text_editor/mention_suggestion.js
@@ -31,7 +31,7 @@ const CONTACT_LIST = [
     avatarSrc: 'https://avatars.githubusercontent.com/u/13851061?s=460&u=1f1b5b0b5b2b2b2b2b2b2b2b2b2b2b2b2b2b2b&v=4',
   },
   {
-    id: 'ignacio.Ropolo',
+    id: 'ignacio.ropolo',
     name: 'Ignacio Ropolo',
     avatarSrc: 'https://avatars.githubusercontent.com/u/13851061?s=460&u=1f1b5b0b5b2b2b2b2b2b2b2b2b2b2b2b2b2b2b&v=4',
   },
--- a/packages/dialtone-vue2/components/rich_text_editor/mention_suggestion.js
+++ b/packages/dialtone-vue2/components/rich_text_editor/mention_suggestion.js
@@ -1,26 +1,53 @@
 /* eslint-disable max-len */
+const CONTACT_LIST = [
+  {
+    id: 'test.person',
+    name: 'Test Person',
+    avatarSrc: 'https://avatars.githubusercontent.com/u/13851061?s=460&u=1f1b5b0b5b2b2b2b2b2b2b2b2b2b2b2b2b2b2b&v=4',
+  },
+  {
+    id: 'test.person2',
+    name: 'Test Person 2',
+    avatarSrc: 'https://avatars.githubusercontent.com/u/13851061?s=460&u=1f1b5b0b5b2b2b2b2b2b2b2b2b2b2b2b2b2b2b&v=4',
+  },
+  {
+    id: 'test.person3',
+    name: 'Test Person 3',
+    avatarSrc: 'https://avatars.githubusercontent.com/u/13851061?s=460&u=1f1b5b0b5b2b2b2b2b2b2b2b2b2b2b2b2b2b2b&v=4',
+  },
+  {
+    id: 'brad.paugh',
+    name: 'Brad Paugh',
+    avatarSrc: 'https://avatars.githubusercontent.com/u/13851061?s=460&u=1f1b5b0b5b2b2b2b2b2b2b2b2b2b2b2b2b2b2b&v=4',
+  },
+  {
+    id: 'bradley.hawkins',
+    name: 'Bradley Hawkins',
+    avatarSrc: 'https://avatars.githubusercontent.com/u/13851061?s=460&u=1f1b5b0b5b2b2b2b2b2b2b2b2b2b2b2b2b2b2b&v=4',
+  },
+  {
+    id: 'julio.ortega',
+    name: 'Tico Ortega',
+    avatarSrc: 'https://avatars.githubusercontent.com/u/13851061?s=460&u=1f1b5b0b5b2b2b2b2b2b2b2b2b2b2b2b2b2b2b&v=4',
+  },
+  {
+    id: 'ignacio.Ropolo',
+    name: 'Ignacio Ropolo',
+    avatarSrc: 'https://avatars.githubusercontent.com/u/13851061?s=460&u=1f1b5b0b5b2b2b2b2b2b2b2b2b2b2b2b2b2b2b&v=4',
+  },
+  {
+    id: 'nina.repetto',
+    name: 'Nina Repetto',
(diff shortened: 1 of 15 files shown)


packages/dialtone-vue2/components/rich_text_editor/mention_suggestion.js
```diff
@@ -1,26 +1,53 @@
...
+    name: 'Bradley Hawkins',
+    avatarSrc: 'https://avatars.githubusercontent.com/u/13851061?s=460&u=1f1b5b0b5b2b2b2b2b2b2b2b2b2b2b2b2b2b2b&v=4',
+  },
+  {
+    id: 'julio.ortega',
+    name: 'Tico Ortega',
+    avatarSrc: 'https://avatars.githubusercontent.com/u/13851061?s=460&u=1f1b5b0b5b2b2b2b2b2b2b2b2b2b2b2b2b2b2b&v=4',
+  },
+  {
+    id: 'ignacio.Ropolo',
```
juliodialpad:
```suggestion
    id: 'ignacio.ropolo',
```

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.stories.js
```diff
@@ -13,7 +13,7 @@ import slashCommandSuggestion from './slash_command_suggestion';
 
 // Default Prop Values
 export const argsData = {
-  value: `I am not a standalone component, please use Message Input instead :v_tone3::robot:!`,
+  value: '<p>I am not a standalone component, please use Message Input instead<emoji-component code=":v_tone3:" id="270c-1f3fd"></emoji-component><emoji-component code=":robot:" id="1f916"></emoji-component> !</p>',
```
juliodialpad: Are those `id` attributes necessary?

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
```diff
@@ -504,79 +502,9 @@ export default {
...
-
-      matches.forEach(match => {
-        const mention = suggestions.find(({ id }) => id === match[1]);
-        this.internalValue = this.internalValue.replace(`@${match[1]}`, ` <mention-component name="${mention.name}" id="${mention.id}"></mention-component>`);
-      });
-    },
-
-    destroyEditor () {
-      this.editor.destroy();
-    },
```
juliodialpad: We don't need this function anymore? or we weren't using it anywere?

juliodialpad (commented):
it's working pretty good! but I have some doubts regarding the changes: 

- Why are we removing the possibility to paste and input channels and mentions?.
- Should I do the same with emoji to keep the same behavior on all plugins?
- Should we update the example to output the HTML instead of plain text so we can copy and paste the output to parse the examples?

juliodialpad:
Found kind of a bug but I'm not able to reproduce it consistently.

Sometimes you loose the keyboard navigation ability until you click an option from the dropdown, It happened sometimes when copy/pasting text and then trying to mention a channel/mention but not happening always so not sure if it's a bug or not.

packages/dialtone-vue2/components/rich_text_editor/mention_suggestion.js
braddialpad: lol I swear I already fixed this... must have not committed it or something.

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.stories.js
braddialpad: hmm it's definitely an attribute in the emoji extension but doesn't seem like it's used.. I'll look into it a bit more.

braddialpad:
> Sometimes you loose the keyboard navigation ability until you click an option from the dropdown, It happened sometimes when copy/pasting text and then trying to mention a channel/mention but not happening always so not sure if it's a bug or not.

Hmm I think I got this once or twice too.. Will try and repro it.

braddialpad:
> * Why are we removing the possibility to paste and input channels and mentions?.

Because the backend doesn't understand the data that TipTap outputs (currently). If you enter a mention in the current dialpad message input you'll see it only outputs as text (unstyled). It doesn't actually get styled as a link until it goes through the server and displays as an output chat message.

> * Should I do the same with emoji to keep the same behavior on all plugins?

I'm not quite sure what we should do about emoji... It's a bit different because we have the "emoji-text-wrapper" component, and the emoji outputs using our custom component.

> * Should we update the example to output the HTML instead of plain text so we can copy and paste the output to parse the examples?

Yeah I think so, that's a good idea.

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
braddialpad: Hmm.. think I didn't mean to delete this.

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.stories.js
braddialpad: Don't think it's necessary, removed.

braddialpad:
> Found kind of a bug but I'm not able to reproduce it consistently.

Do you know if it was on Vue 2 or Vue 3 you were getting this problem?

juliodialpad:
> > Found kind of a bug but I'm not able to reproduce it consistently.
> 
> 
> 
> Do you know if it was on Vue 2 or Vue 3 you were getting this problem?

I was testing on Vue 2

braddialpad:
> Sometimes you loose the keyboard navigation ability until you click an option from the dropdown, It happened sometimes when copy/pasting text and then trying to mention a channel/mention but not happening always so not sure if it's a bug or not.

hmm tried a bunch on both vue 2 and vue 3 wasn't able to repro since the recent fixes I made. Let me know if it's still happening for you

packages/dialtone-vue3/components/rich_text_editor/rich_text_editor_default.story.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <dt-rich-text-editor
-    v-model="$attrs.modelValue"
+    v-model="value"
```
juliodialpad, line 3: why not using `$attrs.modelValue` directly?

juliodialpad (approved):
Looking good! just left a comment, please resolve it before merging

packages/dialtone-vue3/components/rich_text_editor/rich_text_editor_default.story.vue
braddialpad: `$attrs.modelValue` is readonly, and v-model tries to change it. It is common for us to do this with v-model. See `dropdown_default.story.vue` for example.

PR #339, opened 2024-05-30
ci: NO-JIRA add sleep to avoid hitting rate limit on release

braddialpad:
Throttle releases since we are sometimes hitting the rate limit


juliodialpad (approved):
Thanks!

PR #347, opened 2024-06-04
docs(empty-state): NO-JIRA restructure and content refinement

francisrupert:
# Empty State: docs restructure and content refinement

jim-looking-around

## :hammer_and_wrench: Type Of Change

- [x] Documentation
- [ ] Build system
- [ ] CI
- [ ] Style (code style changes, not css changes)
- [ ] Test
- [ ] Other (chore)

## :book: Description

Edited docs content — grammar, concise, tight — and organized it for improved visual flow and hierarchy.

## :pencil: Checklist

For all PRs:

- [x] I have ensured no private Dialpad links or info are in the code or pull request description (Dialtone is a public repo!).
- [x] I have reviewed my changes.
- [x] I have added all relevant documentation.
- [x] I have considered the performance impact of my change.

PR #348, opened 2024-06-04
fix: NO-JIRA empty state update

iropolo:
Following @francisrupert latest change

https://github.com/dialpad/dialtone/pull/347/files#diff-5ac682276f1dad13ee004246335598d7378d7838261efe17e225e5a950bf5a69

Im updating `empty_state_constants.js` to match this change

Vue3 looks good.


francisrupert (approved):
ooh, good catch. thx!

PR #351, opened 2024-06-05
feat(css-utility): DLT-619 add font-variant-numeric utility

ninamarina:
# Add font-variant-numeric utility

Obligatory GIF

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket
[DLT-619]

## :book: Description
Adds `font-variant-numeric` utility classes. Adds doc page for this utility.

## :pencil: Checklist

For all PRs:

image

## :link: Sources
https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-numeric

[DLT-619]: https://dialpad.atlassian.net/browse/DLT-619?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/dialtone-documentation/docs/utilities/typography/font-variant-numeric.md
+++ b/apps/dialtone-documentation/docs/utilities/typography/font-variant-numeric.md
@@ -29,30 +29,6 @@ The class `d-fvn-ordinal` forces the use of special glyphs for ordinal markers, 
 <p class="d-fs-300 d-fvn-ordinal">...</p>
 ```
 
-## Slashed zero
-
-Use `d-fvn-slashed-zero` to add a slash to zeros. This is useful when a clear distinction between O and 0 is needed.
-
-<code-well-header class="d-d-flex d-jc-center d-fd-column d-p24 d-bgc-purple-100 d-w100p d-hmn102" custom>
-  <p class="d-fs-300 d-fc-purple-400 d-fvn-slashed-zero d-ff-mono">1234567890</p>
-</code-well-header>
-
-```html
-<p class="d-fs-300 d-ff-mono d-fvn-slashed-zero">...</p>
-```
-
-## Lining-nums
-
-The class `d-fvn-lining` control the figures used for numbers. Activates the set of figures where numbers are all lying on the baseline.
-
-<code-well-header class="d-d-flex d-jc-center d-fd-column d-p24 d-bgc-purple-100 d-w100p d-hmn102" custom>
-  <p class="d-fs-300 d-fc-purple-400 d-fvn-lining">1234567890</p>
-</code-well-header>
-
-```html
-<p class="d-fs-300 d-fvn-lining">...</p>
-```
-
 ## Proportional-nums
 
 Use `d-fvn-proportional` to set different sizes for each number.
@@ -137,18 +113,6 @@ Use `d-fvn-diagonal` to display numerator and denominator smaller and separated 
 
 ```html
 <p class="d-fs-300 d-fvn-diagonal">...</p>
-```
-
-## Stacked-fractions
-
-Use `d-fvn-stacked` to display numerator and denominator smaller and separated by a horizontal line.
-
-<code-well-header class="d-d-flex d-jc-center d-fd-column d-p24 d-bgc-purple-100 d-w100p d-hmn102" custom>
-  <p class="d-fs-300 d-fc-purple-400 d-fvn-stacked">1/2 3/4 5/6</p>
-</code-well-header>
-
-```html
-<p class="d-fs-300 d-fvn-stacked">...</p>
 ```
 
 ## Unset
--- a/apps/dialtone-documentation/docs/utilities/typography/font-variant-numeric.md
+++ b/apps/dialtone-documentation/docs/utilities/typography/font-variant-numeric.md
@@ -26,7 +26,7 @@ The class `d-fvn-ordinal` forces the use of special glyphs for ordinal markers, 
 </code-well-header>
 
 ```html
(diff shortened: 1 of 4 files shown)


francisrupert:
I know this is draft, but PR name should be `feat(css-utility): DLT-619 add font-variant-numeric utility`

ninamarina:
@francisrupert there are some of these classes that don't do anything when applied with the typographies we are using.
Slashed-zero: This glyph is not present in the fonts we are using, except for Mono, which only has slashed zeros
Lining-nums: doesn't seem to be doing any difference with the fonts we are using
Stacked-fractions: same here, they are not available.

What should I do? should we still support them in case in the future these glyphs are added or a new font is added? does it make sense to show the examples?

francisrupert:
> What should I do? should we still support them in case in the future these glyphs are added or a new font is added? does it make sense to show the examples?

For those that don't work for one or other, place an inline `<dt-notice kind="info">` noting which one is can currently be applied to. For example, Slashed zero only works for monospace.

For those that don't work for either, suppress it from the docs but retain in CSS. 

This isn't the biggest deal at the moment. Tabular-nums is the main one we're likely to use, so we're at least future-proofing for all possible `font-variant-numeric` values.

ninamarina:
> > What should I do? should we still support them in case in the future these glyphs are added or a new font is added? does it make sense to show the examples?
> 
> For those that don't work for one or other, place an inline `<dt-notice kind="info">` noting which one is can currently be applied to. For example, Slashed zero only works for monospace.
> 
> For those that don't work for either, suppress it from the docs but retain in CSS.
> 
> This isn't the biggest deal at the moment. Tabular-nums is the main one we're likely to use, so we're at least future-proofing for all possible `font-variant-numeric` values.

Slashed zero still doesn't make a difference for monospace, because it's doesn't allow a non slashed zero.
So I think that one should also be removed from the docs.

francisrupert:
> For those that don't work for one or other, place an inline `<dt-notice kind="info">` noting which one is can currently be applied to. For example, Slashed zero only works for monospac

Were there any that fit this situation? e.g. one works on monospace, but not default font-family?

apps/dialtone-documentation/docs/utilities/typography/font-variant-numeric.md
```diff
@@ -0,0 +1,154 @@
...
+  </div>
+</code-well-header>
+
+```html
+<div class="d-fs-300 d-fvn-proportional"><table>...</table></div>
+```
+
+## Tabular-nums
+
+Use `d-fvn-tabular` to set the same size for each number. Useful for easily align them in tables.
```
francisrupert: "Useful for aligning tabular data and preventing layout shifts of dynamically-changing content. For example, a timer counting down."

apps/dialtone-documentation/docs/utilities/typography/font-variant-numeric.md
```diff
@@ -0,0 +1,154 @@
...
+## Ordinal
+
+The class `d-fvn-ordinal` forces the use of special glyphs for ordinal markers, like 1st, 2nd, 3rd.
+
+<code-well-header class="d-d-flex d-jc-center d-fd-column d-p24 d-bgc-purple-100 d-w100p d-hmn102" custom>
+  <p class="d-fs-300 d-fc-purple-400 d-fvn-ordinal">1st, 2nd, 3rd, 4th</p>
+</code-well-header>
+
+```html
+<p class="d-fs-300 d-fvn-ordinal">...</p>
```
braddialpad: for this one we may want to show the actual content in the example rather than ...

I was kinda confused whether it would do this when you just put the number "1" or if you had to actually input the full text "1st"

apps/dialtone-documentation/docs/utilities/typography/font-variant-numeric.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+## Proportional-nums
+
+Use `d-fvn-proportional` to set different sizes for each number.
```
braddialpad, line 34: Is this the default in a table?

apps/dialtone-documentation/docs/utilities/typography/font-variant-numeric.md
```diff
@@ -0,0 +1,154 @@
...
+  <p class="d-fs-300 d-fc-purple-400 d-fvn-diagonal">1/2 3/4 5/6</p>
+</code-well-header>
+
+```html
+<p class="d-fs-300 d-fvn-diagonal">...</p>
+```
+
+## Unset
+
+Use `d-fvn-unset` to deactivate other value previusly set to `font-variant-numeric`.
```
braddialpad: typo: "previously"

ninamarina:
> > For those that don't work for one or other, place an inline `<dt-notice kind="info">` noting which one is can currently be applied to. For example, Slashed zero only works for monospac
> 
> Were there any that fit this situation? e.g. one works on monospace, but not default font-family?

No, the slashed zeros are the only option for Monospace, so setting the class would not do anything

apps/dialtone-documentation/docs/utilities/typography/font-variant-numeric.md
ninamarina: No, the default is `normal`. For some fonts, you might see a change between the two.

francisrupert (approved):
Thanks!

braddialpad (approved):

PR #356, opened 2024-06-07
fix(utility): NO-JIRA border color opacity utility

ninamarina:
# fix: NO-JIRA border color opacity utility

The utility `d-bco{n}` was at some point generated as `d-bco-{n}`. We think this was unintended because the docs show `d-bco{n}`, and the rest of the utilities for opacity follow that same pattern.

| Before | Now | 
|----|----|
| image | image |


francisrupert (approved):
All good. Change PR title to `fix(css-utility): NO-JIRA corrected border color opacity name`

braddialpad (approved):

PR #360, opened 2024-06-11
fix(input): NO-JIRA set autocomplete="off" on input

braddialpad:
# fix(input): set autocomplete="off" on input

Obligatory GIF

- [x] Fix

## :book: Description

set autocomplete="off" on dt-input

## :bulb: Context

Noticed that sometimes on the emoji picker the OS would autosuggest while typing in the field. This should disable this behaviour from happening.

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

Update in product

before:

Screenshot 2024-06-11 at 9 23 31 AM

after:

Screenshot 2024-06-11 at 9 24 18 AM

## :link: Sources

https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#disabling_autocompletion


iropolo:
hmm.. would it be better to do it in the emoji-picker intsel instead of globally?

Maybe there is some places where it is useful to have the autocomplete enabled.

braddialpad:
> hmm.. would it be better to do it in the emoji-picker intsel instead of globally?
> 
> Maybe there is some places where it is useful to have the autocomplete enabled.

nah, I think if we ever want autocomplete we want to handle it ourselves via combobox and such. I think we should disable it completely until we get some request from someone that it needs to be enabled.

iropolo (approved):
I would note this in the Dialtone update ticket to QA

PR #368, opened 2024-06-14
docs(prototypes): DLT-1740 left bar prototype

juliodialpad:
# Dialpad's Left bar prototype

Obligatory GIF

- [x] Documentation

## :book: Jira Ticket

[DLT-1740](https://dialpad.atlassian.net/browse/DLT-1740)

## :book: Description

Added Dialpad's leftbar prototype and layout for the future when we want to start adding the rest of the app prototypes.

## :bulb: Context

We need a prototype of how the app should be build with dialtone components, following the correct semantics and fully accessible through screen readers and keyboard navigation.

## :pencil: Checklist

For all PRs:

For all Vue changes:

Screenshot 2024-06-14 at 9 45 37 a m

## :link: Sources

https://www.figma.com/design/oTY56LgQha8ljb4c8w4GtO/Dialpad-App-Components?node-id=9240-7131&m=dev

[DLT-1740]: https://dialpad.atlassian.net/browse/DLT-1740?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dialtone-vue3/.storybook/preview.jsx
```diff
@@ -19,6 +20,7 @@ setCustomEmojiJson(customEmojiJson);
 setup((app) => {
   app.use(fixDefaultSlot)
   app.use(DtTooltipDirective);
+  faker.seed(6687422389464139);
```
juliodialpad, line 24: faker global seed, to make sure results are reproducible on percy and don't change on every reload too.

packages/dialtone-vue3/components/list_item/list_item.vue
```diff
@@ -196,6 +204,10 @@ export default {
           this.onMouseLeave(event);
           this.$emit('mouseleave', event);
         },
+
+        click: event => {
+          this.$emit('click', event);
+        },
```
juliodialpad: This was emitting a warning

packages/dialtone-vue3/prototypes/dialpad/leftbar/components/leftbar_section.vue
```diff
@@ -0,0 +1,64 @@
...
+          label-class="d-jc-flex-start d-w100p"
+          importance="clear"
+          kind="muted"
+          @click="defaultToggleOpen"
+        >
+          <dt-icon
+            :name=" isOpen ? 'chevron-down' : 'chevron-right'"
+            size="300"
+          />
+          <span class="d-truncate d-w128 d-ta-left">{{ title }}</span>
```
juliodialpad: Had to put this d-w128 to achieve truncate, we'll need to make this with JS otherwise if we need it to be dynamically updated as the slot can be anything.

packages/dialtone-vue3/prototypes/dialpad/dialpad.stories.js
```diff
@@ -0,0 +1,25 @@
+import { createTemplateFromVueFile } from '@/common/storybook_utils';
+import DialpadDefaultTemplate from './dialpad_default.story.vue';
+
+// Story Collection
+export default {
+  title: 'Prototypes/Dialpad',
```
ninamarina, line 6: Why call it Dialpad instead of Leftbar?

packages/dialtone-vue3/prototypes/dialpad/leftbar/components/leftbar_section.vue
iropolo: I would put this on a `<!-- -->` in the template, kind of a to do or just to let it know this what you are saying and not missing it in the PR.

packages/dialtone-vue3/prototypes/dialpad/leftbar/components/leftbar_section.vue
```diff
@@ -0,0 +1,64 @@
...
+      <slot name="items" />
+    </template>
+  </dt-collapsible>
+</template>
+
+<script setup>
+import { DtCollapsible } from '@/components/collapsible';
+import { DtButton } from '@/components/button';
+import { DtIcon } from '@/components/icon';
+import { defineProps, ref } from 'vue';
```
iropolo: `defineProps` not necessary to import.

packages/dialtone-vue3/prototypes/dialpad/leftbar/components/options_dropdown.vue
```diff
@@ -0,0 +1,73 @@
...
+    </template>
+  </dt-dropdown>
+</template>
+
+<script setup>
+import { DtDropdown, DtDropdownSeparator } from '@/components/dropdown';
+import { DtListItem } from '@/components/list_item';
+import { DtButton } from '@/components/button';
+import { DtIcon } from '@/components/icon';
+import { ref, defineEmits } from 'vue';
```
iropolo: `defineEmits` not necessary to import.

packages/dialtone-vue3/prototypes/dialpad/leftbar/dialpad_leftbar.vue
```diff
@@ -0,0 +1,234 @@
...
+import OptionsDropdown from './components/options_dropdown.vue';
+import { DtButton } from '@/components/button';
+import { DtIcon } from '@/components/icon';
+import { DtStack } from '@/components/stack';
+import { DtRecipeGeneralRow } from '@/recipes/leftbar/general_row';
+import { DtRecipeContactRow } from '@/recipes/leftbar/contact_row';
+import { faker } from '@faker-js/faker';
+import { ref, computed } from 'vue';
+
+const mainOptions = ref([
```
iropolo: Since all this is static data, I guess these can go without `ref`

packages/dialtone-vue3/.storybook/preview.jsx
iropolo: I would add this comment on the top of the `faker.seed`

packages/dialtone-vue3/prototypes/dialpad/dialpad.stories.js
```diff
@@ -0,0 +1,25 @@
...
+};
+
+// Templates
+const DefaultTemplate = (args, { argTypes }) => createTemplateFromVueFile(
+  args,
+  argTypes,
+  DialpadDefaultTemplate,
+);
+
+export const Default = {
```
iropolo, line 19: Maybe we can rename this `Default` to `Leftbar` then this will be the structure:

image

Just as a suggestion, related to Nina's comment

packages/dialtone-vue3/prototypes/dialpad/leftbar/components/options_dropdown.vue
```diff
@@ -0,0 +1,73 @@
...
+  { value: 3, name: 'Browse channels' },
+]);
+
+const openDropdown = (value) => {
+  emits('open-dropdown', value);
+};
+
+const selectMenuOption = (option, close) => {
+  close?.();
+  emits('select-menu-option', option);
```
iropolo, line 70: Regarding events through the component,
We are emitting them but not showing it in the `Actions` tab of storybook, also we are not using it at all, Im guessing we can remove them or keep it track in the `Actions` tab.

iropolo: Since it is a prototype I guess we can keep it track in the `Actions` tab to let us know which events are being fired through the usage.

Not so sure, open to discussion, it's fine if we leave it as it is now.

iropolo:
Seems like DND or OOO is missing a transparency

Current:
image


Product:
image

iropolo:
Not sure if it is a behaviour that we should replicate because this is just a design perspective, ~~but maybe it is... not sure about the scope of the prototypes 🤔 .~~

~~If you have unread msg in some collapsed section, that user conversation/channel or group chat will remain visible.~~

update edit: not in prototype scope


Screenshot 2024-06-18 at 13 55 55

francisrupert (commented):
Pretty darn cool!

Each section header definitely too big, and expand/collapse chevron icon should align with other icons at its same vertical level.

### Prototype
https://github.com/dialpad/dialtone/assets/1165933/fec83e8f-adcb-4772-8b8a-6e070961491b

### Beta
https://github.com/dialpad/dialtone/assets/1165933/d72d4f23-d78b-4311-b368-eaccfd4d7609

packages/dialtone-vue3/prototypes/dialpad/dialpad.stories.js
braddialpad: I think because by the time it's done it will be a prototype of the entire core dialpad app.

packages/dialtone-vue3/components/list_item/list_item.vue
braddialpad: What was the warning? There might be a problem with doing this because a list item can be non-interactive in which case we don't want to emit a click event.

packages/dialtone-vue3/prototypes/dialpad/leftbar/components/leftbar_section.vue
braddialpad: does max-width work better? or it needs to be fixed?

packages/dialtone-vue3/prototypes/dialpad/leftbar/components/options_dropdown.vue
braddialpad: Hmm probably okay to not emit them in a prototype in an internal component like this.

packages/dialtone-vue3/prototypes/dialpad/leftbar/dialpad_leftbar.vue
braddialpad, line 1: Nice, this is so clean looking compared to the current implementation in DP.

packages/dialtone-vue3/components/list_item/list_item.vue
juliodialpad: The error was that it was emitting a click event but click wasn't declared on 'emits'

packages/dialtone-vue3/prototypes/dialpad/dialpad.stories.js
juliodialpad: Yeah, exactly, it's not just the leftbar, it's the entire dialpad prototype

packages/dialtone-vue3/prototypes/dialpad/dialpad.stories.js
juliodialpad: The prototype will be the entire dialpad app, not only the leftbar, right now it only includes the leftbar but will include everything else in the near future.

packages/dialtone-vue3/prototypes/dialpad/leftbar/components/leftbar_section.vue
juliodialpad: Yes! max-width worked better indeed. 

I would like a design review here, is this ok to do? or should we set a fixed percent of it for the channel name and the rest for the button?

Something like 60/40 instead of a fixed 12.8 rem? I think the problem is that on product the leftbar is resizable.

Screenshot 2024-06-20 at 12 07 21 p m
Screenshot 2024-06-20 at 12 07 42 p m

@francisrupert @fede-dp

packages/dialtone-vue3/prototypes/dialpad/leftbar/components/options_dropdown.vue
juliodialpad: Yes, that's what I was thinking, we don't really need to emit the events in this component. I'll left the component as is right now.

packages/dialtone-vue3/components/list_item/list_item.vue
juliodialpad: https://github.com/dialpad/dialtone/assets/87546543/a1b4d88c-a8dd-490b-b55e-478e748e3323

Only firing the event while pressing `enter` or `space` on a list-item

iropolo (approved):
Looks good to me!,
Lets wait for Design approval. 

Vue2 version would be in a separeted PR?, thanks

iropolo:
Maybe there is a slightly improvement for the structure, since it will hold all the Dialpad structure and not just the Leftbar.
Could we create a `section` folder, move in there `Leftbar` folder.
There is also a component that will be probably shared trough others sections like `options_dropdown.vue`, create a components folder at the same level of `section`

Will be sth like:

```
prototypes/
├─ dialpad/
│  ├─ sections/
│  │  ├─ leftbar/
│  │  ├─ another_section/
│  ├─ components/
│  │  ├─ options_dropdown.vue/

```

I have some concerns about the static data generated by `faked`, for now seems to work softly but is there a reason that we want to have some dynamic data in the prototype? 
I was thinking maybe having a separeted folder to hold all this static data.

This will probably mutate in the near future.
Not necessary to do anything of this now, we can wait to see how it grows.

fede-dp:
looks good to me, at least visually! let's wait for Francis to check the semantics

packages/dialtone-vue3/prototypes/dialpad/leftbar/components/options_dropdown.vue
```diff
@@ -0,0 +1,73 @@
...
+import { ref } from 'vue';
+
+const emits = defineEmits(['select-menu-option', 'open-dropdown']);
+
+const menuOptions = ref([
+  { value: 0, name: 'Mark all as read' },
+  { value: 1, separator: true },
+  { value: 2, name: 'Create new channel' },
+  { value: 3, name: 'Browse channels' },
+]);
```
iropolo: This probably not need to be a `ref`, seems like it is static data

juliodialpad:
> Maybe there is a slightly improvement for the structure, since it will hold all the Dialpad structure and not just the Leftbar. Could we create a `section` folder, move in there `Leftbar` folder. There is also a component that will be probably shared trough others sections like `options_dropdown.vue`, create a components folder at the same level of `section`
> 
> Will be sth like:
> 
> ```
> prototypes/
> ├─ dialpad/
> │  ├─ sections/
> │  │  ├─ leftbar/
> │  │  ├─ another_section/
> │  ├─ components/
> │  │  ├─ options_dropdown.vue/
> ```
Sounds like a good improvement, but I'd prefer to keep it as it is right now and mutate when needed in the future, maybe the requirements change, so let's not double the work now and later.

> I have some concerns about the static data generated by `faked`, for now seems to work softly but is there a reason that we want to have some dynamic data in the prototype? I was thinking maybe having a separated folder to hold all this static data.

We want dynamic data on the prototype to check how it behaves on different cases, with long and short texts, we might also want some status with emojis or something like that in the future, all of that can be achieved with `faker`.

You mean to have a separate folder with some JSON files with this data? I feel like using faker in this way is more natural and automated, which is the advantage of doing the way you're thinking?

iropolo:
> You mean to have a separate folder with some JSON files with this data? I feel like using faker in this way is more natural and automated, which is the advantage of doing the way you're thinking?

All static data should be separated from component logic, since it is not required to be there.
Based on that, there should be room for improvement.

For now it can still working as this way, we just have to pay attention to scalability and to have a readable structure.

PR #371, opened 2024-06-14
fix(message-input): NO-JIRA small issues with slash commands

bianca-artola-dialpad:
# fix(message-input): NO-JIRA small issues with slash commands

Obligatory GIF

- [x] Fix

## :book: Description

While i was integrating pills on new message input, I come up with two small issues.

1. It seems that lastValue prop was not being updated correctly. I replaced that to use value isntead; not sure if that going to work on product side tbh, but nothing else came to my mind. Let me know what you think. Seems to be still working on dialtone side.
image

2. `meeting-pill-close` was not being triggered. I added it and know should work.

## :bulb: Context
Here's the work I had already made on product side for this: https://github.com/dialpad/firespotter/pull/43603

### DIFF
--- a/packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
+++ b/packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
@@ -319,7 +319,6 @@ export default {
   data () {
     return {
       editor: null,
-      lastValue: this.value,
     };
   },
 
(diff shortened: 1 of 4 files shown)


packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
```diff
@@ -535,13 +534,12 @@ export default {
       // The content has changed.
       this.editor.on('update', () => {
         const value = this.getOutput();
-        if (this.preventTyping && value.length > this.lastValue.length) {
-          this.editor.commands.setContent(this.lastValue, false);
+        if (this.preventTyping && value.length > this.value.length) {
+          this.editor.commands.setContent(this.value, false);
```
braddialpad, line 538: hmm yeah I guess this will work since this.value hasn't been updated yet.

braddialpad (approved):

PR #380, opened 2024-06-24
docs: DLT-1791 move storybook to doc site -  Avatar to ButtonGroup

ninamarina:
# Move storybook to doc site - Avatar to ButtonGroup

Obligatory GIF

- [x] Documentation

## :book: Jira Ticket
[DLT-1791]

## :book: Description
Moves storybook docs to docs site for:
- Avatar
- Badge
- Banner
- Breadcrumbs
- Button
- Button Group

Most of them are unchanged since the doc site was already more complete than Storybook. The docs where removed from Storybook and instead a link to the doc site is shown.

Updates CodeExampleTabs component, to allow to automatically generate the HTML tab using a reference to the rendered component. This will make the task of migrating easier to avoid having to copy paste from the inspector. Adds the `pretty` dependency to format the generated HTML. Also, set the default for `showHtmlWarning` to true, since we want to show it in every component.

## :pencil: Checklist

For all PRs:

image

## :link: Sources
https://prettier.io/docs/en/browser

[DLT-1791]: https://dialpad.atlassian.net/browse/DLT-1791?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/dialtone-documentation/docs/components/avatar.md
```diff
@@ -112,10 +121,12 @@ vueCode='
   color="100"
 />
 '
-showHtmlWarning />
+/>
 
 ### Image
 
+If `image-src` is not provided, or if image fails to load, the avatar will fall back to the initials extracted from the `full-name`. By default, the `alt` attribute of the image will be the `full-name`, this can be overridden via `image-alt` prop.
```
juliodialpad: That statement is no longer true, I forgot to update that docs a while ago.

```suggestion
If `image-src` is not provided, or if image fails to load, the avatar will fall back to the initials extracted from the `full-name`.
```

juliodialpad (approved):
Looking good from engineering perspective, lets check with Design if they have some comments before merging.

apps/dialtone-documentation/docs/components/banner.md
```diff
@@ -74,7 +76,36 @@ vueCode='
...
+
+### Pinned
+
+Pins the banner to the top of the window.
+
+<code-well-header>
+    <div class="d-d-flex d-w100p d-flow8 d-ai-flex-end">
+        <dt-button @click="toggleBanner">Toggle Example</dt-button>
+    </div>
+</code-well-header>
```
iropolo: Should be 2 space indentation

francisrupert (approved):
Nice and clean. Definitely approve for the doc syncing and removal from Storybook. 

The toggling variant functionality no longer works. 



https://github.com/dialpad/dialtone/assets/1165933/1d94d29a-f6ba-4a4c-985e-5943758a4fc1

apps/dialtone-documentation/docs/.vuepress/baseComponents/CodeExampleTabs.vue
```diff
@@ -85,6 +126,33 @@ const htmlTabId = getUniqueString();
...
+/**
+ * Transforms a single-line HTML string to an indented multiline HTML string.
+ * Removes comments, id and data-qa attributes, and simplifies svg tags.
+ * Also, adds a new line before each svg, img, and span tag because pretty
+ * doesn't do it.
+ * @param elementHTML - The HTML code to be formatted.
+ * @returns The formatted HTML code.
+ */
+const formatHTML = (elementHTML) => {
+  const normalizedHTML = elementHTML
```
iropolo: method should go after `onMounted`

apps/dialtone-documentation/docs/.vuepress/baseComponents/CodeExampleTabs.vue
```diff
@@ -57,26 +57,67 @@
...
+  /**
+   * Retrieves the reference to the example component to get the HTML code.
+   * If not provided, the HTML code will be used.
+   *
+   * @returns {Component} The reference to the component.
+   */
+  getComponentRef: {
+    type: Function,
+    default: null,
+  },
```
iropolo: I like this addition.
But now we have two ways to do sth internal.

On our `code-example-tabs`, we could use
`getComponentRef` that you are using passing `ref=` on the `code-well-header` section.
or
`htmlCode` 

As I see there are some pro and cons on this.

cons
- since this is internal development, I dont see why we should have two different ways to achieve this.
kind of confusing when you should use one or another, I will probably use on those components that generates large html code but in these cases we just add a `...` and thats it avoid repited code, like datepicker generates a bunch of similar code.
- we dont have the ability to modified sth on the generated html code, now we are using regex. This can be less flexible in some particular case.
- Seems like it will work only on single components, so if our example has multiple we cannot use it.
Maybe we can improve it with a wrapper, dunno.
- It is necessary the addition of the pretty dependency
- In some cases seems like it is not doing the right job 

image


pros
- Much less code and less style error prone (indents, close html tags,etc..)
- Probably we can use it if we improve it, in most of the cases, if this is the case I would agree on have it both ways, for example if we could use it in the 90% of the cases. So `htmCode` will be used in particular cases.


What I would suggest, move this to a different PR, with a different ticket as a proposal to improvement and we can continue this discussion and implementation.

I definitely like this approach so if we can tackle most of the cons will be great to move it forward with it.

iropolo (approved):
Apart from my previous comment, that adds an improvement to how we build the code examples, seems is out of the idea of the ticket.

This `Move storybook doc to docsite` jira ticket scope is correct in this PR changes.

apps/dialtone-documentation/docs/.vuepress/baseComponents/CodeExampleTabs.vue
iropolo: If I'm missing sth on the idea of implement it, let me know, maybe Im skipping some advantage of it.

ninamarina: The new prop `getComponentRef` allows for an "autogenerated" html tab. It has a lot of advantages like it's honest, and it will make it easier to migrate everything from the following tickets in this epic.
The advantage of having both ways of generating the html code is that we don't need to migrate everything we already have there.
It works if there are many components in the example if you add the `ref` to the wrapper.
I don't see any problems really in adding this in this PR. It is a third party dependency needed for formatting, but it's only a dev dependency and I haven't found any problems with it.

braddialpad: > In some cases seems like it is not doing the right job

Would prettier work better? It's considered the standard for opinionated code formatting. Might have a lot more options we can work with.
https://www.npmjs.com/package/prettier

apps/dialtone-documentation/docs/.vuepress/baseComponents/CodeExampleTabs.vue
```diff
@@ -57,26 +57,67 @@
...
 import { getUniqueString } from '@workspaceRoot/common/utils';
 
 const props = defineProps({
+  /**
+   * Retrieves the reference to the example component to get the HTML code.
+   * If not provided, the HTML code will be used.
+   *
+   * @returns {Component} The reference to the component.
+   */
+  getComponentRef: {
```
braddialpad: Instead of creating this new prop within this component, can we just leave this component with the htmlCode prop as it was before and create a utility function (or composable maybe?) that could convert a vue component to a string of formatted code, and then pass that string into the vueCode prop? I think this would solve some of the separation concerns @iropolo had mentioned.

packages/dialtone-vue2/components/avatar/avatar.mdx
```diff
@@ -1,110 +1,10 @@
...
+import { Subtitle, Meta } from '@storybook/blocks';
 import * as AvatarStories from './avatar.stories.js';
 
 <Meta of={AvatarStories}/>
 
 # Avatar
 
 <Subtitle>
-  An avatar is a visual representation of a user or object.
+  Check the documentation [here](https://dialtone.dialpad.com/components/avatar.html).
```
braddialpad: Maybe a bit more info about why we did this since this is a point of confusion for many users.

"Please see the component documentation here: [LINK]

Component documentation has been moved to our main documentation site and storybook remains to be used as a component playground only."

braddialpad (changes requested):

fede-dp (approved):
visually looks great!

apps/dialtone-documentation/docs/.vuepress/baseComponents/CodeExampleTabs.vue
ninamarina: Updated so that htmlCode prop can be a string or a function to get the ref.

apps/dialtone-documentation/docs/.vuepress/baseComponents/CodeExampleTabs.vue
ninamarina: I changed it to use prettier, although it has the same caveats as pretty, but it is more standardised and also might be useful for code formatting in the future.

braddialpad (approved):
Excellent thanks!

PR #389, opened 2024-06-28
fix(emoji-picker): DP-102475 fix keyboard navigation

ninamarina:
# Fix Emoji Picker keyboard navigation when there is no search input

## Obligatory GIF (super important!)

Obligatory GIF

## :hammer_and_wrench: Type Of Change

These types will increment the version number on release:

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

These types will not increment the version number, but will still deploy to documentation site on release:

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Style (code style changes, not css changes)
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket
[DP-102475]

## :book: Description
When the prop `showSearch` is false, hitting tab from the emoji tab set would fail with the error `Cannot read properties of undefined (reading 'focusSearchInput')` because the ref to the search input was not there.

To solve this, I changed the emitted event from `'focus-search-input'` to `'tab-key-pressed'` so that the parent component can handle the event and decide to which section should the focus move.

## :pencil: Checklist

For all PRs:

- [x] I have ensured no private Dialpad links or info are in the code or pull request description (Dialtone is a public repo!).
- [x] I have reviewed my changes.
- [x] I have added all relevant documentation.
- [x] I have considered the performance impact of my change.

For all Vue changes:

- [x] I have added / updated unit tests.
- [ ] I have made my changes in Vue 2 and Vue 3. Note: you may sync your changes from Vue 2 to Vue 3 (or vice versa) using the `./scripts/dialtone-vue-sync.sh` script. Read docs here: [Dialtone Vue Sync Script](../packages/dialtone-vue3/.github/CONTRIBUTING.md#dialtone-vue-sync-script)
- [x] I have validated components with a screen reader.
- [x] I have validated components keyboard navigation.

## :camera: Screenshots / GIFs
Remember that to reproduce this you must set `showSearch` to false.

### Before
emoji-picker-error

### Now

emoji-picker-fixed

## Next Steps
Add changes to Vue 3 once approved

[DP-102475]: https://dialpad.atlassian.net/browse/DP-102475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #393, opened 2024-07-01
fix(emoji-picker): NO-JIRA fix keyboard proposal

iropolo:
# Emoji picker fix keyboard - proposal

I late in my review to https://github.com/dialpad/dialtone/pull/389 .
I was doing it but then I freezed it, then the PR was already merge.

This PR just revert the previous PR changes (tests are ok) and just add a ternary to the events.

`@focus-search-input="showSearch ? $refs.searchInputRef.focusSearchInput() : $refs.emojiSelectorRef.focusEmojiSelector()"`

`@focus-search-input="showSearch ? $refs.searchInputRef.focusSearchInput() : $refs.tabsetRef.focusTabset()"`

Im not fan of logic in template but tbh I see this more clear, less code, less complex and it does not modify any component section.

---------

Some explanation:

I have some concerns of the addition of this methods and the names.

On 
```
<emoji-tabset
 @tab-key-pressed="focusNextSectionFromEmojiTabSet"
@shift-tab-key-pressed="$refs.skinSelectorRef.focusSkinSelector()"
@arrow-down-key-pressed="focusNextSectionFromEmojiTabSet"
```
These event names `tab` `shift-tab` `arrow-down` are being pressed in others sections as well so it sounds off that they does not have it.
It also repeat event with `tab-key-pressed` and `arrow-down-key-pressed`

This addition can be a little bit confusing, since in our code, events are attached to the structure,
`focus-search-input` `focus-skin-selector` `focus-tabset` `focus-emoji-selector`.

On 
```
<emoji-selector
@shift-tab-key-pressed="focusNextSectionFromEmojiSelector"
```
This has the `shift-tab` but not the others events (I would expect them).
It also says focus**Next**SectionFromEmojiSelector. But it goes to the previous one. Next should be `skin-selector`.

I would prefer to keep the previous code and just go with a ternary on the events.

### DIFF
--- a/packages/dialtone-vue2/components/emoji_picker/emoji_picker.vue
+++ b/packages/dialtone-vue2/components/emoji_picker/emoji_picker.vue
@@ -10,9 +10,8 @@
         :scroll-into-tab="scrollIntoTab"
         :tab-set-labels="tabSetLabels"
         :is-scrolling="isScrolling"
-        @tab-key-pressed="focusNextSectionFromEmojiTabSet"
-        @shift-tab-key-pressed="$refs.skinSelectorRef.focusSkinSelector()"
-        @arrow-down-key-pressed="focusNextSectionFromEmojiTabSet"
+        @focus-skin-selector="$refs.skinSelectorRef.focusSkinSelector()"
+        @focus-search-input="showSearch ? $refs.searchInputRef.focusSearchInput() : $refs.emojiSelectorRef.focusEmojiSelector()"
         @selected-tabset="scrollToSelectedTabset"
         @keydown.esc.native="$emit('close')"
       />
@@ -43,7 +42,7 @@
         @highlighted-emoji="updateHighlightedEmoji"
         @selected-emoji="$emit('selected-emoji', $event)"
         @focus-skin-selector="$refs.skinSelectorRef.focusSkinSelector()"
-        @shift-tab-key-pressed="focusNextSectionFromEmojiSelector"
+        @focus-search-input="showSearch ? $refs.searchInputRef.focusSearchInput() : $refs.tabsetRef.focusTabset()"
         @keydown.esc.native="$emit('close')"
       />
     </div>
@@ -233,22 +232,6 @@ export default {
 
     updateHighlightedEmoji (emoji) {
       this.highlightedEmoji = emoji;
-    },
-
-    focusNextSectionFromEmojiTabSet () {
-      if (this.showSearch) {
-        this.$refs.searchInputRef.focusSearchInput();
-      } else {
-        this.$refs.emojiSelectorRef.focusEmojiSelector();
-      }
-    },
-
-    focusNextSectionFromEmojiSelector () {
-      if (this.showSearch) {
-        this.$refs.searchInputRef.focusSearchInput();
-      } else {
-        this.$refs.tabsetRef.focusTabset();
-      }
     },
   },
 };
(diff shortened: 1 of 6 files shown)


iropolo:
This is just a proposal that I lost the opportunity to do it in the previous PR

packages/dialtone-vue2/components/emoji_picker/emoji_picker.vue
```diff
@@ -10,9 +10,8 @@
         :scroll-into-tab="scrollIntoTab"
         :tab-set-labels="tabSetLabels"
         :is-scrolling="isScrolling"
-        @tab-key-pressed="focusNextSectionFromEmojiTabSet"
-        @shift-tab-key-pressed="$refs.skinSelectorRef.focusSkinSelector()"
-        @arrow-down-key-pressed="focusNextSectionFromEmojiTabSet"
+        @focus-skin-selector="$refs.skinSelectorRef.focusSkinSelector()"
+        @focus-search-input="showSearch ? $refs.searchInputRef.focusSearchInput() : $refs.emojiSelectorRef.focusEmojiSelector()"
```
ninamarina, line 14: I don't think it makes it clearer to have a ternary in the template really, while a named function could be better. Maybe there's a better name than "focusNextSectionFromEmojiTabSet".
Also emitting `focus-search-input` is weird if we are not really going to focus the search input. I think the emoji selector shouldn't know if there is a search input or not. It's responsibility is to emit that the tab key was pressed. That is not how this was handled before that's why is inconsistent, but I think that approach is prone to errors (like the bug we just solved)

iropolo: Hiding search input is an optional case but the structure still has it. 
You are just... hiding it.

It is okay if you go with a structure event emits or an abstract event handler. But we should **_not mix them_**.

At this point the component responds to a structured keyboard navigation in order by

1. Tab
2. SearchInput
3. EmojiSelector
4. SkinSelector

Keyboard navigation is being handled by the components itself in order of this structure.
Thats why you can move around the component **_and_** jump to the previous or next section.

Hiding a part of this structure or changing it, its something that the `emoji-picker` component is not prepared for.
We should extract this `**and** jump to the previous or next section.` logic to the parent component and create a proper structure to handle it.

I don't see the previous PR code take part of this, then I propose a simple patch for it.
Then later if this is needed and this becomes a common practice (hidding parts of emoji-picker) we can take care of it accordingly.

> I don't think it makes it clearer to have a ternary in the template really, while a named function could be better

I agree, we can create a method here, I tested something like have a `focusSearchInput` method that handles if it comes from `emoji-selector` or `emoji-tabset` it should go to one or another if the `showSearch` is false.
But I see the ternary is more efficient. 

> Also emitting focus-search-input is weird if we are not really going to focus the search input.

It's what the sections are supposed to do, thats why we have `@focus-` events. 
What is weird for the component is to hide a section.

braddialpad: This external search definitely throws an annoying wrench into things.

It might make sense to emit keyboard events in more generic cases, however the action performed here has logic behind it specifically meaning "focus the input". If we were to emit and capture this event as a consumer it should still be used to "focus the input".

My question is, should we be setting focus at all when showSearch is false? Shouldn't we just emit the event and allow the consumer to focus the search input since it is external? 

I'm not sure we need to spend a ton of time to do this perfectly. We just need the ability to fix this one weird case where we have to use an external input. I doubt this event would be used in any case other than this one.

braddialpad (approved):

packages/dialtone-vue2/components/emoji_picker/emoji_picker.vue
iropolo: We can expose some of this event later if it is required, for now they are all internal just for the correct behaviour of the component.

I will move forward this change, if there is some issue that this fix didn't take we have https://github.com/dialpad/dialtone/pull/389 implementation.

PR #394, opened 2024-07-02
feat(multi-select): DLT-1845 allow duplicated names on multiselect

bianca-artola-dialpad:
# feat(DLT-1845): allow duplicated names on multiselect

Obligatory GIF

- [x] Feature

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DLT-1845

## :book: Description
We need to find a way to avoid console errors when we have duplicate values on the items. Since we are using the same strings that we rendered as keys (which are usually not ids or keys, they can be names) we can have duplicate values. What I have done to allow duplicate names is to convert the key to `{value}-{index}`.
Please feel free to suggest other ideas to achieve this behaviour.

## :bulb: Context
We use the multi-select component to select users for the bulk SMS feature. Sometimes we have more than one user with the same name and in these cases the multi-select component returned an error.

### DIFF
--- a/packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
+++ b/packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
@@ -23,9 +23,9 @@
           class="combobox__chip-wrapper"
         >
           <dt-chip
-            v-for="item in selectedItems"
+            v-for="({ item, key }) in selectedItemsWithKeys"
             ref="chips"
-            :key="getMultiSelectItemKey(item)"
+            :key="key"
             :label-class="['d-chip__label']"
             class="combobox__chip"
             :close-button-props="{ ariaLabel: 'close' }"
@@ -116,7 +116,7 @@ import {
   CHIP_TOP_POSITION,
 } from './combobox_multi_select_constants';
 import SrOnlyCloseButtonMixin from '@/common/mixins/sr_only_close_button';
-import utils from '@/common/utils';
+import { getUniqueString } from '@/common/utils';
 
 export default {
   name: 'DtRecipeComboboxMultiSelect',
@@ -384,6 +384,13 @@ export default {
         },
       };
     },
+
+    selectedItemsWithKeys () {
+      return this.selectedItems.map(selectedItem => ({
+        item: selectedItem,
+        key: getUniqueString(selectedItem),
+      }));
+    },
   },
 
   watch: {
@@ -605,10 +612,6 @@ export default {
       } else {
         this.showValidationMessages = false;
       }
-    },
-
-    getMultiSelectItemKey (item) {
-      return `${item}-${utils.getUniqueString()}`;
     },
   },
 };
--- a/packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
+++ b/packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
@@ -23,9 +23,9 @@
           class="combobox__chip-wrapper"
         >
           <dt-chip
-            v-for="item in selectedItems"
+            v-for="(item, index) in selectedItems"
             ref="chips"
-            :key="item"
+            :key="`${item}-${index}`"
             :label-class="['d-chip__label']"
             class="combobox__chip"
             :close-button-props="{ ariaLabel: 'close' }"
--- a/packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
+++ b/packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
@@ -23,9 +23,9 @@
           class="combobox__chip-wrapper"
         >
           <dt-chip
-            v-for="(item, index) in selectedItems"
+            v-for="item in selectedItems"
             ref="chips"
-            :key="`${item}-${index}`"
+            :key="getMultiSelectItemKey(item)"
             :label-class="['d-chip__label']"
             class="combobox__chip"
             :close-button-props="{ ariaLabel: 'close' }"
@@ -116,6 +116,7 @@ import {
   CHIP_TOP_POSITION,
 } from './combobox_multi_select_constants';
 import SrOnlyCloseButtonMixin from '@/common/mixins/sr_only_close_button';
+import utils from '@/common/utils';
 
 export default {
   name: 'DtRecipeComboboxMultiSelect',
@@ -604,6 +605,10 @@ export default {
       } else {
         this.showValidationMessages = false;
       }
+    },
+
+    getMultiSelectItemKey (item) {
+      return `${item}-${utils.getUniqueString()}`;
     },
   },
 };
--- a/packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.stories.js
+++ b/packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.stories.js
@@ -170,3 +170,12 @@ export const WithMaxSelectValidation = {
     },
   },
 };
+
+export const DuplicatedNames = {
+  render: (argsData) =>
+    createRenderConfig(DtRecipeComboboxMultiSelect, DtRecipeComboboxMultiSelectDefaultTemplate, argsData),
+
+  args: {
+    selectedItems: ['item12', 'item12', 'item12'],
+  },
+};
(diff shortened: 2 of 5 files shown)


ninamarina (approved):

packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
```diff
@@ -23,9 +23,9 @@
           class="combobox__chip-wrapper"
         >
           <dt-chip
-            v-for="item in selectedItems"
+            v-for="(item, index) in selectedItems"
             ref="chips"
-            :key="item"
+            :key="`${item}-${index}`"
```
braddialpad: Can we use getUniqueString function for this instead? We should avoid using index as a key in a v-for

https://vueschool.io/articles/vuejs-tutorials/tips-and-gotchas-for-using-key-with-v-for-in-vue-js-3/

braddialpad (changes requested):

packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
bianca-artola-dialpad: yes sure, let me update that

bianca-artola-dialpad: Hm my current implementation (easiest one) will not work as`getUniqueString` is being calculated more than one time per item and that is causing several issues (not focus properly, remove with keyboard not working, etc)
image
(thanks dev for writing tests, I wouldn't detect this issue without them).

so I was testing different options.. 
1) creating a new computed prop like this 
```
 selectedItemsWithKeys () {
      return this.selectedItems.map(item => ({
        value: item,
        uniqueKey: utils.getUniqueString(item),
      }));
    }, 
```
and then use that for the array, but seems like an overkill for this minor issue..
2) using index
3) allowing client to send an object array instead of string array and then using that key.

Can I ask for an opinion here? What do you think would be better? If you have a better idea please let me know

braddialpad: > 1. creating a new computed prop like this

I think this one will work best. It avoids a breaking change and makes it easiest for the consumer of the component to use.

packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
```diff
@@ -23,9 +23,9 @@
           class="combobox__chip-wrapper"
         >
           <dt-chip
-            v-for="item in selectedItems"
+            v-for="({ item, key }) in selectedItemsWithKeys"
             ref="chips"
-            :key="item"
+            :key="key"
```
iropolo, line 28:
```
v-for="item in selectedItems"
:key="getUniqueString(item)"
```

Im missing sth with this approach?

iropolo: Ohh.. I see.. test is complaining

bianca-artola-dialpad: yes 😞 seems to cause some problems when focusing the chips. Unfortunately I couldn't find a good way to fix the main problem and make the tests happy.
With the last solution I tried, everything seems to work fine, but the tests fail (although those cases seem to work fine). I'll have to investigate a bit more to see if the best solution is to just update the tests or provide another solution that makes the tests happy.

iropolo: Ok yes I supposed that test will complain because we will probably need to mock the `getUniqueString` function.
I did a couple of test and seems like this `UNIQUE_ID_COUNTER` of the function is causing troubles. 
If you replace it with something like return '${prefix}${1};'  test passes.

So we just need to add this at the top of the test to mock the function.

```
import * as utils from '@/common/utils';

vi.spyOn(utils, 'getUniqueString').mockImplementation(() => 'mockedUniqueString');
```

And it should works.

Reference: https://vitest.dev/guide/mocking.html#functions

edit: seems like another test broke, one sec.

iropolo:
Nice, seems test now works. Whether we go with computed property or inline Im ok, looks good with the fix using the `getUniqueString` function.

Add it for vue3 version and should be ok to go.

bianca-artola-dialpad:
thanks for helping @iropolo!!! quick question. The bug I noticed is only on vue2, do you want me to update vue3 anyway? I can do it of course, but let me know what you prefer (I already added a story on Vue3 to reproduce the use case)

iropolo:
> The bug I noticed is only on vue2, do you want me to update vue3 anyway? I can do it of course, but let me know what you prefer (I already added a story on Vue3 to reproduce the use case)

Hum... yes I suppose we should do the same, because it is an improvement also. Please add it in vue3.

bianca-artola-dialpad:
https://github.com/dialpad/dialtone/assets/144126346/09ce9688-1620-49be-9f63-011993d001ea

Updating this on Vue3 brings some problems with the focus (is always after first chip). I'm not sure if this is related to the input behavior on Vue3 or if there's something else that I'm not understanding. do you want me to create a follow-up ticket for this case?

iropolo:
> Updating this on Vue3 brings some problems with the focus (is always after first chip). I'm not sure if this is related to the input behavior on Vue3 or if there's something else that I'm not understanding. do you want me to create a follow-up ticket for this case?

Im Looking...

iropolo:
Okay, yes definetly looks weird.
~~Seems like it lost the node reference.~~ edit: not necessary relevant, I updated it to retain it and still works wrong.

Then for some reason now it populates the chips in reverse, previously it was [first,second,third,fourth], now is [fourth,third,second,first], so if you update this methods:

```
getLastChipButton () {
      return this.$refs.chips && this.getChipButtons()[0];
    },

    getLastChip () {
      return this.$refs.chips && this.getChips()[0];
    },

    getFirstChip () {
      return this.$refs.chips && this.getChips()[this.getChips().length - 1];
    },
```

Now works well..

Im looking for the reason of it

iropolo:
Ohh.. you know what... it because the key is dynamic then in every render it changes, so it really mess it.
Because we are using the uniqueString in the v-for ...

previous:
image


with `uniqueString`:
image

Im looking for an alternative 🥴

braddialpad:
> Im looking for an alternative 🥴

Creating computed prop like mentioned above should fix it

```
selectedItemsWithKeys () {
  return this.selectedItems.map(item => ({
    value: item,
    uniqueKey: utils.getUniqueString(item),
  }));
}, 
```

iropolo:
> > Im looking for an alternative 🥴
> 
> Creating computed prop like mentioned above should fix it
> 

~~Maybe Im missing sth testing it but it is the same whether inline or computed, and I guess Bianca is using computed by the commit~~

Hum.. it is a computed property so when it re evaluates, it generates the unique strings.

image

So I guess the `ref="chips"` just goes crazy and it does not mantain the original  order.


Previous: (same order and all items)
Screenshot 2024-07-04 at 15 27 36

Current with `uniqueString`: (reverse order ~~and seems like this.$refs.chips is one step behind~~)
Screenshot 2024-07-04 at 15 25 23

Dont mind the the amount of chips, it has the 4 items.
image


Vue3 manage refs a little bit different than vue2 so Im not surprise to have an issue here about this.

bianca-artola-dialpad:
> Vue3 manage refs a little bit different than vue2 so Im not surprise to have an issue here about this.

yeap, it seems that in Vue2 the change works fine (although the key is updated when re-evaluating) but in Vue3 it causes those focus problems which, as you mentioned, seem to be related to refs.
I've been reading about the best way to handle keys on Vue3, but the [article](https://github.com/dialpad/dialtone/pull/394#discussion_r1662827308) Brad mentioned seems to have the most important info. to be honest, as the `selectedItems` are dynamic, I don't see a good option for creating dynamic keys on the dialtone side.
Can we make this change for Vue2 only? Because the other option I see is to ask the consumer to send the keys, but as this is already used on Firespotter it can cause some breaking changes

iropolo:
> > Vue3 manage refs a little bit different than vue2 so Im not surprise to have an issue here about this.
> 
> yeap, it seems that in Vue2 the change works fine (although the key is updated when re-evaluating) but in Vue3 it causes those focus problems which, as you mentioned, seem to be related to refs. I've been reading about the best way to handle keys on Vue3, but the [article](https://github.com/dialpad/dialtone/pull/394#discussion_r1662827308) Brad mentioned seems to have the most important info. to be honest, as the `selectedItems` are dynamic, I don't see a good option for creating dynamic keys on the dialtone side. Can we make this change for Vue2 only? Because the other option I see is to ask the consumer to send the keys, but as this is already used on Firespotter it can cause some breaking changes

Yes... if @braddialpad  agree, I would not spend more time on this and just deliver it for vue2, as it is working in vue3 for now.

iropolo (approved):

iropolo:
Brad is OOO today, if it is needed to merge go ahead. This should not be a blocker.

bianca-artola-dialpad:
@iropolo this can wait if you prefer to wait for Brad's opinion, np! Not urgent at all
anyway I cannot merge without Brad's approval
image

braddialpad (approved):
I'm not exactly sure why this is different in vue 2 / vue 3, it is obviously something internal to vue, but if we are not having the problem in vue 3 I think it's okay to just merge this fix.

PR #397, opened 2024-07-05
docs: DLT-1787 implement custom TOC

juliodialpad:
# Dialtone documentation site - Collapsible TOC

Obligatory GIF

- [x] Documentation

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1787

## :book: Description

- Created a custom TOC.
- Removed `@vuepress/plugin-toc` dependency and usages.
- General clean up.
- Replaced `TokensToc` to use the same TOC component than everywhere, we shouldn't have two TOC components.

## :pencil: Checklist

For all PRs:

For all CSS changes:

https://github.com/dialpad/dialtone/assets/87546543/feb96153-6992-4ab3-8ddd-56f7d5a32309


apps/dialtone-documentation/docs/.vuepress/theme/components/PageToc.vue
```diff
@@ -1,29 +1,103 @@
...
 </script>
 
-<style>
+<style lang="less" scoped>
 .dialtone-toc {
   width: var(--dt-size-850);
-  height: calc(100vh - var(--dt-size-700));
+    height: calc(100vh - var(--dt-size-700));
+    top: var(--dt-space-700);
+    right: 3%; // Magic number?
```
ninamarina: Top and right should not be needed for the Page TOC if it's part of the grid, I only needed to add those for the tokens page, since the TOC was mounted from the tokens page, rather than from `Layout.vue`. 
And the magic numbers where the ones that I found that worked for best interaction with the token table.

apps/dialtone-documentation/docs/.vuepress/theme/components/PageToc.vue
```diff
@@ -1,29 +1,103 @@
...
+    z-index: var(--zi-base1);  /* to be on top of the page default toc */
```
ninamarina: Same with this z-index, should only be needed for the tokens page TOC

ninamarina (changes requested):
This was not clear in the ticket, but the collapsible feature is needed only on the token page because there are so many categories, I don't think we want this for the other page's TOC.
Other than this and the position that should be inside the grid for the rest of the pages, I see it correctly and I like that you unified the TOCs.

apps/dialtone-documentation/docs/.vuepress/theme/components/PageToc.vue
braddialpad: Do we need a prop to pass into this component to set these?

juliodialpad: I think it behaves correctly on both cases, can you see any issue by leaving it that way?

juliodialpad: I think this is a good addition to the components/utilities page too, looks cleaner and draws the user attention when clicking on a category.

Would be great to have some comments from @francisrupert @fede-dp here.

ninamarina: For large monitors at least, the TOC looks really far away in pages other than the tokens page:

image

What we have now, the TOC is on the grid:

image

Also, with a single glance I can see what contents are on the page.

francisrupert: > I think this is a good addition to the components/utilities page too, looks cleaner and draws the user attention when clicking on a category.

Actually, I like it! Nice happy accident simplifying the TOC, and it even expands/collapses as you scroll.

And yeah, TOC positioning should be consistent among all pages. I didn't realize https://dialtone.dialpad.com/tokens/ is currently the only page that doesn't retain it within the grid. Take a stab at keeping keeping TOC position (does it even need positioning?). If it takes too long to resolve consistently, we can punt on that as a new task to figure out in the future.

francisrupert:
@juliodialpad,

Thanks for getting the TOC positioning consistent!

I've come to really like that this works across all pages, not just the Tokens page. Is there a debounce adjustment that can be made so the TOC reflects the current view more quickly? 

View the attached video. The TOC expand/collapse and showing active item seems to only update once scrolling is complete.

https://github.com/dialpad/dialtone/assets/1165933/3f74897b-4a05-4d7b-a566-1e01570e76a3

juliodialpad:
> @juliodialpad,
> 
> Thanks for getting the TOC positioning consistent!
> 
> I've come to really like that this works across all pages, not just the Tokens page. Is there a debounce adjustment that can be made so the TOC reflects the current view more quickly?
> 
> View the attached video. The TOC expand/collapse and showing active item seems to only update once scrolling is complete.
> 
>  Screen.Recording.2024-07-09.at.3.15.09.PM.mov

We're using a VuePress plugin to automatically update the route based on the position (check the URL) but looks like it updates the position after a fixed amount of time (for performance reasons). 

Here's their code and that value cannot be changed by us: https://github.com/vuejs/vuepress/blob/9fb4bb00925d9409e6732118e996411c6a82c85d/packages/%40vuepress/plugin-active-header-links/clientRootMixin.js

We could create our own implementation of it, but definitely out of scope for this PR.

francisrupert:
> We could create our own implementation of it, but definitely out of scope for this PR.

Yeah cool. No biggie. We'll keep on eye on it and maybe consider doing that if we think it feels wrong over time or people complain enough or causes confusion. I approve from design perspective for now. Thanks!!

francisrupert (approved):

braddialpad (approved):

PR #392, opened 2024-07-01
feat(input): DLT-1808 icon scoped slot

juliodialpad:
# Add icon scoped slot to input

## Obligatory GIF (super important!)

Obligatory GIF

## :hammer_and_wrench: Type Of Change

These types will increment the version number on release:

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1808

## :book: Description

- Added iconSize scope to icon slot to be able to keep the icon size in sync with the input.
- General CSS cleanup of input.less

## :bulb: Context

- A while ago when we changed the way we were handling icons, forgot to update input accordingly, so the icon sizes where not being handled correctly.

## :pencil: Checklist

For all PRs:

- [x] I have ensured no private Dialpad links or info are in the code or pull request description (Dialtone is a public repo!).
- [x] I have reviewed my changes.
- [x] I have added all relevant documentation.
- [x] I have considered the performance impact of my change.

For all Vue changes:

- [x] I have made my changes in Vue 2 and Vue 3. Note: you may sync your changes from Vue 2 to Vue 3 (or vice versa) using the `./scripts/dialtone-vue-sync.sh` script. Read docs here: [Dialtone Vue Sync Script](../packages/dialtone-vue3/.github/CONTRIBUTING.md#dialtone-vue-sync-script)

For all CSS changes:

- [x] I have used design tokens whenever possible.

## :camera: Screenshots / GIFs

<table>
  <tr>
    <th>Size</th>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>XS</td>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
  <tr>
    <td>SM</td>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
  <tr>
    <td>MD</td>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
  <tr>
    <td>LG</td>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
  <tr>
    <td>XL</td>
    <td>
      
    </td>
    <td>
      
    </td>
  </tr>
</table>

## Links
- [Input component - Dialtone docs](https://dialtone.dialpad.com/deploy-previews/pr-392/components/input.html)
- [DtInput - Dialtone Vue 2](https://dialtone.dialpad.com/vue/deploy-previews/pr-392/?path=/story/components-input--with-both-icons)
[...]

PR #21, opened 2023-11-23
chore: remove utility classes from components

ninamarina:
Jira: https://dialpad.atlassian.net/browse/DLT-1125
Subtask: https://dialpad.atlassian.net/browse/DLT-1385
Dialtone vue: https://github.com/dialpad/dialtone-vue/pull/1314

## Description
Adds styles for the following components in dialtone library: 

Avatar
Chip
Codeblock
Checkbox

Collapsible
Combobox
Input
Input Group

Image Viewer
Keyboard Shortcut
List Item Group

For now I am going to leave this PR as is so it doesn't grow that much. The idea is to merge batches of components and release to progressively incorporate and test these changes in the Dialtone Vue PR.

## Pull Request Checklist

 - [x] Ask the contributors if a similar effort is already in process or has been solved.
 - [x] Review the [contribution guidelines](https://github.com/dialpad/dialtone/blob/staging/.github/CONTRIBUTING.md).
 - [x] Use `staging` as your pull request's base branch. (All PRs using `production` as its base branch will be declined).
 - [x] Ensure all `gulp` scripts successfully compile.
 - [ ] Update, remove, or extend all affected documentation.
 - [x] Ensure no private Dialpad links or info are in the code or pull request description (Dialtone is a public repo!).

PR #401, opened 2024-07-08
fix(input): NO-JIRA add identifier classes

braddialpad:
# fix(input): NO-JIRA add identifier classes

Obligatory GIF

- [x] Fix

## :book: Description

Added identifier classes in the input component. Removed unnecessary class definitions.

I will leave some comments for explanation of some things I did.

## :bulb: Context

Follow up to https://github.com/dialpad/dialtone/pull/392

We removed some deprecated classes / naming conventions above, however it is still good practice to have unique identifiers for important elements within the component, and a lot of overrides in product were using the old deprecated classes. This will also make it much easier to migrate the deprecated classes to the new ones since most of them will be 1-1 mappings. The migration mapping is below:

```
base-input -> d-input__root
base-input__label -> d-input__label
base-input__label-text -> d-input__label-text
base-input__description -> d-input__description
base-input__input -> d-input__input

base-input__icon--left -> d-input-icon--left
base-input__icon--right -> d-input-icon--right

base-input__input--error -> d-input--error
base-input__input--warning -> d-input--warning
base-input__input--success -> d-input--success

d-textarea--success -> d-input--success
d-textarea--warning -> d-input--warning
d-textarea--error -> d-input--error
```

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

migrate product side

### DIFF
--- a/packages/dialtone-css/lib/build/less/components/input.less
+++ b/packages/dialtone-css/lib/build/less/components/input.less
@@ -114,33 +114,6 @@
     }
 }
 
-.d-input__label-text {
-    display: flex;
-    flex: 1 0%;
-    align-items: baseline;
-    justify-content: space-between;
-    box-sizing: border-box;
-    margin-bottom: var(--dt-space-300);
-    color: var(--dt-color-foreground-secondary);
-    font-weight: var(--dt-font-weight-semi-bold);
-    font-size: var(--dt-font-size-200);
-    font-family: inherit;
-    line-height: var(--dt-font-line-height-300);
-    word-break: break-word;
-    overflow-wrap: break-word;
-}
-
-.d-input__description {
-    display: flex;
-    flex-direction: column;
-    box-sizing: border-box;
-    color: var(--dt-color-foreground-tertiary);
-    font-size: var(--dt-font-size-100);
-    font-family: inherit;
-    line-height: var(--dt-font-line-height-400);
-    fill: currentColor;
-}
-
 .d-input__length-description {
     margin-bottom: var(--dt-space-200);
 }
--- a/packages/dialtone-css/lib/build/less/components/input.less
+++ b/packages/dialtone-css/lib/build/less/components/input.less
@@ -270,18 +270,15 @@
 
 //  $$  VALIDATION STATES
 //  ----------------------------------------------------------------------------
-.d-input--warning,
-.d-textarea--warning {
+.d-input--warning {
     --input-color-border: var(--dt-inputs-color-border-warning) !important;
 }
 
-.d-input--error,
-.d-textarea--error {
+.d-input--error {
     --input-color-border: var(--dt-inputs-color-border-critical) !important;
 }
 
-.d-input--success,
-.d-textarea--success {
+.d-input--success {
     --input-color-border: var(--dt-inputs-color-border-success) !important;
 }
 
@@ -290,7 +287,7 @@
 //  ----------------------------------------------------------------------------
 .d-input-icon {
     display: none;
-    
+
     &:not(:empty) {
         display: inline-flex;
         align-items: center;
@@ -299,8 +296,8 @@
     &--right:not(:empty) {
         margin-right: var(--dt-space-400);
     }
-    
+
     &--left:not(:empty) {
         margin-left: var(--dt-space-400);
     }
-} 
+}
--- a/packages/dialtone-vue2/components/input/input.vue
+++ b/packages/dialtone-vue2/components/input/input.vue
@@ -1,10 +1,11 @@
 <template>
   <div
     ref="container"
-    :class="{ 'd-input--hidden': hidden }"
+    :class="['d-input__root', { 'd-input--hidden': hidden }]"
     data-qa="dt-input"
   >
     <label
+      class="d-input__label"
       :aria-details="$slots.description || description ? descriptionKey : undefined"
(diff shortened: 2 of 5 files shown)


packages/dialtone-css/lib/build/less/components/input.less
```diff
@@ -270,18 +270,15 @@
 
 //  $$  VALIDATION STATES
 //  ----------------------------------------------------------------------------
-.d-input--warning,
-.d-textarea--warning {
```
braddialpad, line 274: no reason to have these d-textarea--state classes I think? They do exactly the same thing as d-input--state

packages/dialtone-vue2/components/input/input.vue
```diff
@@ -514,6 +517,7 @@ export default {
   methods: {
     inputClasses () {
       return [
+        'd-input__input',
```
braddialpad, line 520: based on a lot of overrides in the product, it will be easier and less error prone to migrate if we have one single class that identifies this element rather than have to check whether it is d-input OR d-textarea

packages/dialtone-vue2/components/input/input_constants.js
```diff
@@ -45,17 +45,9 @@ export const INPUT_SIZE_CLASSES = {
...
-  },
-
-  textarea: {
-    error: 'd-textarea--error',
-    warning: 'd-textarea--warning',
-    success: 'd-textarea--success',
-  },
+  error: 'd-input--error',
+  warning: 'd-input--warning',
+  success: 'd-input--success',
```
braddialpad, line 50: All we need now that we removed the extra classes

packages/dialtone-vue2/components/input/input.vue
```diff
@@ -15,6 +16,7 @@
           ref="label"
           data-qa="dt-input-label"
           :class="[
+            'd-input__label-text',
```
braddialpad, line 19: @juliodialpad this is an actual class with a lot of styling in `input.less`. Did we remove this on purpose? If so we should also remove it from `input.less`. You can check the percy to see the styling diff.

braddialpad: `d-input__description` as well.

packages/dialtone-css/lib/build/less/components/input.less
juliodialpad: Seems like it, but would that break anything else on product?

packages/dialtone-vue2/components/input/input.vue
juliodialpad: We're using that to correctly set the position of the icons if it's textarea type, that'd affect us too, right?

packages/dialtone-vue2/components/input/input.vue
juliodialpad: Did I remove that? I can't find that line removed on my PR, certainly not intended by me

packages/dialtone-vue2/components/input/input.vue
braddialpad: it's just to allow for a way to reference the input element regardless of whether its input or textarea. No associated styling

packages/dialtone-css/lib/build/less/components/input.less
braddialpad: possible, but it's part of the migration above so should be fine after it's been done.

packages/dialtone-vue2/components/input/input.vue
braddialpad: yeah this line:
https://github.com/dialpad/dialtone/pull/392/files#diff-eada107db256f0591fc3a4fa8c6d065ba54583930249bf3c6a28b9240f55c7a7L20

and this line:
https://github.com/dialpad/dialtone/pull/392/files#diff-eada107db256f0591fc3a4fa8c6d065ba54583930249bf3c6a28b9240f55c7a7L33

juliodialpad: Oh I see, removed all `base-input*` because I couldn't find any reference of style on `input.less` I can't find any reference now either on https://github.com/dialpad/dialtone/blob/staging/packages/dialtone-css/lib/build/less/components/input.less

braddialpad: `base-input__label-text` is not there but `d-input__label-text` is.. I'm actually really confused where this actually came from.

https://github.com/dialpad/dialtone/pull/21/files

Why did we add styles here, but then not apply them to the actual input component?

juliodialpad: I think it was due to the confusion between `base-input*` and `d-input*` 🤔 but yeah, those weren't being applied to the component actually

braddialpad: Alright found this conversation from ages ago.. looks like these classes were just supposed to be deleted. Really confusing lol

https://github.com/dialpad/dialtone-vue/pull/1314/files#r1411122015

juliodialpad: Nice find! yeah, seems like we forgot to remove it 😅

braddialpad: Okay looks good now, percy passing again.

juliodialpad (approved):
This looks cleaner now, thanks Brad!

PR #409, opened 2024-07-15
fix(scrollbar): NO-JIRA fix to explicitly set the viewport

ninamarina:
# Fix scrollbar directive to explicitly set the viewport

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket
NO-JIRA
Related to: https://dialpad.atlassian.net/browse/DP-103842

## :book: Description
Changes the way that OverlayScrollbars are initialized. Instead of passing just the element, it passes an object. This object has `target` (the element to which the plugin will be attached) and `viewport`. This way we are explicitly setting the scrollable viewport and this way all the scroll event listeners can work as expected. 
From the library's docs:

> In the initialization object you can specify how the library handles generated elements. For example, you can specify an existing element as the `viewport' element. Then the library won't generate it, but use the specified element instead

More info: https://github.com/KingSora/OverlayScrollbars/tree/master?tab=readme-ov-file#initialization-with-an-object

This allows for the directive to work in the leftbar of the Dialpad app without having to change the references or how the listeners are attached. The only difference is that the directive has to be attached to the parent `div` (setting `target` and `viewport` to the same element brings problems with the keyboard events).

## :pencil: Checklist

For all PRs:

For all Vue changes:

### DIFF
--- a/packages/dialtone-vue3/directives/scrollbar/scrollbar.js
+++ b/packages/dialtone-vue3/directives/scrollbar/scrollbar.js
@@ -7,7 +7,12 @@ export const DtScrollbarDirective = {
     OverlayScrollbars.plugin(ClickScrollPlugin);
     app.directive('dt-scrollbar', {
       mounted (el, binding) {
-        OverlayScrollbars(el, {
+        OverlayScrollbars({
+          target: el,
+          elements: {
+            viewport: el.children[0],
+          },
+        }, {
           scrollbars: {
             autoHide: `${binding.arg || 'leave'}`,
             clickScroll: true,
--- a/apps/dialtone-documentation/docs/components/scrollbar.md
+++ b/apps/dialtone-documentation/docs/components/scrollbar.md
@@ -71,6 +71,7 @@ Vue.use(DtScrollbarDirective);
 ```
 
 To add a custom overlay scrollbar to a scrollable region, apply the `v-dt-scrollbar` directive to the parent element of the desired region.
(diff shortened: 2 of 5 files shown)


packages/dialtone-vue3/directives/scrollbar/scrollbar.js
```diff
@@ -7,7 +7,12 @@ export const DtScrollbarDirective = {
     OverlayScrollbars.plugin(ClickScrollPlugin);
     app.directive('dt-scrollbar', {
       mounted (el, binding) {
-        OverlayScrollbars(el, {
+        OverlayScrollbars({
+          target: el,
+          elements: {
+            viewport: el.children[0],
```
braddialpad, line 13: Can we always assume `children[0]`? Is it a requirement to only have one child element or could there be more?

braddialpad (commented):
Very nice, good find on this. Just one question.

packages/dialtone-vue3/directives/scrollbar/scrollbar.js
ninamarina: It won't work if there are no children, because there will be no viewport. If there are more than one children it would take the first only as the scrollable element. I can add that to the doc.

braddialpad: Yeah thats a limitation we should highlight if it is the case. Thanks!

braddialpad (approved):
Looks good after you make the doc change you mentioned.

PR #413, opened 2024-07-23
fix(multi-select): NO-JIRA update input padding when resizing

bianca-artola-dialpad:
# [fix(multi-select): NO-JIRA update input padding when resizing](https://github.com/dialpad/dialtone/commit/7cbba705933f0a89ef6bd86357774739ce21bb3d)

Obligatory GIF

- [x] Fix

## :book: Jira Ticket
NO-JIRA

## :book: Description
This pull request is trying to fix an issue that's only reproducible on firespotter side.
When resizing the screen and having a multi-select we have the following listener:
```
 this.resizeWindowObserver = new ResizeObserver(async () => {
      this.setChipsTopPosition();
      this.setInputPadding();
    });
```
So `setInputPadding` is called every time the page is resized. For some reason, the following condition always returns true on the firespotter side and false on the dialtone side:
```
 if (chipsSize > this.initialInputHeight) { input.style.paddingTop
        input.style.paddingTop = `${top}px`;
      }
```
Because this condition is executed on the firespotter side, the input padding is updated each time the user resizes the page. If we have the collapseOnFocusOut prop set to true and the input is not in focus, we shou

## :bulb: Context
When using the collapse-on-focus-out prop on the Firespotter page, I noticed that the input padding was not updated when the window was resized.  I cannot reproduce this bug on the dialtone side, but I think the changes I have made should do the trick, as `setInputPadding` seems to be the function called on resizing.

### DIFF
--- a/packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
+++ b/packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
@@ -581,6 +581,8 @@ export default {
       this.revertInputPadding(input);
       this.popoverOffset = [0, 4];
       if (!lastChip) return;
+      // Avoid adding extra padding when the input is not focused if collapseOnFocusOut is true
+      // This ensures the input returns to its original state when resizing
       if (this.collapseOnFocusOut && !this.inputFocused) return;
 
       // Get the position of the last chip
--- a/packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
+++ b/packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
@@ -581,6 +581,7 @@ export default {
       this.revertInputPadding(input);
       this.popoverOffset = [0, 4];
       if (!lastChip) return;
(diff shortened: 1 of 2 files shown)


packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
```diff
@@ -581,6 +581,7 @@ export default {
       this.revertInputPadding(input);
       this.popoverOffset = [0, 4];
       if (!lastChip) return;
+      if (this.collapseOnFocusOut && !this.inputFocused) return;
```
iropolo, line 586: I would like to provide some `/** */` here since this solution is out of context. (Without mentioning firespotter)

iropolo (approved):
If this fix works for you should be ok to move forward.

packages/dialtone-vue2/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
bianca-artola-dialpad: Sure! I just added them. Let me know if you like them or if I should make any other updates. Thanks!

braddialpad (approved):

PR #414, opened 2024-07-24
fix: DLT-1899 page TOC persisting from page to page

ninamarina:
# fix: DLT-1899 page TOC persisting from page to page

Obligatory GIF

- [x] Fix

## :book: Jira Ticket
[DLT-1899]

## :book: Description
After visiting the Tokens page and reloading, the key `filteredHeaders` remained in the Local Storage, so `Page.vue` would use those for the TOC.
I am reseting the `filteredHeaders` when the page is reloaded.

Also:
- fixes another bug where going to https://dialtone.dialpad.com/tokens/ with the local storage empty would not show the Table of Contents
- removes Tokens page custom class for the grid since it's no longer needed

## Before
toc

## Now
toc-now

[DLT-1899]: https://dialpad.atlassian.net/browse/DLT-1899?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/dialtone-documentation/docs/.vuepress/theme/client.js
+++ b/apps/dialtone-documentation/docs/.vuepress/theme/client.js
@@ -47,6 +47,9 @@ export default defineClientConfig({
 
       provide('currentTheme', currentTheme);
       provide('systemPrefersDark', systemPrefersDark);
+
+      // reset the filtered headers when the page is reloaded
+      localStorage.removeItem('filteredHeaders');
     });
   },
   layouts: {
--- a/apps/dialtone-documentation/docs/.vuepress/theme/client.js
+++ b/apps/dialtone-documentation/docs/.vuepress/theme/client.js
@@ -47,9 +47,6 @@ export default defineClientConfig({
 
       provide('currentTheme', currentTheme);
       provide('systemPrefersDark', systemPrefersDark);
-
-      // reset the filtered headers when the page is reloaded
-      localStorage.removeItem('filteredHeaders');
     });
   },
   layouts: {
--- a/apps/dialtone-documentation/docs/.vuepress/theme/components/Page.vue
+++ b/apps/dialtone-documentation/docs/.vuepress/theme/components/Page.vue
@@ -106,6 +106,7 @@ const includeToc = computed(() => {
 });
 
 watch(route, async () => {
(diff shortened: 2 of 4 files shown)


francisrupert (approved):
Looks good! I can't reproduce it. Defer to anyone else for tech review.

apps/dialtone-documentation/docs/.vuepress/theme/components/Page.vue
juliodialpad, line 1: Why you moved the headers calculation here and not kept on PageToc.vue?

juliodialpad (commented):
I think cleaning the filtered headers should be sufficient to fix this issue, right?

apps/dialtone-documentation/docs/.vuepress/theme/components/Page.vue
ninamarina: Cleaning the headers didn't solve the issue of going directly to  https://dialtone.dialpad.com/tokens/  with the local storage empty. The problem was that the filtered headers was being calculated after `includeToc` was computed. Since there was nothing there, the TOC didn't render, and therefore the `watch` inside is never executed. We need to have the `filteredHeaders` in order to compute `includeToc`, that's why I moved that calculation to the parent Page.

ninamarina: Here's a video of the issue:
toc-tokens

apps/dialtone-documentation/docs/.vuepress/theme/client.js
```diff
@@ -47,6 +47,9 @@ export default defineClientConfig({
 
       provide('currentTheme', currentTheme);
       provide('systemPrefersDark', systemPrefersDark);
+
+      // reset the filtered headers when the page is reloaded
+      localStorage.removeItem('filteredHeaders');
```
braddialpad: What is the point of using localStorage if we're just going to reset it on load? Can't we just use a regular variable then?

ninamarina: Yes, I think it would make more sense 👍🏻

braddialpad (approved):
Thanks!

PR #416, opened 2024-07-24
feat(tokens): DLT-1696 update tokens to output as themes in files

braddialpad:
# feat(tokens): DLT-1696: update tokens to output as themes in files

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

## :book: Description

NOTE: this PR is split into 3 parts for ease of review. Token processing, PostCSS processing, and Documentation site updates. This is the tokens processing PR which includes the updates to tokens studio / style dictionary / sd-transforms

In this PR we update sd-transforms to output into two separate base files containing only base tokens:

- base-light
- base-dark

and separate theme files which contain only semantic tokens:

- dp-light
- dp-dark
- tmo-light
- tmo-dark
- expressive-light
- expressive-dark
- expressive-sm-light
- expressive-sm-dark

I also found I could move android / ios processing all to sd-transforms, you just have to use the custom formatters which we already had. This means we no longer have to run both token-transformer and sd-transforms together!

Fede made some changes to the tokens on the figma side so we can still reference base tokens but not output them as part of the theme.

## :bulb: Context

To implement themes in a robust and more efficient way.

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

Get a test branch going and try this out on product
Update documentation with these new changes


packages/dialtone-tokens/build-android.js
```diff
@@ -4,5 +4,6 @@
  * Builds the android tokens to the correct final package format (output to dist_android)
  */
 
-const mvn = require('maven').create();
-mvn.execute(['package'])
+import Maven from 'maven';
+const mvn = Maven.create();
+mvn.execute(['package']);
```
braddialpad, line 9: shouldn't really be using require in a js file since we are using type module

packages/dialtone-tokens/package.json
```diff
@@ -23,18 +20,19 @@
   },
   "homepage": "https://dialtone.dialpad.com/",
   "devDependencies": {
-    "@tokens-studio/sd-transforms": "^0.14.2",
+    "@tokens-studio/sd-transforms": "^0.12.2",
```
braddialpad: Per sd-transforms website we were actually using a version of sd-transforms that was incompatible with style-dictionary 3. We should really update to style-dictionary 4 but that's outside the scope of this PR.

packages/dialtone-tokens/build-sd-transforms.js
```diff
@@ -43,22 +43,36 @@ export async function run () {
     // use $metadata to iterate through the selected token sets
     // as this contains the correct token set order. The $themes
     // file does not.
-    const source = $metadata
+
+    // includes are treated as SOURCE tokens, meaning they
+    // will be used as a reference but will not be output.
+    const include = $metadata
```
ninamarina, line 44: Does this mean that we can no longer use tokens like `--dt-color-black-700`? Should these be updated in the product?

braddialpad: nope they can still be used. they are still included in the base-tokens file. They are just not included in the brand tokens files where we want to output semantic tokens only

packages/dialtone-tokens/package.json
```diff
@@ -4,10 +4,7 @@
   "description": "Design tokens for Dialtone.",
   "type": "module",
   "scripts": {
-    "build": "pnpm run build:token-transformer && node ./build.js",
-    "build:token-transformer": "pnpm run build:token-transformer:light && pnpm run build:token-transformer:dark",
-    "build:token-transformer:light": "token-transformer tokens token_transformer/tokens-light.json root,base/default,semantic/dp/default,components/checkbox,components/avatar/default,components/badge/default,components/checkbox/default,components/icon/default,components/radio/default root --expandTypography=true --expandShadow=true",
-    "build:token-transformer:dark": "token-transformer tokens token_transformer/tokens-dark.json root,base/default,base/dark,semantic/dp/default,semantic/dp/dark,components/avatar/default,components/badge/default,components/checkbox/default,components/checkbox/dark,components/icon/default,components/radio/default,components/radio/dark root --expandTypography=true --expandShadow=true",
```
juliodialpad, line 10: Nice clean up!

packages/dialtone-tokens/package.json
juliodialpad: Does this version change in any way the format of the tokens? If so, it might impact the Figma variables sync process

juliodialpad (commented):
- What's the difference between `tokens-base-light` and `tokens-dp-light` ? I'm seeing icon size tokens on `tokens-dp-light` but that variable doesn't exist on `tokens-base-light` That's a bit confusing, I'd think that the `dp-light` theme would only include overrides to some tokens like colors and maybe sizes, but those should exist on the base, right?

packages/dialtone-tokens/package.json
braddialpad: hmm do you mean the format of how the tokens are output to file? I don't think that would have changed at all.

braddialpad:
> * What's the difference between `tokens-base-light` and `tokens-dp-light` ? I'm seeing icon size tokens on `tokens-dp-light` but that variable doesn't exist on `tokens-base-light` That's a bit confusing, I'd think that the `dp-light` theme would only include overrides to some tokens like colors and maybe sizes, but those should exist on the base, right?

base: only base tokens (ex size-100, color-black-800)
dp-light: only semantic tokens (ex foreground-primary)

icon size tokens are considered "component tokens" rather than base tokens. Since these are technically semantic tokens, they are included within the semantic files. Icon size token values reference base tokens so they are semantic rather than base.

Perhaps one day we will also need to separate component tokens from theme/brand tokens for something like web components.

packages/dialtone-tokens/package.json
juliodialpad: Yeah, wanted to ask just in case, but its a minor update so it should not change, thanks!

juliodialpad (approved):
This looks good, thanks!

PR #417, opened 2024-07-24
feat(tokens): DLT-1696 generate hsl / composition tokens with postcss

braddialpad:
# feat(tokens) DLT-1696: generate hsl / composition tokens with postcss

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

## :book: Description

NOTE: this PR is split into 3 parts for ease of review. Token processing, PostCSS processing, and Documentation site updates. This is the PostCSS PR which processes hsl and composition tokens within the dialtone-tokens repo itself rather than in dialtone-css. dialtone-css still contains the PostCSS processing for utility classes.

the tokens css files are now output as part of dialtone-css since we now load them on demand.

We are now generating based on actual css declarations in postcss rather than referencing the json file.

We now have to actually write to the docs object from postcss or else composition tokens and such will not be included in our doc.json. The nice thing about this is we have the control to be selective about it. For example we don't want -hsl variables in our documentation, or tokens that exist only to "make up" composition tokens.

Removed duplicated utility class specificity hack we used back when Dialpad was running 2 versions of Dialtone. No longer needed now that the migration is complete.

## :bulb: Context

Easier processing and makes more sense to have all the tokens included in... the tokens.

## :pencil: Checklist

For all PRs:

## :link: Sources

PostCSS API docs: https://postcss.org/api/


packages/dialtone-tokens/build-token-transformer.js
ninamarina, line 1: This file was deleted but it's still being referenced in `./build-ios.js`

packages/dialtone-css/gulpfile.cjs
```diff
@@ -108,6 +109,10 @@ const paths = {
     inputLib: './lib/build/less/dialtone.less',
     outputLib: './lib/dist/',
   },
+  tokens: {
+    input: 'node_modules/@dialpad/dialtone-tokens/dist/css/*.css',
```
juliodialpad, line 113: Wonder if it's safer to reference the relative path to the actual files instead of the installed package.

packages/dialtone-css/lib/build/less/utilities/colors.less
juliodialpad, line 1: Please mention this cleanup in the PR description.

packages/dialtone-css/postcss/dialtone-generators.cjs
juliodialpad, line 1: Nice! this is a lot more easier to follow up now, thanks!

packages/dialtone-tokens/postcss/constants.cjs
```diff
@@ -0,0 +1,190 @@
+/**
+ * This file contains the constants that we use to dynamically generate
+ * the dialtone utility classes and variables.
+ *
+ * Put here the unique values that are needed to generate the classes or mappings
+ * to utilities/variables.
+
+ * This data shouldn't duplicate values under `/docs/_data/` folder
```
juliodialpad: Please update the comment to match the new location and usage

juliodialpad (commented):
Looks good! couple questions/comments before approving.

packages/dialtone-tokens/build-token-transformer.js
braddialpad: thanks! will fix.

francisrupert:
I'm not entirely sure how to design review this one. My initial assumption was one part of this results in all color tokens wind up as a full HSLA value (including the `/ var(--fco)` or `/ var(--bgo)` or `/ var(--bco)` part, such that we can do...

```
<p class="d-fc-critical d-fco50">The quick brown fox...</p>
```
...like we can with...
```
<p class="d-fc-red-300 d-fco50">The quick brown fox...</p>
```

packages/dialtone-css/gulpfile.cjs
braddialpad: could yeah. do you think that's better? In theory either way should work just as well shouldn't it?

packages/dialtone-css/lib/build/less/utilities/colors.less
braddialpad: done!

packages/dialtone-tokens/postcss/constants.cjs
braddialpad: done, also removed some consts that weren't used.

packages/dialtone-css/gulpfile.cjs
juliodialpad: Yeah, either way should work. Guess pnpm installed local package is just a soft-link  so it's the same and it's safer this way.

juliodialpad (approved):
This is ok from the technical perspective, thanks!

PR #418, opened 2024-07-24
feat(tokens): DLT-1696 update docsite for new themes

braddialpad:
# feat(tokens): DLT-1696 update docsite for new themes

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

## :book: Description

NOTE: this PR is split into 3 parts for ease of review. Token processing, PostCSS processing, and Documentation site updates. This is the documentation site updates which includes updating the doc site to handle the new tokens system.

- Added brand selector on the tokens page. This will allow you to display the tokens for a specific brand (dialpad / tmo / expressive) in addition to just a theme (dark / light).
- Use the new isCompositionToken property to prevent displaying "sub-composition" tokens.
- Added a new theme function in dialtone/common that allows you to set the theme and brand. Implemented this into the "theme changer" on the dialtone docsite. This will also set data-dt-brand and data-dt-theme on the root element for ease of use on the application side.

## :bulb: Context

Major changes in theme tokens, docsite needs be updated to document and display them.

## :pencil: Checklist

For all PRs:

Screenshot 2024-07-24 at 3 15 10 PM


ninamarina:
I am seeing some duplicated tokens in the list:

image

This could be my local build though

braddialpad:
> I am seeing some duplicated tokens in the list:
> 
> This could be my local build though

Nope, that appears to be a legitimate problem. Looking into it.

ninamarina:
I am also not seeing the examples of the shadows, seems like some of the tokens that make the composition are undefined: 
image

braddialpad:
> I am also not seeing the examples of the shadows, seems like some of the tokens that make the composition are undefined:

ah yeah looks like only on the focus shadows. thanks

braddialpad:
> I am seeing some duplicated tokens in the list:

Silly logic error by me in the doc generation, should be fixed now.

apps/dialtone-documentation/docs/.vuepress/baseComponents/tokens/AllTokens.vue
```diff
@@ -111,7 +113,6 @@ const updateHeadersRecursively = (node, category) => {
 
 onBeforeMount(() => {
   addTokensToStructure(processedTokens);
-  addComposedTokens(processedTokens);
```
ninamarina, line 114: You can remove this function definition if it's not being used anywhere anymore

braddialpad:
> ah yeah looks like only on the focus shadows. thanks

shadow issue fixed.

ninamarina:
The shadows are looking good!

I see that the tooltips for the composition tokens are no longer there:
image

These are not showing the preview for this PR

braddialpad:
> I see that the tooltips for the composition tokens are no longer there:

Yep, workin on that one

francisrupert:
When switching to `Expressive` Brand, I'm seeing duplicates of at least font size. Haven't found other duplicates.

https://github.com/user-attachments/assets/1640a92c-752b-4fbb-93fa-bb1e5ed48650

braddialpad:
> I see that the tooltips for the composition tokens are no longer there:

These are now fixed!

braddialpad:
> When switching to `Expressive` Brand, I'm seeing duplicates of at least font size. Haven't found other duplicates.

Hmm bit of an edge case here.. Seems like in the expressive brand we are redefining base font sizes in a semantic file which is not really supposed to happen. Only semantic tokens are meant to be defined in a semantic file. I'm guessing this is a case we'll need for the expressive theme though right? In the case of a conflict like this between base and semantic I guess we could just have the one from the semantic file override the base. It already works like that in css (just tested it)

braddialpad:
> When switching to `Expressive` Brand, I'm seeing duplicates of at least font size. Haven't found other duplicates.

Fixed! was actually pretty simple

apps/dialtone-documentation/docs/.vuepress/baseComponents/tokens/TokenValue.vue
```diff
@@ -2,16 +2,7 @@
...
-          <span v-dt-tooltip="getTooltipValue(getCompositionTokenNumerator(value))" class="h:d-fc-secondary">
-            {{ getCompositionTokenNumerator(value) }}
-          </span>
-          /
-          <span v-dt-tooltip="getTooltipValue(getCompositionTokenDenominator(value))" class="h:d-fc-secondary">
-            {{ getCompositionTokenDenominator(value) }}&nbsp;
-          </span>
-        </span>
-        <span v-else v-dt-tooltip="getTooltipValue(value)" class="h:d-fc-secondary">
+        <span v-dt-tooltip="getTooltipValue(value)" :class="{ 'h:d-fc-secondary': getTooltipValue(value) }">
```
ninamarina, line 5: Why add this condition to `h:d-fc-secondary`? Is there a case where `getTooltipValue(value)` is undefined?

ninamarina (approved):
Looks really good!

apps/dialtone-documentation/docs/.vuepress/baseComponents/tokens/TokenValue.vue
braddialpad: Yeah, it's only for the /

ninamarina:
> > When switching to `Expressive` Brand, I'm seeing duplicates of at least font size. Haven't found other duplicates.
> 
> Hmm bit of an edge case here.. Seems like in the expressive brand we are redefining base font sizes in a semantic file which is not really supposed to happen. Only semantic tokens are meant to be defined in a semantic file. I'm guessing this is a case we'll need for the expressive theme though right? In the case of a conflict like this between base and semantic I guess we could just have the one from the semantic file override the base. It already works like that in css (just tested it)

Should this be fixed in Figma at some point in the future? because it means that some tokens are repeated in the base and in the brand file

braddialpad:
> Should this be fixed in Figma at some point in the future? because it means that some tokens are repeated in the base and in the brand file

Yeah, but I guess I'm not entirely sure that's wrong? If you needed to change a base token at theme level like in this case, how else would you do it?

ninamarina:
> > Should this be fixed in Figma at some point in the future? because it means that some tokens are repeated in the base and in the brand file
> 
> Yeah, but I guess I'm not entirely sure that's wrong? If you needed to change a base token at theme level like in this case, how else would you do it?

I guess for now is fine, but I understood that in the future we want base tokens to not change with the theme / brand.

PR #419, opened 2024-07-25
feat(message-input): DLT-1912 add slot for split button

bianca-artola-dialpad:
# feat(message-input): DLT-1912 add slot for split button

Obligatory GIF

- [x] Feature

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DLT-1912

## :book: Description
Add a slot that will be used on the product side to display a split button instead of a button.

## :bulb: Context
On the product side, we have a page that needs to have a split button instead of a button on the news input. To achieve this goal, we should have a slot so we can add it when needed.

Designs: https://www.figma.com/design/PZ8VMSupLJtdE9kb74K61K/%5BDP-86884%5D-Bulk-SMS%2FMMS?node-id=5945-155374&t=zvkFWohTgJ7K9yth-0

### DIFF
--- a/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
+++ b/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
@@ -200,8 +200,8 @@
           </template>
         </dt-button>
 
-        <!-- @slot Slot for button picker -->
-        <slot name="button" />
+        <!-- @slot Slot for sendButton picker -->
+        <slot name="sendButton" />
       </div>
     </section>
   </div>
--- a/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
+++ b/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
@@ -199,6 +199,9 @@
             <p>{{ showSend.text }}</p>
           </template>
         </dt-button>
+
+        <!-- @slot Slot for button picker -->
+        <slot name="button" />
       </div>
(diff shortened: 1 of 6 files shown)


juliodialpad (commented):
Could we rename it to something more specific like `sendButton`?

braddialpad:
Manually re-running the preview.. getting this intermittent error again:

the offset is lower than the number of bytes written. The server has 0 bytes and while 10 bytes has been uploaded - thus 10 bytes are missing. Stopping as this could result in data loss. Initiate a new upload to continue.

bianca-artola-dialpad:
> Could we rename it to something more specific like `sendButton`?

done! thank you. Let's see if previews run now 🙏🏽

packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
```diff
@@ -199,6 +199,9 @@
             <p>{{ showSend.text }}</p>
           </template>
         </dt-button>
+
+        <!-- @slot Slot for sendButton picker -->
+        <slot name="sendButton" />
```
braddialpad: shouldn't this slot wrap the entire existing send button? or do you want it to render beside?

bianca-artola-dialpad: oh I was thinking of setting showSend to false, but this is **_much_** better

braddialpad (approved):

PR #431, opened 2024-08-02
fix(avatar): NO-JIRA make color 000 available

braddialpad:
# fix(avatar): make color 000 available

Obligatory GIF

- [x] Fix

## :book: Description

Added 000 as a color you can set manually on avatar but this color will NOT be used in any random color generation. Documentation updated.

## :bulb: Context

Had someone asking about whether you could set the default grey color as a background for initials on avatar. Noticed that 000 is an actual color in the tokens and setting the color prop to 000 works, but is not currently documented as a valid color.

## :pencil: Checklist

For all PRs:


francisrupert:
I want to be careful here. While yes, the `000` color is the default background color, it definitely shouldn't ever be used for the Initials. What would be a use case when someone would manually set to `000`?

braddialpad:
> I want to be careful here. While yes, the `000` color is the default background color, it definitely shouldn't ever be used for the Initials. What would be a use case when someone would manually set to `000`?

Conversation with Tehuel this morning:

Screenshot 2024-08-02 at 1 03 46 PM

designs here:
https://www.figma.com/design/rqnDsUk9SgbxSLAa4HNrgi/%5BDP-86517%5D-Mark-contacts-as-external-%2F-internal?node-id=5518-21291&m=dev

francisrupert:
> Conversation with Tehuel this morning:
> ... 
> designs here:

OK. I see what they're doing. It looks like there's some design decisions still to be made to really drive it home that an item is an external contact, but this at least is a reasonable addition to allow manually setting it.

francisrupert (approved):

PR #433, opened 2024-08-06
fix: NO-JIRA dark mode in storybook

braddialpad:
# fix: NO-JIRA dark mode in storybook

- [x] Fix

## :book: Description

Fixing the storybook dark mode that I broke. Using the dark mode event to set the theme.

### DIFF
--- a/packages/dialtone-vue2/.storybook/preview.jsx
+++ b/packages/dialtone-vue2/.storybook/preview.jsx
@@ -1,10 +1,12 @@
 import '../css/dialtone-globals.less';
 import '@dialpad/dialtone-css/lib/dist/dialtone.css';
+import { addons } from '@storybook/preview-api';
 import { setTheme } from '@/../../common/themes/config';
 import DpLight from '@/../../common/themes/dp-light.js';
+import DpDark from '@/../../common/themes/dp-dark.js';
 import { MINIMAL_VIEWPORTS } from '@storybook/addon-viewport';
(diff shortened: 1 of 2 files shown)


packages/dialtone-vue2/.storybook/preview.jsx
```diff
@@ -1,20 +1,25 @@
...
+import { addons } from '@storybook/preview-api';
 import { setTheme } from '@/../../common/themes/config';
 import DpLight from '@/../../common/themes/dp-light.js';
+import DpDark from '@/../../common/themes/dp-dark.js';
 import { MINIMAL_VIEWPORTS } from '@storybook/addon-viewport';
 import { DocsContainer } from '@storybook/addon-docs';
-import { useDarkMode } from "storybook-dark-mode";
+import { useDarkMode, DARK_MODE_EVENT_NAME } from "storybook-dark-mode";
 import Vue from 'vue';
-import React from 'react';
```
juliodialpad, line 9: Don't remove this react import, seems like it is not being used here, but it breaks the storybook preview window if removed.

juliodialpad (approved):
Just add the react import back and should be fine, thanks!

packages/dialtone-vue2/.storybook/preview.jsx
braddialpad: Ah yeah, thought that might be there for a reason

PR #434, opened 2024-08-07
feat(message-input): DLT-1947 add slot for sms count badge

bianca-artola-dialpad:
# feat(message-input): DLT-1947 add slot for sms count badge

Obligatory GIF

- [x] Feature

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DLT-1947

## :book: Description
Add a slot that will be used on the product side to display the amount of SMS/segments that the user is sending.

## :bulb: Context
As part of the SMS Count project, we need to add a badge to the message input that displays the amount of SMS/segments the user is sending. The logic for this will be calculated on firespotter side, so I think the best way to handle this is with a slot.
image

### DIFF
--- a/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
+++ b/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
@@ -132,7 +132,9 @@
       <!-- Right content -->
       <div class="dt-message-input__bottom-section-right">
         <!-- @slot Slot for sms count -->
-        <slot name="smsCount" />
+        <div class="d-d-flex d-ai-center">
+          <slot name="smsCount" />
+        </div>
 
         <!-- Optionally displayed remaining character counter -->
         <dt-tooltip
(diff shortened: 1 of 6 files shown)


packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
```diff
@@ -131,6 +131,8 @@
       </div>
       <!-- Right content -->
       <div class="dt-message-input__bottom-section-right">
+        <slot name="smsCount" />
```
ninamarina: Add a comment like this one:
`<!-- @slot Slot for sms count -->`
so that the slot will show in the docs

braddialpad (commented):
Do you need to put a wrapper around this slot with default styling so it is aligned to the other elements? Might be more difficult to do this within the slot itself.

Screenshot 2024-08-07 at 1 54 24 PM

bianca-artola-dialpad:
> Do you need to put a wrapper around this slot with default styling so it is aligned to the other elements? Might be more difficult to do this within the slot itself.
> 
> Screenshot 2024-08-07 at 1 54 24 PM

good call. I added a wrapper with flex and center alignment and looks good now
image

braddialpad (approved):

PR #435, opened 2024-08-07
feat(stack): DLT-1805 add responsive gap prop

ninamarina:
# feat(gap): DLT-1805 enable responsive gap prop

Obligatory GIF

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket
[DLT-1805]

## :book: Description
Allows setting the gap as an object with breakpoints and values.

## :pencil: Checklist

For all PRs:

For all Vue changes:

For all CSS changes:

## :crystal_ball: Next Steps
Add Vue 3

stack-gap

[DLT-1805]: https://dialpad.atlassian.net/browse/DLT-1805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/dialtone-documentation/docs/components/stack.md
+++ b/apps/dialtone-documentation/docs/components/stack.md
@@ -156,6 +156,22 @@ vueCode='
 '
 showHtmlWarning />
 
+Set gap to 600 at extra large screen size, 500 for large screen size, 400 for medium screen size, and 300 for small screen size
+
+<code-well-header>
+    <dt-stack :gap="{ xl: '600', lg: '500', md: '400', sm: '300' }">
+      <div class="d-bgc-magenta-100">
+        Stack item 1
+      </div>
+      <div class="d-bgc-magenta-100">
+        Stack item 2
+      </div>
+      <div class="d-bgc-magenta-100">
+        Stack item 3
+      </div>
+    </dt-stack>
+</code-well-header>
+
 Stacks row with gap 300 and stacks in row reverse the second element with gap 600
 
 <code-well-header>
--- a/apps/dialtone-documentation/docs/components/stack.md
+++ b/apps/dialtone-documentation/docs/components/stack.md
@@ -156,7 +156,7 @@ vueCode='
 '
 showHtmlWarning />
 
-Set gap to 300 at default size, 600 at extra large screen size, 500 for large screen size, 400 for medium screen size, and 300 for small screen size. Check how our breakpoints work [here](/utilities/responsive/breakpoints.html).
+Set 300 as the default gap, 600 at <= XL, 500 at <= L, 400 at <= M, and 300 at <= SM. Check how our breakpoints work [here](/utilities/responsive/breakpoints.html).
 
 <code-well-header>
   <dt-stack :gap="{ default: '300', xl: '600', lg: '500', md: '400', sm: '300' }">
--- a/apps/dialtone-documentation/docs/components/stack.md
+++ b/apps/dialtone-documentation/docs/components/stack.md
@@ -156,10 +156,10 @@ vueCode='
 '
 showHtmlWarning />
 
-Set gap to 600 at extra large screen size, 500 for large screen size, 400 for medium screen size, and 300 for small screen size
+Set gap to 300 at default size, 600 at extra large screen size, 500 for large screen size, 400 for medium screen size, and 300 for small screen size. Check how our breakpoints work [here](/utilities/responsive/breakpoints.html).
 
 <code-well-header>
-  <dt-stack :gap="{ xl: '600', lg: '500', md: '400', sm: '300' }">
+  <dt-stack :gap="{ default: '300', xl: '600', lg: '500', md: '400', sm: '300' }">
     <div class="d-bgc-magenta-100">
       Stack item 1
     </div>
@@ -174,14 +174,14 @@ Set gap to 600 at extra large screen size, 500 for large screen size, 400 for me
 
 <code-example-tabs
 htmlCode='
-<div class="d-stack d-stack--sm-gap-300 d-stack--md-gap-400 d-stack--lg-gap-500 d-stack--xl-gap-600">
+<div class="d-stack d-stack--gap-300 d-stack--sm-gap-300 d-stack--md-gap-400 d-stack--lg-gap-500 d-stack--xl-gap-600">
   <div class="d-bgc-magenta-100">Stack item 1</div>
   <div class="d-bgc-magenta-100">Stack item 2</div>
   <div class="d-bgc-magenta-100">Stack item 3</div>
 </div>
 '
 vueCode='
-<dt-stack :gap="{ xl: `600`, lg: `500`, md: `400`, sm: `300` }">
+<dt-stack :gap="{ default: `300`, xl: `600`, lg: `500`, md: `400`, sm: `300` }">
   <div class="d-bgc-magenta-100">
     Stack item 1
   </div>
--- a/apps/dialtone-documentation/docs/components/stack.md
+++ b/apps/dialtone-documentation/docs/components/stack.md
@@ -118,7 +118,7 @@ showHtmlWarning />
 Stacks column at small screen size and column reverse at large screen
 
 <code-well-header>
-    <div class="d-stack d-stack--row d-stack--sm--column d-stack--lg--column-reverse d-stack--gap-100">
+    <div class="d-stack d-stack--row d-stack--sm-column d-stack--lg-column-reverse d-stack--gap-100">
       <div class="d-bgc-magenta-100">
         Stack item 1
       </div>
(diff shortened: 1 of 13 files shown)


packages/dialtone-css/lib/build/less/components/stack.less
```diff
@@ -42,6 +42,36 @@
...
+  }
+
+  &--500 {
+    .d-stack--gap-500();
+  }
+
+  &--600 {
+    .d-stack--gap-600();
+  }
+}
```
francisrupert: I seem to remember we're aiming for shying away from mixins. I guess the precedent is already set for `DtStack`'s use of `.direction-options()`. 

I don't strictly have an opinion on it since I've lived and breathed mixins w/ SCSS and LESS, so I don't fault its use. Just wondering.

francisrupert:
Recommend renaming PR to `feat(stack): DLT-1805 add responsive gap prop`.

packages/dialtone-css/lib/build/less/components/stack.less
ninamarina: Hm, it's not recommended to use here: https://github.com/dialpad/dialtone/blob/staging/packages/dialtone-vue3/.github/CONTRIBUTING.md#dialtone-usage
Maybe it's referring to some global mixins?

packages/dialtone-css/lib/build/less/components/stack.less
```diff
@@ -91,6 +133,7 @@
 .d-stack--xl {
   @media screen and (max-width: 1264px) {
```
ninamarina, line 134: I am not so sure about this breakpoint, should it be `min-width`?

packages/dialtone-css/lib/build/less/components/stack.less
braddialpad: > I seem to remember we're aiming for shying away from mixins. I guess the precedent is already set for `DtStack`'s use of `.direction-options()`.

This is only on the product side, within dialtone itself it's not a problem.

ninamarina: It's a little bit confusing because it says "Please do not use any mixins or utility classes in Dialtone Vue components." but utilities are allowed on the consumer side

packages/dialtone-css/lib/build/less/components/stack.less
braddialpad: Yeah I think these were accidentally done "backwards" many years ago. We can probably fix them, but we need to be careful about what this may break. Perhaps we need separate story for this.

packages/dialtone-vue2/components/stack/utils.js
```diff
@@ -24,17 +32,36 @@ export function getDefaultDirectionClass (direction) {
...
-  } else { return null; }
+  } else { return []; }
+}
+
+function getResposiveGapClasses (gap) {
+  if (directionPropType(gap) === 'object') {
+    return [
+      ...DT_STACK_RESPONSIVE_BREAKPOINTS.map((breakpoint) => {
+        return DT_STACK_GAP.includes(gap[breakpoint])
+          ? `d-stack--${breakpoint}--gap-${gap[breakpoint]}`
```
braddialpad: based on BEM conventions we should never go over 1 level of nesting: https://getbem.com/faq/#css-nested-elements

Easy fix, just change to a single dash:
```
d-stack--sm-gap-500
```

braddialpad (approved):
Looks good, just one small naming convention change.

francisrupert (changes requested):
Watch my attached screencast. You might also see if what I describe is currently a problem for the Stack's responsive `direction` prop.

https://github.com/user-attachments/assets/27461105-d198-4abc-b4f7-62105cc47622

ninamarina:
> Watch my attached screencast. You might also see if what I describe is currently a problem for the Stack's responsive `direction` prop.
> 
>  Screen.Recording.2024-08-08.at.11.47.09.AM.mov

@francisrupert I updated the media queries here: https://github.com/dialpad/dialtone/pull/435/commits/7b8aa172799f452d238b50d759c5eb335ecded2a
Let me know if those are the expected

apps/dialtone-documentation/docs/components/stack.md
```diff
@@ -156,6 +156,45 @@ vueCode='
 '
 showHtmlWarning />
 
+Set gap to 600 at extra large screen size, 500 for large screen size, 400 for medium screen size, and 300 for small screen size
```
braddialpad: If you were reading just this page you would not know what a large screen size, medium screen size etc is. We should probably link to our current responsive page to show the sizes (I know it's kinda wrong right now due to the reversal issue).

https://dialtone.dialpad.com/utilities/responsive/breakpoints.html

braddialpad (approved):
Seems to work well after the change.

braddialpad:
This could potentially be a breaking change with usage of the existing stack component.. but seems nowhere in DP or DPM the responsive stack object is being used, so I think it should be pretty safe.

packages/dialtone-css/lib/build/less/components/stack.less
```diff
@@ -84,47 +114,53 @@
...
+//  $$  LARGE
 //  ----------------------------------------------------------------------------
-.d-stack--sm {
-  @media screen and (max-width: 480px) {
+.d-stack--lg {
+  @media screen and (min-width: 640px) and (max-width: 980px) {
     .direction-options();
+    .gap-options();
   }
 }
```
francisrupert: This isn't quite how screen mq should work. While yes, recognizing it's backwards, the range you've created with `min-width` and `max-width` limits it to apply **only** within those bounds. That's against the grain of RWD. 

For example...
```less
.d-stack--lg {
  @media screen and (max-width: 980px) {
    ...
  }
}
```
... means "up to large size".

apps/dialtone-documentation/docs/components/stack.md
```diff
@@ -156,6 +156,45 @@ vueCode='
...
+<code-example-tabs
+htmlCode='
+<div class="d-stack d-stack--sm-gap-300 d-stack--md-gap-400 d-stack--lg-gap-500 d-stack--xl-gap-600">
+  <div class="d-bgc-magenta-100">Stack item 1</div>
+  <div class="d-bgc-magenta-100">Stack item 2</div>
+  <div class="d-bgc-magenta-100">Stack item 3</div>
+</div>
+'
+vueCode='
+<dt-stack :gap="{ xl: `600`, lg: `500`, md: `400`, sm: `300` }">
```
francisrupert: Much like `direction` has...
```
:direction="{ `default`: `row`, `sm`: `column`, `lg`: `column-reverse` }"
```
...I was expecting there to be a `default` for `gap` as well, like so...
```
:gap="{ `default`: `600`, `lg`: `300` }"
```
That reads as:
"`600` gap by default. If viewport is `lg` or smaller, the gap is `300` (including for `md` and `sm`)."

Another example:
```
:gap="{ `default`: `600`, `lg`: `300`, `sm`: `100`  }"
```
In that one, the `lg` gap of `300` applies to `lg` **and** `md`, and then the `sm` value overrides that.

So yeah, our media queries are backward by going downward in an **overriding** fashion, rather than **upward**. Changing that for all media queries is a massive breaking change, and is part of the debt we've inherited from Dialtone's initial creation.

francisrupert (changes requested):

apps/dialtone-documentation/docs/components/stack.md
ninamarina: The expected then is that each breakpoint represents the breakpoint or smaller? What is the expected then for the largest breakpoint, xl?

francisrupert: > The expected then is that each breakpoint represents the breakpoint or smaller?

Yeah, that's the "backward" part of RWD we've inherited. I hate it.

> What is the expected then for the largest breakpoint, xl?

That would be `default`, just like the Stack's `direction` prop has. It's not a mq breakpoint, per se, it's just the default it starts with until a narrower breakpoint is declared to do something different. Screencast of `direction` prop, which is what I expect `gap` to follow:

https://github.com/user-attachments/assets/0ed8fc91-7d98-4dd9-ba81-a12344aad828

ninamarina: Oh, I understand now. The default direction is column, but the default gap is 0, that's why you won't get any gap for bigger sizes than "xl" if you don't explicitly define a default. 

image

apps/dialtone-documentation/docs/components/stack.md
```diff
@@ -156,10 +156,10 @@ vueCode='
 '
 showHtmlWarning />
 
-Set gap to 600 at extra large screen size, 500 for large screen size, 400 for medium screen size, and 300 for small screen size
+Set gap to 300 at default size, 600 at extra large screen size, 500 for large screen size, 400 for medium screen size, and 300 for small screen size. Check how our breakpoints work [here](/utilities/responsive/breakpoints.html).
```
francisrupert:
```suggestion
Set 300 as the default gap, 600 at <= XL, 500 at <= L, 400 at <= M, and 300 at <= SM. Check how our breakpoints work [here](/utilities/responsive/breakpoints.html).
```

francisrupert (commented):
Yep. Perfect. Thanks much!

francisrupert (approved):

PR #436, opened 2024-08-09
fix(rich-text-editor): DLT-1822 fix hard break default configuration for rich text editor

emariotti3:
# fix(rich-text-editor): DLT-1822 fix hard break default configuration for rich text editor

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1949

## :book: Description

Add configuration to tip tap Hard Break extension so that when enter key is pressed, the <br> tag will be added to the content.

## :bulb: Context

Editor default Hard Break configuration not adding <br> tags. This is the reason why when the agent when using the Editor component to write emails can enter one or multiple line breaks, but when the email is sent they are not being rendered because the <br> are not being added.

## :pencil: Checklist

For all PRs:

For all Vue changes:

Screenshot 2024-08-09 at 10 50 05

## :link: Sources

[Tip-Tap Hard Break extension docs](https://tiptap.dev/docs/editor/extensions/nodes/hard-break)

### DIFF
--- a/packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
+++ b/packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
@@ -411,7 +411,15 @@ export default {
           }),
         );
       } else {
-        extensions.push(HardBreak);
+        extensions.push(HardBreak.extend({
+          addKeyboardShortcuts () {
+            return {
+              Enter: () => {
+                this.editor.commands.setHardBreak();
+              },
+            };
+          },
+        }));
       }
 
       if (this.mentionSuggestion) {
--- a/packages/dialtone-vue2/recipes/conversation_view/editor/editor.test.js
+++ b/packages/dialtone-vue2/recipes/conversation_view/editor/editor.test.js
@@ -580,6 +580,18 @@ describe('DtRecipeEditor tests', () => {
       });
     });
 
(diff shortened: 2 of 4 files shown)


juliodialpad:
IIRC the behavior we we're looking for was to add new lines on `Shift + Enter`, it works that way currently.

- With `Enter` key
https://github.com/user-attachments/assets/8459b475-24f6-401d-936d-7df73b1898b6

- With `Shift + Enter` 
https://github.com/user-attachments/assets/5fd6e3b6-208c-4daa-a34c-436b55c3f139

I think the real problem is that only pressing `Enter` the line break shouldn't be happening like if it was actually entering it.

braddialpad:
> IIRC the behavior we we're looking for was to add new lines on `Shift + Enter`, it works that way currently.

I think we probably need to support both, I think shift + enter makes sense for the message input, but for the editor we probably want just enter to make a new line.

emariotti3:
@juliodialpad  as @braddialpad says, this is something we added for the Editor to be used in the email composer, where the agent will be pressing enter and expect line breaks to be introduced. Initially when I added the Editor component I introduced a prop to enable this configuration for this case

braddialpad (approved):

PR #444, opened 2024-08-13
fix: NO-JIRA es6 errors with json imports

braddialpad:
# fix: es6 errors with json imports

Obligatory GIF

- [x] Fix

## :book: Description

Trying to get this one out quickly to unblock the team that is having this problem. 

["import attribute"](https://github.com/tc39/proposal-import-attributes?tab=readme-ov-file) syntax is required on json imports as of newer versions of node due to some sort of security issue. I think this may only be happening in the dx-console project because they have their tsconfig set to "esnext". Going forward this will be the correct way to import json though so we should change it.

I'm putting in this quick fix in because the version of eslint we're on doesn't support this syntax at all and completely errors out, you cant even ignore the error in the file using a comment. 

The real fix should be updating to ESLint 9 which supports this, but that's a huge breaking change to configuration files and will be alot of work to migrate.

I confirmed that the tests do run correctly using pnpm link.

Also it says that firefox doesn't support this on caniuse: https://caniuse.com/?search=import%20attributes

However I tested in firefox and everything seemed fine. Maybe vite/rollup is doing some kind of magic?

## :link: Sources

https://github.com/tc39/proposal-import-attributes
https://caniuse.com/?search=import%20attributes

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,19 +1,3 @@
-# dependencies
-node_modules
-
-# storybook
-storybook-static
-
-# tests
-tests/.cache
-tests/compiled
-
-# ejs files
-**/*.ejs
-
-# builds
-dist
-
 # TEMPORARY can be removed once we upgrade to eslint 9
 **/common/emoji/index.js
 **/common/storybook_utils.js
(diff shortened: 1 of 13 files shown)


juliodialpad (approved):
Quick fix looks good, we'll have an eye open in case something fails related to this.

juliodialpad:
Don't forget to add the Jira ticket to PR Title

PR #451, opened 2024-08-14
fix(message-input): DP-107570 fix long names on mentions

bianca-artola-dialpad:
# fix(message-input): DP-107570 fix long names on mentions

Obligatory GIF

- [x] Fix

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DP-107570

## :book: Description
Added some styles to improve the UI when having long names on mentions

https://github.com/user-attachments/assets/b605053b-4ef7-4bf8-bbfd-f429d8431618

## :bulb: Context
QA reported an UI issue in which the mention dropdown was not looking as expected when users have long names
Old UI:
image

### DIFF
--- a/packages/dialtone-vue2/components/rich_text_editor/extensions/mentions/MentionSuggestion.vue
+++ b/packages/dialtone-vue2/components/rich_text_editor/extensions/mentions/MentionSuggestion.vue
@@ -9,7 +9,9 @@
       :image-alt="name"
       size="xs"
     />
-    {{ name }}
+    <span class="mention-suggestion-name">
+      {{ name }}
+    </span>
   </dt-stack>
 </template>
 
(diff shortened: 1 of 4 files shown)


packages/dialtone-vue2/components/rich_text_editor/extensions/mentions/MentionSuggestion.vue
```diff
@@ -9,7 +9,9 @@
       :image-alt="name"
       size="xs"
     />
-    {{ name }}
+    <span class="mention-suggestion-name">
+      {{ name }}
+    </span>
```
bianca-artola-dialpad, line 14: I can also make this here:
```suggestion
    <span class="d-truncate">
      {{ name }}
    </span>
```
and skip the eslint warning with `<!-- eslint-disable vue/no-restricted-class -->`
Let me know what works better for you

braddialpad: nah the way you did it is better. We don't want to use util classes in components because of `!important`

braddialpad (approved):
Thanks!

PR #454, opened 2024-08-15
fix: NO-JIRA use require for json imports

braddialpad:
# fix: NO-JIRA use require for json imports

- [x] Fix

## :book: Description

Follow up to https://github.com/dialpad/dialtone/pull/444 which worked with pnpm link but not with the actual production package.

It appears json imports are an absolute mess in ES6. There is no standardization and the correct way to import it changes every 2 months and different 3rd party libraries have different levels of support. Just added require support to vite via a plugin so we can be done with this waste of time.

## :pencil: Checklist

For all PRs:

### DIFF
--- a/packages/dialtone-icons/vue2/package.json
+++ b/packages/dialtone-icons/vue2/package.json
@@ -11,9 +11,10 @@
     "glob": "^10.3.10",
     "npm-run-all": "^4.1.5",
     "vite": "^5.0.12",
+    "vite-plugin-require": "1.2.14",
     "vue": "^2.7.16",
-    "vue-tsc": "^1.8.25",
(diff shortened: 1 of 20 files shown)


juliodialpad (commented):
So this means that from now on, we need to require instead of importing json files?

Can you share a link we're I could find more info about this, please?

braddialpad:
> So this means that from now on, we need to require instead of importing json files?
> 
> Can you share a link we're I could find more info about this, please?

Not even sure if we're going to keep this. I'm going to test if it even works first.

braddialpad:
Yep doesn't work as I suspected. vite just converts everything back to `import` statements when it bundles it. I will probably just revert this.

PR #456, opened 2024-08-15
fix(docs): DLT-1994 use dialtone scrollbar in docsite

braddialpad:
# fix(docs): DLT-1994 use dialtone scrollbar in docsite

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1994

## :book: Description

Update the main scrollbar of the docsite to use the new dialtone scrollbar. Could not use the directive because the vuepress scrollbar actually exists outside of the vue instance on the body, so just used overlayscrollbars directly.

Can probably update the leftbar as well, but I have one technical concern I want to discuss outside the scope of this PR.

## :bulb: Context

Mostly because this "jumping" problem annoys me.

## :pencil: Checklist

For all PRs:

- [x] I have considered the performance impact of my change..

## :link: Sources

https://github.com/KingSora/OverlayScrollbars

### DIFF
--- a/apps/dialtone-documentation/docs/.vuepress/theme/client.js
+++ b/apps/dialtone-documentation/docs/.vuepress/theme/client.js
@@ -2,27 +2,53 @@ import { defineClientConfig } from '@vuepress/client';
 import Layout from './layouts/Layout.vue';
 import NotFound from './layouts/NotFound.vue';
 import customEmojis from '@data/custom-emoji.json';
+import { OverlayScrollbars, ClickScrollPlugin } from 'overlayscrollbars';
+import { onBeforeMount, provide, ref } from 'vue';
+import { flushPromises } from '@workspaceRoot/common/utils';
 
 // CSS
 import '@dialpad/dialtone-css/lib/dist/dialtone.css';
 import '@dialpad/dialtone/vue3/css';
 import './assets/less/dialtone-docs.less';
 import './assets/less/dialtone-syntax.less';
-import { onBeforeMount, provide, ref } from 'vue';
+
+// The default scrollbar exists outside of the vue instance on the body so
+// we cannot use the vue directive for our custom scrollbar. Init it manually here.
+const initOverlayScrollbars = () => {
+  return new Promise((resolve) => {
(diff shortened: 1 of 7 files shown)


ninamarina:
In Firefox the scroll works correctly, but the scrollbar doesn't follow along:
firefox-scroll

ninamarina (changes requested):
It's not working correctly in Firefox. In chrome is ok

braddialpad:
Looks like they added support in version 2.3.0 for something called ScrollTimeline API, which firefox doesn't support. It seems like overlayscrollbars is only supposed to use it in supported browsers but that is clearly not working here. Downgrading to version 2.2.0, before this was implemented fixed the issue. I'll report a bug on their github.

https://github.com/KingSora/OverlayScrollbars/releases/tag/v2.3.0

braddialpad:
Actually... seems to be silently fixed in 2.10.0, so just updated to that version.

francisrupert (approved):
> Mostly because this "jumping" problem annoys me.

Good man 😛

> Could not use the directive because the vuepress scrollbar actually exists outside of the vue instance on the body, so just used overlayscrollbars directly.

Does this imply we should make "use overlayscrollbars directly" an option in the documentation? 

Overall, this is highly satisfying to see. Since you and @ninamarina have implemented it several times, I wonder how much effort all of [DLT-1984 epic](https://dialpad.atlassian.net/browse/DLT-1984) would be.

> Can probably update the leftbar as well...

I'm assuming that'd be [DLT-1878, Root Layout](https://dialpad.atlassian.net/browse/DLT-1878)

ninamarina (approved):
Looks good, thanks!

braddialpad:
> Does this imply we should make "use overlayscrollbars directly" an option in the documentation?

Yeah there could be some use cases for this, though it would be pretty rare currently. Created jira here: https://dialpad.atlassian.net/browse/DLT-1997

> I'm assuming that'd be [DLT-1878, Root Layout](https://dialpad.atlassian.net/browse/DLT-1878)

Yep

PR #457, opened 2024-08-16
fix(message-input): DLT-1996 fix line breaks when initializing input value

bianca-artola-dialpad:
# fix(message-input): DLT-1996 fix line breaks when initializing input value

Obligatory GIF

- [x] Fix

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DLT-1996

## :book: Description
We need a solution to ensure line breaks function correctly across both the product and Dialtone. The primary issue is that when the message input component is initialized with a value, we must use `<br>` tags instead of \n to represent line breaks. This is problematic because, on the product side, we rely on \n as `<br>` tags are not supported by the back-end.

After discussing various approaches with @braddialpad, I believe the best solution is to handle this on the Dialtone side. This approach minimizes the number of replacements required in Firespotter. Initially, I considered managing this on the Firespotter side, but I think that solution can bring some issues on the future (like when having SMS count projects to correctly bill our clients, it would be better if we can just display on the input what we will send) and also would be better to don't modify/update the text we send. In case you wanna check, these would be the changes needed on firespotter if we don't go with this one: https://github.com/dialpad/firespotter/pull/45975

## Context
We encounter two scenarios where the input needs to be initialized with text:

- When editing a message
- When working with drafts

Currently, if the text contains line breaks, it doesn’t function well, as the product side uses \n for line breaks, which isn’t compatible with Dialtone.

### DIFF
--- a/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
+++ b/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
@@ -665,6 +665,12 @@ export default {
     },
   },
 
+  created () {
+    if (this.value && this.outputFormat === 'text') {
+      this.internalInputValue = this.value.replace(/\n/g, '<br>');
+    }
+  },
+
   methods: {
     // Mousedown instead of click because it fires before the blur event.
     onMousedown (e) {
--- a/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.stories.js
+++ b/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.stories.js
@@ -216,6 +216,13 @@ export const Default = {
   render: (argsData) => createRenderConfig(DtRecipeMessageInput, DtRecipeMessageInputDefaultTemplate, argsData),
 };
 
+export const InitializeWithLineBreaks = {
+  render: (argsData) => createRenderConfig(DtRecipeMessageInput, DtRecipeMessageInputDefaultTemplate, argsData),
+  args: {
+    value: 'Always the Padawan,\n never the Jedi.',
+  },
+};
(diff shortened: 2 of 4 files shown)


packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
```diff
@@ -665,6 +665,12 @@ export default {
     },
   },
 
+  created () {
+    if (this.value && this.outputFormat === 'text') {
+      this.internalInputValue = this.value.replace(/\n/g, '<br>');
+    }
+  },
```
braddialpad, line 672: This only needs to be done on created? Couldn't this potentially be set after the message input is initialized as well?

braddialpad (approved):
It's certainly a bit of a hack but I think it will all be fixed once we make future changes to the backend.

packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
bianca-artola-dialpad: I think we should be fine with doing this only on creation. When the input is already initialized, it means user is writing or on the same page; when adding a line break there, it is correctly calculated as a `\n`. So I think we should be fine

https://github.com/user-attachments/assets/1be87a5a-7d17-4aa9-bee9-0c14eaf1c73e

I agree that this is a bit tricky, but I couldn't find anything better without modifying the backend.

braddialpad: yeah no prob, the backend definitely needs a lot of changes to work smoothly with this.

PR #461, opened 2024-08-20
fix(docs): NO-JIRA popover with dt-scrollbar

braddialpad:
# fix(docs): popover with dt-scrollbar

Obligatory GIF

- [x] Fix

## :book: Description

There are some pretty big differences in how to make popovers / modals disable the scrollbar when they are opened when you are using a custom scrollbar on the root/body as opposed to a native scrollbar. We would have also had this problem if we used dt-scrollbar on the body in any other application, it is not only specific to our docsite.

Made changes so that this will work with popover. Also still backwards compatible with native scrollbars.

Modal is still messed up. Going to work on that one next. Separate PR.

## :bulb: Context

Francis noticed popovers were pretty broken on the docsite after we switched it over

## :pencil: Checklist

For all PRs:

For all Vue changes:

For all CSS changes:

## :crystal_ball: Next Steps

Do the same thing for modal... and check if there is anything else messed up because of this.


apps/dialtone-documentation/docs/.vuepress/exampleComponents/ExamplePopover.vue
braddialpad, line 1: This entire example component was using old non vue components from the days before we could render vue components 😨. Migrated it to get rid of this inconsistency.

packages/dialtone-css/lib/build/less/components/scrollbar.less
```diff
@@ -1,4 +1,4 @@
-.scrollbar {
+.d-scrollbar {
```
braddialpad, line 1: Changed to d-scrollbar to avoid collisions

packages/dialtone-css/lib/build/less/dialtone-transitions.less
braddialpad, line 1: Moved transitions over because it was causing me problems in the docsite. This completes bug https://dialpad.atlassian.net/browse/DLT-1952

francisrupert:
> "Modal is still messed up. Going to work on that one next. Separate PR."

Algolia search as well? 

image

ninamarina (approved):
Looks good for all the popover examples. As @francisrupert mentioned, there's still the case of the Algolia search, and modals.

PR #472, opened 2024-08-26
fix: NO-JIRA eslint documentation links

ninamarina:
# Fix eslint plugin doc links

Update the links to point to the correct URL. They were probably like this since the migration to the monorepo.

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor


juliodialpad (approved):
Thanks for fixing this!

PR #473, opened 2024-08-26
feat(filter-pill): DLT-2704 create component

juliodialpad:
# Add DtFilterPill component

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2704

## :book: Description

- Added filter-pill component on both Vue 2 and Vue 3 versions
- Added simple filter-pill docs

## :bulb: Context

This is the rework of filter pill component after @francisrupert finished #484

## :pencil: Checklist

For all PRs:

For all Vue changes:

For all CSS changes:

If new component:

- I am exporting any new components or constants:
  - [x] from the index.js in the component directory.
  - [x] from the index.js in the root (either `packages/dialtone-vue2` or `packages/dialtone-vue3`).
- [x] I have added the styles for the new component to the `packages/dialtone-css` package.
- [x] I have created a page for the new component on the documentation site in `apps/dialtone-documentation`.
- [x] I have added the new component to `common/components_list.cjs`
- [x] I have created a component story in storybook
- [x] I have created story / stories for any relevant component variants in storybook
- [x] I have created a docs page for the component in storybook.
- [x] I have checked that changing all props/slots via the UI in storybook works as expected.

Captura de pantalla 2025-09-11 a las 4 31 15 p m

## Links
- https://dialtone.dialpad.com/deploy-previews/pr-473/components/filter-pill.html
- https://dialtone.dialpad.com/vue/deploy-previews/pr-473/?path=/story/components-filter-pill--default
- https://dialtone.dialpad.com/vue3/deploy-previews/pr-473/?path=/story/components-filter-pill--default


packages/dialtone-vue2/components/filter_pill/filter_pill.vue
```diff
@@ -0,0 +1,194 @@
+<template>
+  <dt-popover
```
francisrupert: Does this restrict it directly just Popover? Might be a Dropdown, which still is a Popover. I can't think of another thing it would trigger though.

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
```diff
@@ -0,0 +1,194 @@
...
+    class="d-filter-pill"
```
francisrupert, line 3: Will need to support all Button sizes, much like Split Button.

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
```diff
@@ -0,0 +1,194 @@
...
+            v-else
+            :text-option="{ width: loadingSkeletonWidth }"
+          />
+          <template
+            #icon
+          >
+            <dt-icon-chevron-down
+              data-qa="dt-filter-pill__icon"
+              class="d-filter-pill__icon"
+              size="200"
```
francisrupert: Icon size would be the corresponding Button size's icon size.

francisrupert (commented):
Great start! Functionality, props, and slots are as I expected. The design very likely will change the raw html/css.

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
juliodialpad: Yes, it's restricted directly to just the popover slots, when refining the ticket we didn't found other usages that were not using popover so we decided to make it this way.

francisrupert: Would it work with DtDropdown? Current Filter Pills may not invoke Dropdown, but it definitely should be an option.

juliodialpad: That could be a future improvement if needed, I would need to reestructure the component right now to add that support

francisrupert (approved):
Approved at least in current designed state. Defer to others for technical review.

packages/dialtone-vue3/components/filter_pill/filter_pill.vue
```diff
@@ -0,0 +1,203 @@
+<template>
+  <dt-popover
+    class="d-filter-pill"
+    data-qa="dt-filter-pill"
+    :open="isPopoverOpen"
+    @update:open="isPopoverOpen = $event"
```
braddialpad: Shouldn't we just use v-model / sync if we're going to do this?

braddialpad: Also should this component support v-model / sync?

packages/dialtone-css/lib/build/less/components/filter-pill.less
```diff
@@ -0,0 +1,53 @@
...
+    color: var(--dt-color-foreground-muted);
+  }
+
+  &__reset {
+    position: absolute;
+    top: 50%;
+    right: 0;
+    color: var(--dt-color-foreground-primary);
+    transform: translateY(-50%);
+    opacity: .5;
```
braddialpad: var(--dt-opacity-700)

packages/dialtone-css/lib/build/less/components/filter-pill.less
```diff
@@ -0,0 +1,53 @@
...
+
+    &:hover, &:focus-visible {
+      opacity: 1;
```
braddialpad: var(--dt-opacity-1300)

packages/dialtone-css/lib/build/less/components/filter-pill.less
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  &__icon {
+    color: var(--dt-color-foreground-muted);
+  }
+
+  &__reset {
+    position: absolute;
+    top: 50%;
+    right: 0;
```
braddialpad: var(--dt-space-0)

packages/dialtone-css/lib/build/less/components/filter-pill.less
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+
+  &__icon {
+    color: var(--dt-color-foreground-muted);
+  }
+
+  &__reset {
+    position: absolute;
+    top: 50%;
```
braddialpad: var(--dt-space-50-percent)

packages/dialtone-css/lib/build/less/components/filter-pill.less
```diff
@@ -0,0 +1,53 @@
...
+    opacity: .5;
+
+    &:hover, &:focus-visible {
+      opacity: 1;
+    }
+  }
+
+  &--active {
+    background-color: var(--dt-action-color-background-base-hover);
+    border: 1px solid var(--dt-action-color-border-base-outlined-default);
```
braddialpad: var(--dt-size-border-100)

braddialpad (commented):
Looks pretty great so far, will review more tomorrow morning.

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
```diff
@@ -0,0 +1,203 @@
...
+      default: undefined,
+    },
+
+    /**
+     * Determines if the filter is loading
+     */
+    loading: {
+      type: Boolean,
+      default: false,
+    },
```
braddialpad, line 211: Hmm not sure about this loading implementation. Why did we use a skeleton instead of the built in "loading" prop on button that shows a spinner? This looks a bit strange:

Screenshot 2024-08-28 at 9 56 16 AM

There was also no loading prop requirement in the ticket https://dialpad.atlassian.net/browse/DLT-1917. Not to say it won't be useful at some point, but I didn't find any existing implementations of filter pill that show a loading indicator within the button.

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
```diff
@@ -0,0 +1,203 @@
...
+
+    /**
+     * Sets the size of the skeleton
+     * accepts CSS width attribute values
+     */
+    loadingSkeletonWidth: {
+      type: String,
+      default: '128px',
+    },
```
braddialpad: Changing this changes the entire width of the button rather than the width of the skeleton. Intended?

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
```diff
@@ -0,0 +1,203 @@
+<template>
+  <dt-popover
+    class="d-filter-pill"
+    data-qa="dt-filter-pill"
```
braddialpad, line 4: We probably want to pass through "placement" and "fallbackPlacements" prop since they're very commonly used.

placement can default to "bottom-start"
fallbackPlacements can default to ["top-start", "auto"]

braddialpad: A couple other props I think we'll need to pass through

appendTo
maxHeight
maxWidth
padding

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
```diff
@@ -0,0 +1,203 @@
...
+          importance="clear"
+          kind="muted"
+          :size="size"
+          :aria-label="resetButtonAriaLabel"
+          @click="$emit('reset', $event)"
+        >
+          <template #icon="{ iconSize }">
+            <dt-icon-close :size="iconSize" />
+          </template>
+        </dt-button>
```
braddialpad: Screenshot 2024-08-28 at 10 20 02 AM

Size of the button feels really big and actually overlaps the text at XL size

lint-staged.config.cjs
```diff
@@ -1,11 +1,27 @@
...
+const removeIgnoredFiles = async (files) => {
+  const eslint = new ESLint();
+  const isIgnored = await Promise.all(
+    files.map((file) => {
+      return eslint.isPathIgnored(file);
+    }),
+  );
+  const filteredFiles = files.filter((_, i) => !isIgnored[i]);
+  return filteredFiles.join(' ');
+};
```
braddialpad: Is this because lint-staged is passed filenames directly so it wasn't ignoring files that ESLint normally ignores?

braddialpad (changes requested):
Thanks, couple more comments from this morning. I screen reader tested as well, looks good!

packages/dialtone-vue3/components/filter_pill/filter_pill.vue
juliodialpad: > Also should this component support v-model / sync?

Didn't found any use-case on product that had the filter previously opened, or do you mean for something like the "active" state?

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
juliodialpad: Makes sense, I wasn't thinking on those, thanks!

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
juliodialpad: Can't find the button I saw using that, but I copied the functionality from there.

Makes sense to use the loading prop of the button as we don't want to keep introducing new loaders.

It made sense for me as there was actually going to be text there, but I'll change it to the `loading` prop, it's going to be easier to maintain.

juliodialpad: - Loading prop
Screenshot 2024-09-02 at 3 33 25 p m

- Loading skeleton
Screenshot 2024-09-02 at 3 35 02 p m

Actually, the `loading` prop of the button replaces the entire button content. I'd prefer to maintain the loading skeleton in that case.

@francisrupert can you help us to decide here please?

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
juliodialpad: As the skeleton is growing/shrinking it'd be expected that the button size change, right? The use-case I saw for this was setting the loading-skeleton-width to the actual text-size which could be calculated either on the client or I can add that here but I'd be more flexible if I left the client to decide the width of the loading skeleton.

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
juliodialpad: The button is the same size as the other button, should be ok. What they're doing on product is only changing the "X" icon color but not changing the background color on hover. I can do the same here but I wanted to keep the button implementation.

I thinks this might change with the upcoming @francisrupert updates

lint-staged.config.cjs
juliodialpad: Yeah, forgot to put the link to the reference: https://github.com/lint-staged/lint-staged?tab=readme-ov-file#eslint--7-1

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
francisrupert: Preview of [WIP Design](https://dialtone.dialpad.com/deploy-previews/pr-484/components/filter-pill.html), via Draft PR https://github.com/dialpad/dialtone/pull/484.

juliodialpad: I think I should wait until you have the final HTML implementation so I can make the appropriate changes here.

Maybe closing this PR and open a new one with the final implementation, what do you think? @francisrupert @braddialpad

braddialpad: sure, or you can just PR a new PR into this branch. I don't think there's really any reason to release this early.

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
braddialpad: If the client is receiving the text from a server it won't know its size until the request is complete in which case it is done loading. Calculating width based on the length of a string even if we had it is VERY difficult due to variable width fonts, i18n, and text scaling. Basically the same problem we had in this ticket where we were trying to dynamically change the size of the field based on the language and it was impossible to size it correctly.

https://dialpad.atlassian.net/browse/DP-107950

juliodialpad: Ok, setting it to the text size was too ambitious 😅 but having the flexibility to change the size is ok, right?

juliodialpad:
Closing this PR as we haven't done anything related to this since Sep 2024

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
```diff
@@ -0,0 +1,322 @@
...
+    omegaTooltipText: {
+      type: String,
+      default: '',
+    },
+
+    /**
+     * Padding size class for the popover content.
+     * @values none, small, medium, large
+     */
+    padding: {
```
francisrupert: I know you're carrying over the `padding` prop from Popover, but this prop name might make one think it's padding for the pill button. Would it be reasonable to be `popoverPadding`? And perhaps prepend the other Popover props with `popover`?

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
```diff
@@ -0,0 +1,322 @@
...
+      default: false,
+    },
+
+    /**
+     * The size of the button.
+     * @values xs, sm, md, lg, xl
+     */
+    size: {
+      type: String,
+      default: 'md',
```
francisrupert:
```suggestion
      default: 'sm',
```

I'd say nearly all Filter Pill uses will be this size.

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
```diff
@@ -0,0 +1,322 @@
...
+      default: () => ['top-start', 'auto'],
+    },
+
+    /**
+     * Label of the button
+     */
+    label: {
+      type: String,
+      default: undefined,
+    },
```
francisrupert: I'd like to make `label` a slot, and the default slot would fall back to the label prop. We anticipate the flexibility of something like...
```html
<dt-filter-pill active>
  <dt-stack direction="row" gap="300">
    <span>Person</span>
    <span class="d-fw-bold">=</span>
    <span>Sarah Jones</span>
  </dt-stack>
  <template #content>
    ...
  </template>
</dt-filter-pill>
```

Image

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
```diff
@@ -0,0 +1,322 @@
+<template>
+  <div
+    class="d-filter-pill"
+    data-qa="dt-filter-pill"
+  >
+    <dt-popover
```
francisrupert, line 6: How would DtDropdown work?

Image

francisrupert (changes requested):
Sweet. Looking pefect!

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
```diff
@@ -0,0 +1,322 @@
...
+          :disabled="disabled"
+          :kind="buttonKind"
+          :size="size"
+          data-qa="dt-filter-pill__button"
+          icon-position="right"
+          importance="outlined"
+          @click="isPopoverOpen = true"
+        >
+          <span class="d-filter-pill__label">
+            <span class="d-filter-pill__label-alpha">{{ label }}</span>
```
francisrupert:
```suggestion
            <span class="d-filter-pill__label-alpha">
              <slot>{{ label }}</slot>
            </span>
```
See previous default slot comment.

packages/dialtone-vue2/components/filter_pill/filter_pill.vue
juliodialpad: Hmm, working on it, will have to refactor to make it work with popover and dropdown.

juliodialpad: Resolved all the other comments, I'm having a hard time trying to achieve this in an optimized way.

Tried with creating two components like `DtRecipeCallbarButtonWithDropdown` and `DtRecipeCallbarButtonWithPopover` but I don't like that pattern too much as it's repeating a lot of code which is not ideal.

Tried `<component :is` but I had to add a variable named `type` to distinguish if it needed to use dropdown or popover which I don't like as most developers might not know which one they should use.

Tried changing completely to use `DtDropdown` as base instead of `DtPopover` but the keyboard navigation is not working quite well, clicking the item works, but keyword interactivity doesn't work properly.

We should try to recreate the product components to see if the functionality can be replaced by what we have currently and then create another ticket to improve the filter pill or split in two in case that's needed.

juliodialpad: cc @braddialpad @iropolo @ninarepetto

iropolo: Seems like this component hold a tricky behaviour.

Maybe you can bring it in today's grooming. We can discuss it a little bit there :)

francisrupert (commented):
This is great. Nice solution to enable customizing label and popover. 

### Examples

I lazily and quickly prototyped on this branch https://github.com/dialpad/dialtone/tree/feat/filter-pill-component-vue--FRANCIS-SPIN. Defer to your judgement on doing it "correctly", but should give you the idea.

Can you update the custom default slot and custom popover content examples (both in doc site and Vue stories) to something like these...

#### Default slot example of custom label structure

https://github.com/user-attachments/assets/c75ab4a8-bd78-47d5-87fd-cd41e5f4284f

#### Custom popover content

If I understand it correctly, you can't directly use DtDropdown in place of the popover but you can achieve it with DtListItems. Unless you can... and I couldn't figure it out?

https://github.com/user-attachments/assets/7376dfba-269a-46f4-90b5-8ea2e8ba80cd

Thanks much!!

juliodialpad:
> This is great. Nice solution to enable customizing label and popover.
> 
> ### Examples
> I lazily and quickly prototyped on this branch https://github.com/dialpad/dialtone/tree/feat/filter-pill-component-vue--FRANCIS-SPIN. Defer to your judgement on doing it "correctly", but should give you the idea.
> 
> Can you update the custom default slot and custom popover content examples (both in doc site and Vue stories) to something like these...
> 
> #### Default slot example of custom label structure
>  Screen.Recording.2025-09-17.at.8.54.40.AM.mov 

Updated on https://github.com/dialpad/dialtone/pull/473/commits/1a5d6a4d05e928132395ca70e1193b99986a3851


> #### Custom popover content
> If I understand it correctly, you can't directly use DtDropdown in place of the popover but you can achieve it with DtListItems. Unless you can... and I couldn't figure it out?
> 
>  Screen.Recording.2025-09-17.at.9.56.43.AM.mov 
> Thanks much!!

You can implement list-like single element selection, but the keyboard navigation is a bit complex to achieve and it'd make this component a complex one, which I don't think is required as the implementations I saw on product are completely custom, mostly what they share is only the button styling.

As @iropolo suggested, we should go with this component, try to migrate some usages on product and see what we're actually missing so we can iterate over this component and do the appropriate changes

braddialpad (approved):
Looks good, really nice work

braddialpad:
> You can implement list-like single element selection, but the keyboard navigation is a bit complex to achieve and it'd make this component a complex one, which I don't think is required as the implementations I saw on product are completely custom, mostly what they share is only the button styling.
> 
> As @iropolo suggested, we should go with this component, try to migrate some usages on product and see what we're actually missing so we can iterate over this component and do the appropriate changes

Yes there are way too many different usages inside the filter pill popover to be able to handle them all within this component. We can possibly create new recipes on top of this component that define more specific focused cases, but this component should be the base. It already has enough complexity.

iropolo (approved):

francisrupert (approved):
Awesome, thanks for seeing this through patiently.

PR #476, opened 2024-08-26
fix(tokens): NO-JIRA move themes to tokens package

braddialpad:
# fix(tokens): NO-JIRA move themes to tokens package

Obligatory GIF

- [x] Fix

## :book: Description

Moved the new theme logic and configuration to the tokens package instead of the root. Removed type: module from the root.

## :bulb: Context

This change was causing all kinds of issues with imports on ES6 repos, due to `type: "module"`. Needed to be moved to a sub package.

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

Release and test in ui-kits and dx-console


common/utils.js
```diff
@@ -3,18 +3,6 @@ export function getUniqueString (prefix = 'dt') {
...
- * @param string
- * @returns {string}
- */
-export function kebabCaseToPascalCase (string) {
-  return string?.toLowerCase()
-    .split('-')
-    .map(word => word.charAt(0).toUpperCase() + word.slice(1))
-    .join('');
-}
-
```
juliodialpad: I think your linter did this, right? this is being used on icon component IIRC

packages/dialtone-tokens/build-sd-transforms.js
```diff
@@ -16,6 +15,18 @@ register(StyleDictionary);
...
+ * @param string
+ * @returns {string}
+ */
+function kebabCaseToPascalCase (string) {
+  return string?.toLowerCase()
+    .split('-')
+    .map(word => word.charAt(0).toUpperCase() + word.slice(1))
+    .join('');
+}
+
```
juliodialpad: Why moving it to this file? I think that utility it's more general than just for build-sd-transforms, don't you think?

juliodialpad (approved):
Couple minor comments, please address them before merging, otherwise, looks good.

common/utils.js
braddialpad: I get this...

```
import { kebabCaseToPascalCase } from '../../common/utils.js';
         ^^^^^^^^^^^^^^^^^^^^^
SyntaxError: Named export 'kebabCaseToPascalCase' not found. The requested module '../../common/utils.js' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '../../common/utils.js';
const { kebabCaseToPascalCase } = pkg;
```

Do we need to rename this to utils.mjs since package.json is not type: "module" anymore?

braddialpad: Changed this as I mentioned above, seems to have fixed it.

packages/dialtone-tokens/build-sd-transforms.js
braddialpad: Moved back with `.mjs` change mentioned above.

PR #479, opened 2024-08-27
feat(tokens): DLT-2014 add postcss plugin to transform rem to px

braddialpad:
# feat(tokens): DLT-2014 add postcss plugin to transform rem to px

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2014

## :book: Description

Publishing a PostCSS plugin along with dialtone tokens that will convert rem values to px.

## :bulb: Context

When embedding a module using dialtone into an existing page it will always use the rem of the root of the page, even though the module has no control over it. In this case it is ideal to use px over rem.

See thread: https://dialpad.com/app/messages/agxzfnViZXItdm9pY2VyGAsSC1RleHRNZXNzYWdlGICA1b-viJwIDA

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

Test with Dhruba

## :link: Sources

https://postcss.org/docs/writing-a-postcss-plugin

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,6 +69,10 @@
     "./css-default-theme": "./dist/css/dialtone-default-theme.min.css",
     "./tokens/*.css": "./dist/tokens/css/*.css",
     "./tokens/*.less": "./dist/tokens/less/*.less",
+    "./tokens/postcss/rem-to-px": {
+      "import": "./dist/tokens/postcss/rem-to-px.js",
+      "require": "./dist/tokens/postcss/rem-to-px.cjs"
+    },
     "./vue2": {
       "types": "./dist/vue2/types/index.d.ts",
       "import": "./dist/vue2/dialtone-vue.js",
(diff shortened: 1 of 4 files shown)


juliodialpad (commented):
We're starting to output a bunch of stuff from dialtone-tokens, do you think it's ok to keep doing this or should we separate this postcss plugin along with the othe postcss plugin we have: https://github.com/dialpad/postcss-responsive-variations?

Is this plugin meant to be used exclusively on dialtone-tokens? if not, I'll suggest to move it over.

braddialpad:
> We're starting to output a bunch of stuff from dialtone-tokens, do you think it's ok to keep doing this or should we separate this postcss plugin along with the othe postcss plugin we have: https://github.com/dialpad/postcss-responsive-variations?
> 
> Is this plugin meant to be used exclusively on dialtone-tokens? if not, I'll suggest to move it over.

Hmm I don't really think this would be used anywhere other than tokens, it's really only to solve one unique problem a team is having

juliodialpad (changes requested):
Cool, I'm ok with the implementation then, please add some docs related to usage of this, at least how to import and what it does, we don't want to keep this in the dark and forget about it later.

braddialpad:
Added usage docs into the dialtone-tokens readme. I'll send a link to Dhruba when he goes to try this out.

juliodialpad (approved):
Nice! thanks

PR #485, opened 2024-09-04
fix(docs): NO-JIRA broken links in new content guidelines

braddialpad:
# fix(docs): NO-JIRA broken links in new content guidelines

- [x] Documentation

## :book: Description

Fixed a couple links that were broken in the new content guidelines documentation due to not including "and"


juliodialpad (approved):
Thanks for fixing those! just remembered that someone reported those long ago but totally forgot about them 😅

PR #486, opened 2024-09-04
fix(rich-text-editor): DLT-2017 emojis positioning

juliodialpad:
# Fix DtRichTextEditor emojis positioning

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2017

## :book: Description

- Changed nodeInputRule to a custom Input rule to be able to replace the emojis the correct way.
- Added custom backspace behavior to delete the emoji instead of "undo" the inserting.
- Removed the space that was added after the emoji when added through suggestion plugin.
- General cleanup and import reordering.

## :bulb: Context

- Issues reported while inserting emojis.

## :pencil: Checklist

For all PRs:

For all Vue changes:

- Before

https://github.com/user-attachments/assets/0f8e0be7-0183-4f79-a4ed-3fc0491e4d1c

- After

https://github.com/user-attachments/assets/809ed501-44bf-4228-b161-db8520f4382c


packages/dialtone-vue3/components/rich_text_editor/extensions/emoji/emoji.js
```diff
@@ -40,19 +34,16 @@ const shortCodePasteMatch = (text) => {
...
-        char: ':',
-        pluginKey: EmojiPluginKey,
-      },
     };
   },
-  name: 'emoji',
   group: 'inline',
   inline: true,
-  selectable: true,
+  selectable: false,
```
braddialpad, line 45: Shouldn't this be selectable? I notice if I try to highlight when an emoji is at the end of the text it doesn't actually highlight anything, though not sure if making it `selectable: true` will even make a difference...

packages/dialtone-vue3/components/rich_text_editor/extensions/emoji/emoji.js
```diff
@@ -145,10 +121,31 @@ export const Emoji = Node.create({
...
+        if (!empty) { return false; }
+        state.doc.nodesBetween(anchor - 1, anchor, (node, pos) => {
+          if (node.type.name === this.name) {
+            isEmoji = true;
+            tr.insertText('', pos, pos + node.nodeSize);
+            return false;
+          }
+        });
+        return isEmoji;
+      }),
```
braddialpad, line 148: I question whether this is a good way to handle this.. what if they press delete instead of backspace? It seems like this sort of thing you shouldn't have to do manually in TipTap. I think we don't have something configured correctly at a core level with this node view. Do we possibly have to use node-view-content as shown here?

https://tiptap.dev/docs/editor/extensions/custom-extensions/node-views/vue#adding-a-content-editable

packages/dialtone-vue3/components/rich_text_editor/extensions/emoji/emoji.js
juliodialpad: Must be something related to the rich-text-editor besides the emoji, you can actually select it with the keyboard but looks like rich-text-editor is trying to remove some spacing at the end while clicking.

The "selectable" option there is to avoid putting the cursor behind the emojis.

packages/dialtone-vue3/components/rich_text_editor/extensions/emoji/emoji.js
juliodialpad: It works, that function is to avoid "undo" while pressing backspace after entering an emoji.

This is the behavior without that.

packages/dialtone-vue3/components/rich_text_editor/extensions/emoji/emoji.js
juliodialpad: Exactly the same issue with mentions.

juliodialpad: Updating `@tiptap/*` dependencies to latests fixed the issues.

braddialpad: nice! I'm happy it was that easy

packages/dialtone-vue3/components/rich_text_editor/extensions/emoji/emoji.js
braddialpad: oh... strange.. That is not what I thought it was for haha

braddialpad (approved):

PR #487, opened 2024-09-04
fix(tokens): NO-JIRA keep references in css output

braddialpad:
# fix(tokens): NO-JIRA keep references in css output

Obligatory GIF

- [x] Fix

## :book: Description

Enabled the "outputReferences" option in style dictionary

## :bulb: Context

We probably should have always had this enabled, but I got a question from Alex if it was possible to override the default font family, and he found a method of doing it where he had to override tons of different tokens painfully.

If we change our tokens to maintain references he could do this by changing only one single token, plus our CSS file size is smaller and easier to read.

before:

```
--dt-typography-body-md-font-family: -apple-system, BlinkMacSystemFont, "SF Pro", "Segoe UI Adjusted", "Segoe UI", SFMono, "Helvetica Neue", Cantarell, Ubuntu, Roboto, Arial, "Noto Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji"; /* Base default medium text style for main content. */
```

after:

```
--dt-typography-body-md-font-family: var(--dt-font-family-body); /* Base default medium text style for main content. */
```

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

Test for any potential breaking changes.

## :link: Sources

https://styledictionary.com/reference/config/#platform


francisrupert (commented):
Gosh yeah, that would be awesome and a heckuva lot tighter. I would have expected that to be the default in Style Dictionary.

Other than it obviously being halfway broken, I did at least confirm the references, e.g. 

```
// CURRENT
--dt-color-surface-secondary: #F9F9F9;

// UPDATED
--dt-color-surface-secondary: var(--dt-color-black-100);
```

From a CSS perspective, I can't think of a reason we would prefer to embed the value instead of references.

braddialpad:
Yeah for some reason this change completely breaks the storybook.. Looking into it.

juliodialpad:
> Yeah for some reason this change completely breaks the storybook.. Looking into it.

Screenshot 2024-09-05 at 11 29 35 a m

Looks like the fonts are not defined

francisrupert:
Color is a mixed bag actually. So while I did confirm the color references are used, it's not working for those that are then adjusted. e.g. 

```
// CURRENT PRODUCTION
--dt-color-surface-moderate-opaque: hsl(0 0% 0% / .09);

// PR DEPLOY
--dt-color-surface-moderate-opaque: var(--dt-color-black-900);
```

`black-900` is indeed the reference, but it's **modified** in Tokens Studio, i.e...

https://github.com/dialpad/dialtone/blob/e239b22a1c87f3f57402499994c850681a94938e/packages/dialtone-tokens/tokens/semantic/dp/default.json#L264-L277

braddialpad:
> `black-900` is indeed the reference, but it's **modified** in Tokens Studio, i.e...

Hmm I'm guessing style-dictionary probably doesn't understand that and/or sd-transforms doesn't translate it correctly. I think it's possible to ignore specific entries that's probably the best solution for now.

braddialpad:
Alright, all issues should now be fixed. There was a strange issue with style-dictionary not using calc that took me forever to figure out. Ended up just using a simple postcss to fix it but it's a little hacky. Should work without issues for the time being though. Seems to be an ongoing issue with style-dictionary / sd-transforms that is yet unresolved. Left a comment in the code about this and linked to tickets.

braddialpad:
Updated storybook / sd-transforms / style-dictionary as well as I was having some local issues that were fixed in a newer release. Shouldn't be any breaking changes.

braddialpad:
Looks like 1 remaining issue with shadows now that visual tests have run

braddialpad:
Alright, all visual tests passing now. Ready for re-review. @francisrupert if you want to give it lookover

PR #490, opened 2024-09-05
feat(message input): DLT-1916 remove dt-icon from message input

ninamarina:
# Remove dt-icon from message input

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket
[DLT-1916]

image

[DLT-1916]: https://dialpad.atlassian.net/browse/DLT-1916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
+++ b/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
@@ -67,8 +67,7 @@
           @blur="imagePickerFocus = false"
         >
           <template #icon>
-            <dt-icon
-              name="image"
+            <dt-icon-image
               size="300"
             />
           </template>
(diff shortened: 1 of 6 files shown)


packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
```diff
@@ -182,20 +185,21 @@
...
               #icon
             >
-              <dt-icon
-                :name="showSend.icon"
-                size="300"
+              <!-- @slot Slot for send button icon -->
+              <slot
+                name="sendIcon"
+                :icon-size="'300'"
               />
```
braddialpad: This is not exactly the same way as it worked before right? If nothing is passed into the "showSend" prop or "sendIcon" slot it should render the "send" icon as default. You can do this by rendering the sendIcon within the slot. If the user wants to override it, they use the sendIcon slot

ninamarina: You are right, thanks for catching this. Will fix.

ninamarina: I changed it to have default content for the send icon. Also, `showSendIcon` will be true when there is no text since text and icon are exclusive.

juliodialpad (approved):
Looks good, thanks!

PR #492, opened 2024-09-09
feat(utilities): DLT-1763 add gap utility class

juliodialpad:
# Add gap utility class

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1763

## :book: Description

- Added `d-g*`, `d-rg*`, `d-cg*`  utility classes.
- Added flex/gap documentation.
- Updated grid/gap documentation. 
- Added linter rules to warn about `d-g*`, `d-rg*`, `d-cg*` and `d-flg*` usages.
- Migrated deprecated classes on our documentation site.
- Left comments in the code were we need to do a migration in the future with a link pointing to the ticket so we don’t loose it.

## :bulb: Context

- We currently have a grid-gap utility class, but it is confusing that this shows up under grid, because gap is also used for flexbox. Also we should have a util class for just gap rather than grid-gap.
- Found that we previously had an implementation of gap in flexbox using margins through d-flg2.

## :pencil: Checklist

For all PRs:

For all CSS changes:

## :crystal_ball: Next Steps

- Migrate on product and update/remove implementations on dialtone-css.


ninamarina:
Looks good! Should we also remove `d-flg{n}` from the docs or will that be done in another ticket?
image

juliodialpad:
> Looks good! Should we also remove `d-flg{n}` from the docs or will that be done in another ticket? image

You're right, I need to update that docs to use `d-g` too

ninamarina (approved):

apps/dialtone-documentation/docs/utilities/flex/gap.md
```diff
@@ -0,0 +1,96 @@
...
+</code-well-header>
+
+```html
+<div class="d-fl-col4 d-cg16">
+  <div>1</div>
+  <div>2</div>
+  <div>3</div>
+  <div>4</div>
+</div>
+```
```
braddialpad, line 70: Aren't row gap and column gap only relevant to grids? So they should be under the `grid/gap` document rather than `flex/gap` document?

juliodialpad: No, it is possible to use it in both Flex and Grid layouts.

https://developer.mozilla.org/en-US/docs/Web/CSS/row-gap#flex_layout
https://developer.mozilla.org/en-US/docs/Web/CSS/column-gap#flex_layout

braddialpad (approved):

francisrupert (approved):
Awesome. Worth considering a migration script to push the deprecation of `d-gg##` along? Although, this does seem simple enough to just search-and-replace directly in product.

francisrupert:
`d-cg{#}` and `d-rg{#}` too?

francisrupert:
> `d-cg{#}` and `d-rg{#}` too?

Oh, maybe not. `cg` and `rg` are very relevant if you want row and column gap to differ. nm I guess.

juliodialpad:
> Awesome. Worth considering a migration script to push the deprecation of `d-gg##` along? Although, this does seem simple enough to just search-and-replace directly in product.

It's easier to just do the migration instead of a migration script, based on my experience migrating through our repo it was a straightforward replacement with no changes visually.

PR #494, opened 2024-09-10
fix(lint): NO-JIRA fix linter rule for text styles

ninamarina:
# Fix recommend-typography-style rule false positives

In this rule we were checking for `d-fw` and `d-ff` prefixes, which are also used for flex-wrap and flex-flow properties.
I improved the rule to target the specific properties to avoid getting a warning in those cases.

image

### DIFF
--- a/packages/eslint-plugin-dialtone/lib/rules/recommend-typography-style.js
+++ b/packages/eslint-plugin-dialtone/lib/rules/recommend-typography-style.js
@@ -38,7 +38,6 @@ module.exports = {
             'd-fw-medium',
             'd-fw-semibold',
             'd-fw-bold',
-            'd-fw-unset',
             'd-lh',
             'd-ff-custom',
             'd-ff-sans',
--- a/packages/eslint-plugin-dialtone/lib/rules/recommend-typography-style.js
+++ b/packages/eslint-plugin-dialtone/lib/rules/recommend-typography-style.js
@@ -32,7 +32,20 @@ module.exports = {
       VAttribute(node) {
         if (node.key.name === 'class') {
           const classes = node.value.value.split(' ');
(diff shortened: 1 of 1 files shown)


packages/eslint-plugin-dialtone/lib/rules/recommend-typography-style.js
```diff
@@ -32,7 +32,20 @@ module.exports = {
...
         if (node.key.name === 'class') {
           const classes = node.value.value.split(' ');
-          const typographyClasses = ['d-fs', 'd-fw', 'd-lh', 'd-ff'];
+          const typographyClasses = [
+            'd-fs',
+            'd-fw-normal',
+            'd-fw-medium',
+            'd-fw-semibold',
+            'd-fw-bold',
+            'd-fw-unset',
```
juliodialpad: `d-fw-unset` is actually a flex wrap utility class https://dialtone.dialpad.com/utilities/flex/direction-wrap-flow.html#flex-wrap

```suggestion
```

juliodialpad (approved):
That's something we should fix in the near future, it's not ok that more that one utility class has the same prefix, can lead to confusions like the one you just had with `d-fw-unset`

Just apply the suggested change before merging, thanks!

packages/eslint-plugin-dialtone/lib/rules/recommend-typography-style.js
juliodialpad: Actually just found that we have it duplicated.

image
Screenshot 2024-09-10 at 6 24 29 p m

The thing is we don't have font weight d-fw-unset documented https://dialtone.dialpad.com/utilities/typography/font-weight.html#classes

ninamarina: Oh, I didn't realize this, thanks for catching it!

PR #503, opened 2024-09-19
feat(combobox-multi-select): DLT-2066 chip truncation

braddialpad:
# feat(combobox-multi-select): DLT-2066 chip truncation

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2066

## :book: Description

Added new prop chipMaxWidth that will truncate text if the chip is greater than that width.

## :bulb: Context

Some users are having issues with multiselect in smaller width interfaces.

## :pencil: Checklist

For all PRs:

Screenshot 2024-09-19 at 11 32 52 AM


packages/dialtone-vue3/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
```diff
@@ -1,3 +1,4 @@
+<!-- eslint-disable vue/no-static-inline-styles -->
```
juliodialpad, line 1: why is this line needed? The same applies for Vue 2 version

juliodialpad (approved):
This works pretty well, along with the change I made recently, should be enough for the users to avoid overflowing issues with this component.

Not related to this change but the focus ring looks pretty strange.

Screenshot 2024-09-19 at 1 49 03 p m

packages/dialtone-vue3/recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
braddialpad: because we're setting `:style` directly on the component which is generally not something we want to do, but in this case it makes sense because the style is passed in via a prop and we have to set it dynamically.

braddialpad:
> Not related to this change but the focus ring looks pretty strange.

hmm this IS actually related to this change. it's fine on https://dialtone.dialpad.com/vue

braddialpad:
Problem fixed, I removed the overflow: hidden from the truncation class. Doesn't seem to hurt the truncation at all.

PR #504, opened 2024-09-20
fix(tokens): DLT-2053 android tokens color value

ninamarina:
# Fix Android color tokens

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket
[DLT-2053]

## :book: Description
Fixes the color tokens output for Android. As an example you can check out `dtColorForegroundMuted`, that before had the value `Color(0xff000000)` and now has the correct `Color(0x80222222)`.

## :bulb: Context
It started working wrong when removing token-transformer.

### Before

image

### Now

image

[DLT-2053]: https://dialpad.atlassian.net/browse/DLT-2053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dialtone-tokens/dialtone-transforms.js
```diff
@@ -85,6 +85,16 @@ export function registerDialtoneTransforms (styleDictionary) {
       return ['color'].includes(token.type);
     },
     transform: (token) => {
+      if (token.$extensions?.['studio.tokens']?.modify?.type === 'alpha') {
+        const alpha = parseFloat(token.$extensions['studio.tokens'].modify.value);
+        const alphaHex = Math.round(alpha * 255).toString(16).padStart(2, '0');
```
braddialpad, line 90: I'm pretty sure you could just simply tack alphaHex on to the end of token.value and everything else would just work right?

Example:

```
transform: (token) => {
  let alphaHex = '';
  if (token.$extensions?.['studio.tokens']?.modify?.type === 'alpha') {
    const alpha = parseFloat(token.$extensions['studio.tokens'].modify.value);
    alphaHex = Math.round(alpha * 255).toString(16).padStart(2, '0');
  }
  const hex8 = Color(token.value + alphaHex).toHex8();
  return `Color(0x${hex8.slice(6) + hex8.slice(0, 6)})`;
},
```

braddialpad (commented):
Looks good functionally. Can probably simplify the code a little.

packages/dialtone-tokens/dialtone-transforms.js
braddialpad: If token.value contains 8 characters already, might need to replace the final two characters instead of append.

ninamarina: That won't work because for the tokens with `$extensions` the value comes in Flutter format already, i.e `Color(0xffabcd)`, so you can't append the alphaHex.

braddialpad: Alright, we can merge this then.

braddialpad (approved):

PR #505, opened 2024-09-24
build: DLT-2059 fix NX performance issues

juliodialpad:
# Fix NX performance issues

Obligatory GIF

- [x] Documentation
- [x] Build system
- [x] CI

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2059

## :book: Description

- Updated NX configuration to be more specific to each project.
- Updated NX commands through all the repo
  -  NX target project -> NX run project:target, this is more reliable as it avoids confusions.
- Updated documentation
- Updated CI with the new commands
- Updated VSCode tasks.
- Removed unneeded npm scripts from package.json files
- Moved shared devDependencies to root's package.json, this makes it easier to update versions.
  - Only DevDependencies that are specific to a package should be included on projects package.json, if shared, add to workspace package.json instead.
  - package dependencies should be always on their corresponding package.
- Moved scripts that have dependencies to package's project.json.
  - This prevents people from running `pnpm run build` instead of `nx run build`.
  - Let's us add inputs/outputs to help NX to know exactly what needs to be cached.
  - We can add specific dependencies or commands that can be cached individually or run in parallel to improve performance.
- General cleanup

## :bulb: Context

- Performance issues while building the dialtone-documentation site locally were reported.
- While investigating, found a bunch of other issues that might be affecting NX cache and performance so fixed all issues found.

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

- Clean up unneeded dependencies, detected that we have a few dependencies that are not being used (forgot to remove them maybe?)
- Fix linter errors/warnings, We have a bunch of linter errors/warnings that we've been ignoring for long time.
- Run linters on before commit hook.


.github/CONTRIBUTING.md
```diff
@@ -102,14 +102,41 @@ Any new components or updates to existing components require the following:
   - Read by a screen reader.
   - Minimum contrast ratio.
 - Changes must be made for Vue 2 as well as Vue 3, `dialtone-vue2` and `dialtone-vue3` package folders respectively
+
 - Unit tests are passing locally.
-  - `nx test dialtone-vue2` or `nx test dialtone-vue3`
+  - Test all projects
+    - `nx run dialtone:test:all`
```
braddialpad, line 108: doesn't it also work if you just do `nx run test` from the root directory? I thought the idea was that the commands all have the same name throughout all the packages.

braddialpad (commented):
One quick question, got interview now so will review more later.

.github/CONTRIBUTING.md
juliodialpad: All the test commands have the same name, you can run them with `nx run-many -t test` the `nx run dialtone:test:all` is just like a shortcut `dialtone` is the name of the root package.

apps/dialtone-documentation/project.json
```diff
@@ -1,13 +1,43 @@
...
     },
     "start": {
-      "executor": "nx:run-script",
-      "options": { "script": "start" }
+      "executor": "nx:run-commands",
+      "dependsOn": [
+        "dialtone-icons:build",
+        "dialtone-tokens:build",
+        "dialtone-vue3:build"
+      ],
```
braddialpad: So currently if I run the dialtone-documentation local server and then make a small change in a vue 3 component, that change will not be reflected live by the server. It will only show up if I manually do an nx build from `packages/dialtone-vue3`. Is it supposed to be smart enough to know to when to rebuild dependent packages?

package.json
```diff
@@ -157,28 +171,50 @@
...
+    "nx": "^19.8.0",
     "playwright": "^1.42.1",
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1",
     "semantic-release": "^21.0.6",
     "semantic-release-npm-github-publish": "^1.5.4",
     "semantic-release-plus": "^20.0.0",
-    "stylelint": "15.11.0",
+    "source-map-support": "^0.5.21",
+    "storybook": "7.6.20",
```
braddialpad, line 197: It's still kind of unclear to me when a package should be put in the root package.json vs when it should be put in the packages/package-folder/package.json. Is it just whenever a dependency is used in more than a single package?

packages/dialtone-css/.github/CONTRIBUTING.md
```diff
@@ -87,7 +86,7 @@ Note that we use lesshint only for the utilities folder and stylelint for everyt
 Here are some important directories to know within the Dialtone repository
 
 - `lib/build`: All source code and assets for the Dialtone library.
-- `lib/dist`: The compiled bundle will be output here upon `nx build dialtone-css`.
+- `lib/dist`: The compiled bundle will be output here upon `nx run dialtone-css:build`.
```
braddialpad, line 89: Should it be `pnpm nx` in case they don't have nx installed locally? I saw you used that in some other examples

braddialpad (commented):
Really nice work, definitely much faster. Thanks for going the extra mile on this and reworking so much of the system!

For some reason even after the first build mine hangs for so long at this point.

```
[14:38:56] Finished 'libStylesDev' after 1.18 s
[14:38:56] Starting 'watchFiles'...
[14:38:56] Finished 'watchFiles' after 4.08 ms
[14:38:56] Finished 'watch' after 1.32 s
✔ Initializing and preparing data - done in 26.10s
```

That 26.10s... so much slower than whatever you said your time was, 9 seconds total I think? Anyways it's a minor issue now and probably something wrong with my local system.

braddialpad:
For some reason accessibility tests ran for 6 hours.. weird thing is the log shows they all passed too.

juliodialpad:
> Really nice work, definitely much faster. Thanks for going the extra mile on this and reworking so much of the system!
> 
> For some reason even after the first build mine hangs for so long at this point.
> 
> ```
> [14:38:56] Finished 'libStylesDev' after 1.18 s
> [14:38:56] Starting 'watchFiles'...
> [14:38:56] Finished 'watchFiles' after 4.08 ms
> [14:38:56] Finished 'watch' after 1.32 s
> ✔ Initializing and preparing data - done in 26.10s
> ```
> 
> That 26.10s... so much slower than whatever you said your time was, 9 seconds total I think? Anyways it's a minor issue now and probably something wrong with my local system.

Please try running `nx reset` that'd reset NX Daemon and clear cache, maybe something's messed up

braddialpad:
> Please try running `nx reset` that'd reset NX Daemon and clear cache, maybe something's messed up

Didn't seem to make a difference. NX seems to be caching / working perfectly, it's the building of vuepress itself that's taking a while.

apps/dialtone-documentation/project.json
braddialpad: Can probably do this with the vuepress "onWatched" hook https://v2.vuepress.vuejs.org/reference/plugin-api.html#onwatched

juliodialpad: Found a way to make this, NX has a "watch" function to watch for changes on other projects and run a command when that happens. It's working correctly I think, will push the change in a moment.

package.json
juliodialpad: Yes, note that this rule only applies for DevDependencies. 

Whenever we have a dependency that we want to share among our packages, put it on the workspace root, pnpm will make the magic so the dependency is available to the rest of the packages and if we need to updated that dependency, we just need to update it in one place.

When we need e.g. a specific Vue version, or a dependency that's only used on a single project put it on the project's specific package.json.

packages/dialtone-css/.github/CONTRIBUTING.md
juliodialpad: > Should it be `pnpm nx` in case they don't have nx installed locally? I saw you used that in some other examples

Used `pnpm nx` only on `tasks.json` which is for VSCode, did that because I had problems previously when VSCode didn't found nx so commands didn't run properly, hopefully that fixes that issue. But for the rest, I'll prefer to run `nx run ...`

juliodialpad:
> For some reason accessibility tests ran for 6 hours.. weird thing is the log shows they all passed too.

I think that happened because the `pnpm exec http-server` command is not actually ending, need to find a way to end the command after the testing is done. Working on it.

apps/dialtone-documentation/project.json
braddialpad: sweet, that sounds like the right way to do it.

juliodialpad:
> > Please try running `nx reset` that'd reset NX Daemon and clear cache, maybe something's messed up
> 
> Didn't seem to make a difference. NX seems to be caching / working perfectly, it's the building of vuepress itself that's taking a while.

Found an issue with `vue-tsc`  not caching correctly, might help with performance.

juliodialpad:
Running visual tests to make sure they're working correctly

braddialpad (approved):
Seems to work nicely now, great job!

PR #506, opened 2024-09-24
fix(emoji-picker): NO-JIRA multiple fixs

iropolo:
# fix(emoji-picker): no-jira multiple fixs

- Removed all related to `is-scrolling` function, if I remember correctly that was introduced because previously we had a scroll animation when the user navigate with the Tab section. Later the animation was removed so I think this is no longer necessary.
- There was an issue were we pass a `string` to our `scrollToTab` from the `emoji_tabset.vue` component. This function need a `number` type to work properly.
- The offset was adjust from `20` to `15` hopefully this will fit better.
- Now the `focusLastEmoji` method works more precisely.

Reported issues to solve:
https://dialpad.atlassian.net/browse/DP-111895
https://dialpad.atlassian.net/browse/DLT-2063

### DIFF
--- a/packages/dialtone-vue2/components/emoji_picker/modules/emoji_selector.vue
+++ b/packages/dialtone-vue2/components/emoji_picker/modules/emoji_selector.vue
@@ -334,26 +334,7 @@ export default {
 
       vm.$nextTick(function () {
         const container = vm.$refs.listRef;
-        const offsetTop = tabIndex === '1' ? 0 : tabElement.offsetTop - 20;
-
-        let isScrolling = true;
-        let prevScrollTop = container.scrollTop;
-        vm.$emit('is-scrolling', true);
-
-        /* eslint-disable-next-line complexity */
-        container.addEventListener('scroll', function () {
-          if (isScrolling) {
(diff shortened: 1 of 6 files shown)


packages/dialtone-vue2/components/emoji_picker/modules/emoji_selector.vue
```diff
@@ -334,26 +334,7 @@ export default {
...
-              (prevScrollTop < scrollTop && scrollTop >= offsetTop) ||
-            (prevScrollTop > scrollTop && scrollTop <= offsetTop)
-            ) {
-              isScrolling = false;
-              vm.$emit('is-scrolling', false);
-            }
-            prevScrollTop = scrollTop;
-          }
-        });
+        const offsetTop = tabIndex === 1 ? 0 : tabElement.offsetTop - 15;
```
iropolo, line 337: Main fix

braddialpad (approved):
Looks good!

juliodialpad (approved):
Working well, can't longer reproduce the issue, thanks!

PR #509, opened 2024-09-26
feat: DLT-1915 add stylelint warning when using px or rem instead of tokens

ninamarina:
# feat: DLT-1915 add stylelint warning when using px or rem instead of tokens

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket
[DLT-1915]

## :book: Description
Adds a stylelint rule to detect the use of px or rem instead of tokens.
In the case of px there are two warnings, one suggesting the use of tokens, and another suggesting the use of rem if that is not posible.

## :bulb: Context
We observed in the Adoption Audit that in some cases px values where used for spacing, sizing, typography instead of Dialtone tokens.

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps
Enable the rule in Ubervoice once its released.

image

## :link: Sources

[DLT-1915]: https://dialpad.atlassian.net/browse/DLT-1915?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/stylelint-plugin-dialtone/lib/rules/use-dialtone-tokens.js
+++ b/packages/stylelint-plugin-dialtone/lib/rules/use-dialtone-tokens.js
@@ -9,9 +9,15 @@ const ruleName = '@dialpad/stylelint-plugin-dialtone/use-dialtone-tokens';
 
 const messages = ruleMessages(ruleName, {
   useDialtoneTokens: (value) => `Use a Dialtone token instead of "${value}".
-    A list of available tokens can be found here: https://dialtone.dialpad.com/tokens.`,
-  useRemInsteadOfPx: (value) => `If it's not possible to use a Dialtone token, use rem instead of px for "${value}".`,
+    A list of available tokens can be found here: https://dialtone.dialpad.com/tokens.
+    Consult with design on which token should be used and only with their approval ignore this rule.`,
+  useRemInsteadOfPx: (value) => `If it's not possible to use a Dialtone token, use ${pxToRem(value)} instead of ${value}.`,
 });
+
+const pxToRem = (value) => {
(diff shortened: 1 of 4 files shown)


packages/stylelint-plugin-dialtone/lib/rules/use-dialtone-tokens.js
```diff
@@ -0,0 +1,63 @@
...
+  createPlugin,
+  utils: { report, ruleMessages, validateOptions },
+} = stylelint;
+
+const ruleName = '@dialpad/stylelint-plugin-dialtone/use-dialtone-tokens';
+
+const messages = ruleMessages(ruleName, {
+  useDialtoneTokens: (value) => `Use a Dialtone token instead of "${value}".
+    A list of available tokens can be found here: https://dialtone.dialpad.com/tokens.`,
+  useRemInsteadOfPx: (value) => `If it's not possible to use a Dialtone token, use rem instead of px for "${value}".`,
```
braddialpad: Can we suggest the rem value to replace it with?

ex
`If it's not possible to use a Dialtone token, use ${value/10.0}rem instead of ${value}px.`

braddialpad: Let's say to consult with design on which token should be used, and only if the designer approves it should we use rem instead of a token. In the case of this decision, they can then set a lint ignore on this line.

braddialpad (approved):
Thanks!

PR #516, opened 2024-10-02
fix(tokens): NO-JIRA incorrect hsl for reference tokens

braddialpad:
# fix(tokens): NO-JIRA incorrect hsl for reference tokens

Obligatory GIF

- [x] Fix

## :book: Description

Fixing a problem Julio noticed where referenced tokens were not correctly generating HSL colors because a string like `var(--dt-color-foreground-critical-strong-inverted);` obviously doesn't contain any color information. Instead made the hsl reference its token instead of trying to generate new color data where it cannot.

Not generating HSL for gradient because it's very abnormal and I don't even think possible

Haven't really tested this much yet, will do more testing tomorrow.

### DIFF
--- a/packages/dialtone-tokens/postcss/constants.cjs
+++ b/packages/dialtone-tokens/postcss/constants.cjs
@@ -9,6 +9,7 @@
  */
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


juliodialpad (approved):
Nice! This seems to work as expected, thanks @braddialpad

juliodialpad:
I'll merge this so I can test it on my branch

PR #519, opened 2024-10-03
docs: DLT-2002 fix remaining modal scrollbar issues

braddialpad:
# docs: DLT-2002 fix remaining modal scrollbar issues

Obligatory GIF

- [x] Documentation

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2002

## :book: Description

- Fixed modal example component to use vue component instead of pure HTML/CSS
- Fixed modal component not removing the scrollbar on open
- Fixed algolia search not removing the scrollbar on open
- Moved rootScrolling functions to common so any package can use them.
- Added functionality to rootScrolling functions to support shadowDom

## :bulb: Context

Modal components were no longer correctly hiding the scrollbar since we implemented the dt-scrollbar

## :pencil: Checklist

For all PRs:

For all Vue changes:

## :link: Sources

https://kingsora.github.io/OverlayScrollbars/

### DIFF
--- a/apps/dialtone-documentation/docs/.vuepress/exampleComponents/ExampleModal.vue
+++ b/apps/dialtone-documentation/docs/.vuepress/exampleComponents/ExampleModal.vue
@@ -7,7 +7,7 @@
     :banner-kind="bannerKind"
     :fixed-header-footer="fixedHeaderFooter"
     :size="size"
-    :copy="modalDescription"
+    :copy="copy"
     @update:show="isOpen = $event"
   >
     <template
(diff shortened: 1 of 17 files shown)


apps/dialtone-documentation/docs/.vuepress/exampleComponents/ExampleModal.vue
```diff
@@ -1,224 +1,84 @@
...
-    @keydown.esc="closeModal"
+  <dt-modal
+    title="Example title"
+    :close-button-props="{ ariaLabel: 'Close modal' }"
+    :show="isOpen"
+    :banner-title="bannerTitle"
+    :banner-kind="bannerKind"
+    :fixed-header-footer="fixedHeaderFooter"
+    :size="size"
+    :copy="modalDescription"
```
ninamarina: Previously, in the case where "fixedHeaderFooter" prop was true, we had the copy repeated to show that the header and footer are fixed and only the content is scrollable. I think we should keep that.

braddialpad: ah yeah good catch, I'll add that back.

ninamarina:
I can still scroll while the algolia search modal is open:

algolia-scroll

apps/dialtone-documentation/docs/.vuepress/exampleComponents/ExampleModal.vue
braddialpad: This one is fixed

braddialpad:
> I can still scroll while the algolia search modal is open:

Well... This was actually pretty hard to make work. I could trigger the scrollbar disable when clicking the search button easily, but because the dialog is completely controlled by algolia and they don't appear to trigger any capturable close event it was quite difficult to enable the scrollbar again when closed.

I ended up using what I think is a pretty clever hack using focusout events on that dialog window to determine when it is closed (relatedEvent === null). Have a look, let me know what you think.

ninamarina:
> > I can still scroll while the algolia search modal is open:
> 
> Well... This was actually pretty hard to make work. I could trigger the scrollbar disable when clicking the search button easily, but because the dialog is completely controlled by algolia and they don't appear to trigger any capturable close event it was quite difficult to enable the scrollbar again when closed.
> 
> I ended up using what I think is a pretty clever hack using focusout events on that dialog window to determine when it is closed (relatedEvent === null). Have a look, let me know what you think.

It works much better now, the only detail is that when changing tabs or focusing another app and going back to the site the scroll would now be enabled, I don't think it's serious anyway since it's an edge case.
The other problem that I found is that the sidebar is cropped so it's not visible if the search is open in the bottom of the page:

image

braddialpad:
> he only detail is that when changing tabs or focusing another app and going back to the site the scroll would now be enabled, I don't think it's serious anyway since it's an edge case.
> The other problem that I found is that the sidebar is cropped so it's not visible if the search is open in the bottom of the page:

Hmm. yeah this is a bit janky I'll see if I can improve it.

braddialpad:
Alright, changed it to use a mutation observer instead. Should work much more reliably now.

ninamarina (approved):
Nice, works as expected now

PR #520, opened 2024-10-03
fix: DLT-1548 fix opacity classes selectors

juliodialpad:
# Fix: :hover, :focus, :focus-visible selectors on semantic colors

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1548

## :book: Description

- Fix semantic colors generation so they can handle opacity changes through selectors.
- Fixed HSL color extraction process, found an issue were all the HSLA colors were always 100% opacity.
- Added `--dt-color-${name}-a` to generated individual HSL color variables
- Migrated a deprecated class `d-fc-error` -> `d-fc-critical`
- Updated Font colors documentation, it was showing pretty outdated colors.
- Removed found references to `d:d-*` prefix which was related to some old dark-theme implementation that no longer works.
- General cleanup

## :bulb: Context

- It wasn't possible to change the opacity of semantic colors
- Semantic colors weren't including 

## :pencil: Checklist

For all PRs:

For all CSS changes:

## :crystal_ball: Next Steps

- Migrate `d-fc-error` -> `d-fc-critical` on Dialpad and Dialpad meetings

## :link: Sources

https://colorjs.io/


apps/dialtone-documentation/docs/utilities/typography/font-color.md
```diff
@@ -32,7 +32,7 @@ Use `d-fc-{color}` to change an element's text color.
 
 ## Changing opacity
 
-Use `d-fco{n}` to change an element's text color opacity. You can also change font color opacity on `:hover`, `:focus`, `:focus-visible`, or in dark mode by using the respective `h:d-fco{n}`, `f:d-fco{n}`, `fv:d-fco{n}`, or `d:d-fco{n}` prefixes.
+Use `d-fco{n}` to change an element's text color opacity. You can also change font color opacity on `:hover`, `:focus`, `:focus-visible`, or in dark mode by using the respective `h:d-fco{n}`, `f:d-fco{n}`, `fv:d-fco{n}` prefixes.
```
braddialpad: Remove  "or in dark mode"

apps/dialtone-documentation/docs/utilities/typography/text-opacity.md
```diff
@@ -6,7 +6,7 @@ The ability to control an element's color is <em>only</em> provided as a class.
 
 ## Usage
 
-Use `d-fco{n}` to change a font-color's opacity. You can also change font-color's opacity on `:hover`, `:focus`, `:focus-visible`, or in dark mode by using the respective `h:d-fco{n}`, `f:d-fco{n}`, `fv:d-fco{n}`, or `d:d-fco{n}` prefixes.
+Use `d-fco{n}` to change a font-color's opacity. You can also change font-color's opacity on `:hover`, `:focus`, `:focus-visible`, or in dark mode by using the respective `h:d-fco{n}`, `f:d-fco{n}`, `fv:d-fco{n}` prefixes.
```
braddialpad: Remove "or in dark mode"

packages/dialtone-css/lib/build/less/utilities/colors.less
```diff
@@ -13,110 +13,17 @@
...
-.d-fc-success-strong { color: var(--dt-color-foreground-success-strong) !important; }
-.d-fc-success-strong-inverted { color: var(--dt-color-foreground-success-strong-inverted) !important; }
-.d-fc-warning { color: var(--dt-color-foreground-warning) !important; }
-.d-fc-warning-inverted { color: var(--dt-color-foreground-warning-inverted) !important; }
-.d-fc-error { color: var(--dt-color-foreground-critical) !important; }
-.d-fc-error-strong { color: var(--dt-color-foreground-critical-strong) !important; }
-.d-fc-critical { color: var(--dt-color-foreground-critical) !important; }
-.d-fc-critical-inverted { color: var(--dt-color-foreground-critical-inverted) !important; }
-.d-fc-critical-strong { color: var(--dt-color-foreground-critical-strong) !important; }
-.d-fc-critical-strong-inverted { color: var(--dt-color-foreground-critical-strong-inverted) !important; }
```
braddialpad, line 39: Should we have just kept them here instead of generating with postcss?

example:
```
.d-fc-primary { 
  color: hsl(var(--dt-color-foreground-primary-h) var(--dt-color-foreground-primary-s) var(--dt-color-foreground-primary-l) / var(--fco, var(--dt-color-foreground-primary-a))) !important; 
}
```

On the one hand, doing them in postcss is more automated, on the the other hand, it is much harder to understand what's going on... Generally we use the pattern that all util classes are defined in the less files and then postcss is used to generate hover/focus variants. It's kind of weird to not do that just in this one case for color. What do you think?

francisrupert:
The semantic CSS Utilities don't seem to work, the `--fco`/`--bgo`/`--bco` are not defined. Though it is working for base colors.

image

francisrupert:
This has all been purely about CSS Utilities, though the same must work for the CSS Variables too. I suppose the below might have some non-trivial Style Dictionary adjustments.

I [updated my CodePen](https://codepen.io/francisrupert/pen/rNEooEo?editors=1100) to reflect enabling this feature in both CSS Utilities and CSS Variables. 

Bottom line for CSS Variables: instead of...
```
--dt-color-black-900: #000000;
...
--dt-color-foreground-primary: var(--dt-color-black-900);
```

...it would be...

```
...
--dt-color-black-900: hsl(var(--dt-color-black-900-hsl) / var(--fco));
...
--dt-color-foreground-primary: hsl(var(--dt-color-foreground-primary-hsl) / var(--fco));
```

The above isn't comprehensive, though I *think* that's correct. Refer to my [CodePen](https://codepen.io/francisrupert/pen/rNEooEo?editors=1100) for the general pattern I think we're aiming for. The CodePen doesn't have Base colors, but same pattern probably applies.

francisrupert:
Hmm, just had a thought... we're supposed to be retaining the references, so maybe this is more accurate...

```css
  --dt-color-foreground-primary-h: var(--dt-color-black-900-h); // retain reference
  --dt-color-foreground-primary-s: var(--dt-color-black-900-s); // retain reference
  --dt-color-foreground-primary-l: var(--dt-color-black-900-l); // retain reference
  --dt-color-foreground-primary-hsl: var(--dt-color-foreground-primary-h) var(--dt-color-foreground-primary-s) var(--dt-color-foreground-primary-l);
  --dt-color-foreground-primary-hsla: hsl(var(--dt-color-foreground-primary-hsl) / var(--fco));
  --dt-color-foreground-primary: hsl(var(--dt-color-foreground-primary-hsl) / var(--fco));
```
I'm not making your life easy am I.

packages/dialtone-css/lib/build/less/utilities/colors.less
juliodialpad: In fact, we were already generating most color utility classes on PostCSS, we were missing the semantic `font-color`, `border-color` and `background-color` utility classes.

We can move all of them to either LESS or PostCSS but I completely agree, having them on PostCSS is more automated but pretty obscure.

Having them in LESS is way easier to find and fix something that is broken.

We're already generating all this utility classes.

Screenshot 2024-10-03 at 10 01 14 p m

Certainly we don't want to come back to what we had years ago, maybe we just need to add some documentation to make it easier to find what we're generating on PostCSS and what on LESS.

juliodialpad:
> The semantic CSS Utilities don't seem to work, the `--fco`/`--bgo`/`--bco` are not defined. Though it is working for base colors.
> 
> image

Forgot to add the alpha value to tokens that were already referencing a variable.

juliodialpad:
> This has all been purely about CSS Utilities, though the same must work for the CSS Variables too. I suppose the below might have some non-trivial Style Dictionary adjustments.
> 
> I [updated my CodePen](https://codepen.io/francisrupert/pen/rNEooEo?editors=1100) to reflect enabling this feature in both CSS Utilities and CSS Variables.
> 
> Bottom line for CSS Variables: instead of...
> 
> ```
> --dt-color-black-900: #000000;
> ...
> --dt-color-foreground-primary: var(--dt-color-black-900);
> ```
> 
> ...it would be...
> 
> ```
> ...
> --dt-color-black-900: hsl(var(--dt-color-black-900-hsl) / var(--fco));
> ...
> --dt-color-foreground-primary: hsl(var(--dt-color-foreground-primary-hsl) / var(--fco));
> ```
> 
> The above isn't comprehensive, though I _think_ that's correct. Refer to my [CodePen](https://codepen.io/francisrupert/pen/rNEooEo?editors=1100) for the general pattern I think we're aiming for. The CodePen doesn't have Base colors, but same pattern probably applies.

I think what you want to achieve is easily achievable through `--dt-color-black-900-hsla` CSS Variable and overriding the `--alpha`

Screenshot 2024-10-03 at 10 12 59 p m

You shouldn't really use `--fco` CSS var on a base token because that token is not specific for `font color` I think `--alpha` is more generic, and that's what we're currently using.

juliodialpad:
> Hmm, just had a thought... we're supposed to be retaining the references, so maybe this is more accurate...
> 
> ```css
>   --dt-color-foreground-primary-h: var(--dt-color-black-900-h); // retain reference
>   --dt-color-foreground-primary-s: var(--dt-color-black-900-s); // retain reference
>   --dt-color-foreground-primary-l: var(--dt-color-black-900-l); // retain reference
>   --dt-color-foreground-primary-hsl: var(--dt-color-foreground-primary-h) var(--dt-color-foreground-primary-s) var(--dt-color-foreground-primary-l);
>   --dt-color-foreground-primary-hsla: hsl(var(--dt-color-foreground-primary-hsl) / var(--fco));
>   --dt-color-foreground-primary: hsl(var(--dt-color-foreground-primary-hsl) / var(--fco));
> ```
> 
> I'm not making your life easy am I.

References are retained correctly, the only change you need to make for this tokens to be able to change "opacity" is setting `--alpha` and using `*-hsla` variable.

Screenshot 2024-10-03 at 10 18 53 p m

braddialpad:
> You shouldn't really use `--fco` CSS var on a base token because that token is not specific for `font color` I think `--alpha` is more generic, and that's what we're currently using.

Does this mean `--fco/--bgo/--bco` aren't used anymore there's just one generic `--alpha`?

braddialpad:
> Does this mean `--fco/--bgo/--bco` aren't used anymore there's just one generic `--alpha`?

I see there is `--fco` etc still in the code, so I guess `--alpha` is used for tokens while `--fco` is used for util classes.

braddialpad (approved):
These changes seem reasonable to me.

francisrupert:
> References are retained correctly, the only change you need to make for this tokens to be able to change "opacity" is setting `--alpha` and using `*-hsla` variable.

The case I'm thinking of is using **just** the `d-fco##`/`d-bgo##`/`d-bco##` utilities on something that may or may not be using a corresponding `.d-fc-{color}`/`.d-bgc-{color}`/`.d-fc-{color}`/ CSS Utility. For example...

```html
<div class="thing d-bgc50">
  asdf
</div>
```

The expectation there is that the background color would be 50% opacity of whatever background color was applied in `.thing`'s CSS (e.g. `bgc: var(--dt-color-surface-contrast`).

juliodialpad:
> > References are retained correctly, the only change you need to make for this tokens to be able to change "opacity" is setting `--alpha` and using `*-hsla` variable.
> 
> The case I'm thinking of is using **just** the `d-fco##`/`d-bgo##`/`d-bco##` utilities on something that may or may not be using a corresponding `.d-fc-{color}`/`.d-bgc-{color}`/`.d-fc-{color}`/ CSS Utility. For example...
> 
> ```
> <div class="thing d-bgc50">
>   asdf
> </div>
> ```
> 
> The expectation there is that the background color would be 50% opacity of whatever background color was applied in `.thing`'s CSS (e.g. `bgc: var(--dt-color-surface-contrast`).

In that case we could set `--alpha` value to. `.d-bgc50` utility class. BUT that's a pretty weird case IMO I know there's someone out there that might end up doing this, but that's not the right way to do it.

They should do:

```html
<div class="thing">
  asdf
</div>
```

```css
.thing {
  --alpha: 50%;
  background-color: var(--dt-color-surface-contrast-hsla);
}
```

francisrupert:
> They should do:
> 
> ```
> <div class="thing">
>   asdf
> </div>
> ```
> 
> ```css
> .thing {
>   --alpha: 50%;
>   background-color: var(--dt-color-surface-contrast-hsla);
> }
> ```

That works for when one is in a position to author CSS for `.thing`. I'm thinking of a situation I can see occurring when someone wants to use a CSS Utility to change the opacity of a style property that is already declared in CSS, and they're not in a position to author CSS. Example...

```html
<dt-notice kind="info" class="d-bgo50">
  ...
</dt-notice>
```

That background-color is already set in the CSS (light blue). Without authoring custom CSS, they just want to tweak its background opacity with a CSS Utility `class="d-bgo50"`.

That's how it currently works on this branch's deploy preview for...
```html
<div class="d-bgc-secondary d-bgo50">
  ...
</div>
```

In principle `.d-bgc-secondary` is the same as `.thing`.

Seems we can do the same for whenever a CSS file has a `background-color`/`border-color`, or (foreground) `color` in `.thing {}`.

braddialpad:
> ```
> <dt-notice kind="info" class="d-bgo50">
>   ...
> </dt-notice>
> ``` 
>   
> That background-color is already set in the CSS (light blue). Without authoring custom CSS, they just want to tweak its background opacity with a CSS Utility class="d-bgo50".

It's true it won't work for this edge case.

@juliodialpad It's possible to have multiple fallbacks in css vars.. Maybe we could do something like:

```
--dt-color-foreground-primary: hsl(var(--dt-color-foreground-primary-hsl) / var(--fco, var(--alpha, ${alpha}%)));
```

juliodialpad:
> <dt-notice kind="info" class="d-bgo50">
>   ...
> </dt-notice>

We have multiple problems to make that work, I don't think this is an easy task to acomplish, we'll need to make a lot of changes on how we get the tokens from Figma and how we process them on dialtone-tokens to make it work.

1. While generating `hsla` variables on `dialtone-generators.cjs`, we're not really taking into account what's a `foreground`, `surface`, `border`, etc. color to set the corresponding `--fco`, `--bgo` or `--bco` property, that's why we're using a more generic `--alpha` variable.

2. The tokens are being declared on `:root` so the `--alpha` variable would be undeclared even if we add that to the `.d-bgo10` utility class.

With the example you provided.

- The background color is: `--notice-color-background: var(--dt-color-surface-info);` -> `--dt-color-surface-info: var(--dt-color-blue-100);` -> `--dt-color-blue-100: #EAF2FA;` which is not even a hsla color that supports transparency. (We'll need to change all the base colors to use HSL instead of HEX values)

Screenshot 2024-10-07 at 1 20 37 p m

Changing the background color to use `*-hsla` variable:
- The background color is: `--notice-color-background: var(--dt-color-surface-info-hsla);` -> `--dt-color-surface-info-hsla: var(--dt-color-blue-100-hsla);` -> `--dt-color-blue-100-hsla: hsl(var(--dt-color-blue-100-h) var(--dt-color-blue-100-s) var(--dt-color-blue-100-l) / var(--alpha, 100%));`

Screenshot 2024-10-07 at 1 19 46 p m

Screenshot 2024-10-07 at 1 25 35 p m

----

To accomplish what you want to do, we need to change every component to use the HSL tokens and take the corresponding opacity css var like so:

- Before
image

- After
Screenshot 2024-10-07 at 1 36 02 p m

We also should remove the `var(--alpha, %100)` from every token, it doesn't really work.

braddialpad:
> That works for when one is in a position to author CSS for `.thing`. I'm thinking of a situation I can see occurring when someone wants to use a CSS Utility to change the opacity of a style property that is already declared in CSS, and they're not in a position to author CSS. Example...
> 
> ```
> <dt-notice kind="info" class="d-bgo50">
>   ...
> </dt-notice>
> ```

@francisrupert as Julio has illustrated this is a really really big change that I don't think is necessary. I think if a developer wants to do the above, they simply have to set the background color via utility class along with the opacity. That should be sufficient right?

```
<dt-notice kind="info" class="d-bgc-secondary d-bgo50">
```

francisrupert (approved):
Cool, thanks for the very aggressive attempt at it, and definitely recognize the massive shift it would require to accomplish what I aimed for in my [codepen](https://codepen.io/francisrupert/pen/rNEooEo?editors=1100). The updates you've accomplished here go far beyond [DLT-1548](https://dialpad.atlassian.net/browse/DLT-1548). Thanks!

PR #529, opened 2024-10-16
fix(avatar): DP-113524 avatar icon not reactive

ninamarina:
# Fix avatar icon not displaying when transferring the call

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket
[DP-113524]

## :book: Description
The problem is that `$slots` (and `$scopedSlots`) is not reactive. That is explained here: https://github.com/vuejs/vue/issues/3517
So in this case, initially the slot was undefined, but then when transferring the call to an external it was defined, but the computed `isIconType` would not reevaluate. 
Changing the computed to a method makes it work as expected.

## :pencil: Checklist

For all PRs:

For all Vue changes:

## :crystal_ball: Next Steps
- Add changes to vue 3

### Before

call-2

### After

call

[DP-113524]: https://dialpad.atlassian.net/browse/DP-113524?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dialtone-vue2/recipes/item_layout/contact_info/contact_info.vue
+++ b/packages/dialtone-vue2/recipes/item_layout/contact_info/contact_info.vue
@@ -28,7 +28,6 @@
             image-alt=""
             :overlay-text="avatar.text"
             :avatar-class="[{ 'd-mln24': index > 0, 'd-bc-brand': !!avatar.halo }]"
-            :show-icon="showAvatarIcon"
           >
             <template
               #icon="{ iconSize }"
@@ -56,7 +55,6 @@
           :seed="avatarSeed"
           :color="avatarColor"
           :presence="presence"
-          :show-icon="showAvatarIcon"
         >
           <template
             #icon="{ iconSize }"
@@ -216,14 +214,6 @@ export default {
     avatarList: {
       type: Array,
       default: null,
-    },
-
-    /**
-     * Show avatar icon
-     */
-    showAvatarIcon: {
-      type: Boolean,
-      default: false,
     },
   },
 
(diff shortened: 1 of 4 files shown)


packages/dialtone-vue2/recipes/item_layout/contact_info/contact_info.vue
```diff
@@ -28,13 +28,13 @@
...
             :overlay-text="avatar.text"
             :avatar-class="[{ 'd-mln24': index > 0, 'd-bc-brand': !!avatar.halo }]"
+            :show-icon="showAvatarIcon"
           >
             <template
               #icon="{ iconSize }"
             >
               <!-- @slot Slot for avatar icon in a list -->
               <slot
-                v-if="hasAvatarIcon"
```
braddialpad, line 37: If I'm not mistaken, all we have to do is remove this condition and everything works correctly right?

braddialpad (changes requested):

packages/dialtone-vue2/recipes/item_layout/contact_info/contact_info.vue
ninamarina: No,  `isIconType` in the avatar component would still not be reactive

braddialpad: What we can do is change isIconType from a computed to a function

```
v-else-if="isIconType()"

...

methods: {
    isIconType () {
      return this.$scopedSlots.icon && this.$scopedSlots.icon();
    },
```

I tested this to make sure it works and it seems to.

before:


after:


When you call a method in a template like this it is called "every time" the component is updated, so it's less efficient than a computed property however it works great for cases like this.

PS: Don't reply if you're sick 😁

ninamarina: great, will try this tomorrow if I feel better, thank you!

braddialpad (approved):
Looks great thanks!

PR #530, opened 2024-10-16
fix: NO-JIRA dialtone-vue incorrect build output

juliodialpad:
# Fix Wrong dialtone-vue incorrect build

Obligatory GIF

- [x] Fix

## :book: Description

- Moved `disableRootScrolling` and `enableRootScrolling` utils from `common/utils` to corresponding dialtone-vue2/3 utils.
- Fixed a warning about a circular dependency of `emoji_picker`
- Fixed an issue where `component-documentation.json` was not generated for dialtone-vue2
- Removed `emoji-toolkit` in favor of `@dialpad/dialtone-emojis`
- Added missing exports.

## :bulb: Context

Detected the dialtone-vue  output changed and was generating files like `dist/packages/dialtone-vue2/components/button.js` instead of just `dist/components/button.js`.

This was due to the usage of common utility classes outside the `packages/dialtone-vue2/3` folders and the `preserveModules` option on vite config. 

## :pencil: Checklist

For all PRs:

For all Vue changes:

- Before

image

- After

image

### DIFF
--- a/common/utils.mjs
+++ b/common/utils.mjs
@@ -22,46 +22,8 @@ export function kebabCaseToPascalCase (string) {
     .join('');
 }
 
-/**
- * checks whether the dt-scrollbar is being used on the root element.
- * @param rootElement {HTMLElement}
- * @returns {boolean}
- */
-function isDtScrollbarInUse (rootElement = document.documentElement) {
-  if (rootElement.hasAttribute('data-overlayscrollbars')) {
-    return true;
-  }
-  return false;
-}
-
-/**
- * This will disable scrolling on the root element regardless of whether you are using dt-scrollbar or not.
- * @param rootElement {HTMLElement}
- */
-export function disableRootScrolling (rootElement = document.documentElement) {
-  if (isDtScrollbarInUse(rootElement)) {
-    rootElement.classList.add('d-scrollbar-disabled');
-  } else {
-    rootElement.classList.add('d-of-hidden');
-  }
-}
-
-/**
- * This will enable scrolling on the root element regardless of whether you are using dt-scrollbar or not.
- * @param rootElement {HTMLElement}
- */
-export function enableRootScrolling (rootElement = document.documentElement) {
-  if (isDtScrollbarInUse(rootElement)) {
-    rootElement.classList.remove('d-scrollbar-disabled');
-  } else {
-    rootElement.classList.remove('d-of-hidden');
-  }
-}
-
 export default {
   getUniqueString,
   kebabCaseToPascalCase,
   flushPromises,
-  disableRootScrolling,
-  enableRootScrolling,
 };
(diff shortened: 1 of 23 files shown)


packages/dialtone-vue2/common/emoji/index.js
```diff
@@ -1,5 +1,5 @@
 import { emojiPattern } from 'regex-combined-emojis';
-import emojiJsonLocal from 'emoji-toolkit/emoji_strategy.json' with { type: 'json' };
+import { emojisIndexed } from '@dialpad/dialtone-emojis';
```
braddialpad, line 2: Nice, this is much better to not have to reference emoji-toolkit anymore. However, there is some process and/or script that nacho has to transfer emoji data from emoji-toolkit into these categories. We will have to perform this whenever we want to update to a newer version of emojis. I'll make a note to get this from him when he's back

common/utils.mjs
```diff
@@ -22,46 +22,8 @@ export function kebabCaseToPascalCase (string) {
...
- * This will enable scrolling on the root element regardless of whether you are using dt-scrollbar or not.
- * @param rootElement {HTMLElement}
- */
-export function enableRootScrolling (rootElement = document.documentElement) {
-  if (isDtScrollbarInUse(rootElement)) {
-    rootElement.classList.remove('d-scrollbar-disabled');
-  } else {
-    rootElement.classList.remove('d-of-hidden');
-  }
-}
```
braddialpad, line 59: Should we have some rule for this sort of thing, when should it go in common/utils vs dialtone-vue?

Should we not have common/utils at all? Or should we have a separate package for common utils?

packages/dialtone-vue2/common/emoji/index.js
juliodialpad: Ah right, I was wondering why we were not using  `@dialpad/dialtone-emojis` on Dialtone-vue 2 version.

So should I remove this change? ESLint was complaining so just figure out updating it to dialtone-emojis fixed it.

common/utils.mjs
juliodialpad: Yes, I thought moving everything to a common package was the way to go BUT as we're using relative paths, that messes up the dialtone-vue build process.

If we move those utility functions to a `@dialpad/dialtone-utilities" package, would be possible to reference them and move common utilities across our monorepo easily, could also be an external dependency, tree-shakeable and everything.

packages/dialtone-vue2/common/emoji/index.js
braddialpad: Nah should be okay to make this change

common/utils.mjs
braddialpad: I think that's probably the way to go, but not necessary for this PR.

braddialpad (approved):

PR #531, opened 2024-10-16
fix(popover): NO-JIRA set -webkit-app-region: no-drag

braddialpad:
# fix(popover): NO-JIRA set -webkit-app-region: no-drag

Obligatory GIF

- [x] Fix

## :book: Description

Add `-webkit-app-region: no-drag` to popover and modal overlays.

## :bulb: Context

The header region in the native app header is very aggressive and by default overrides all mouse events. So this means if you have a popover that's over the header, it won't highlight items and won't trigger click events if you click on the portion that is over top of the app header. It also will not close the popover if you click outside of the popover on the header, which is not expected behaviour. Setting `-webkit-app-region: no-drag` allows click events to pass through.

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

test that it's fixed in the native app

## :link: Sources

https://github.com/electron/electron/issues/1354

### DIFF
--- a/packages/dialtone-css/lib/build/less/components/modal.less
+++ b/packages/dialtone-css/lib/build/less/components/modal.less
@@ -33,6 +33,8 @@
     --modal-header-color-text: var(--dt-color-foreground-primary);
     --modal-dialog-shadow: var(--dt-shadow-large);
 
+    // If we don't set this the native app header region will override all click events on the modal overlay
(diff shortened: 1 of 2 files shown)


juliodialpad (approved):
There's really no way we can test this ourselves, right?.

So I'll approve so you can merge and test on the native app.

braddialpad:
Yeah not really

PR #540, opened 2024-10-24
fix(tokens): NO-JIRA resolveMath for android

braddialpad:
# fix(tokens): resolveMath for android

Obligatory GIF

- [x] Fix

## :book: Description

Created a new transformer to resolve math for android kotlin to fix errors.

## :bulb: Context

Since some of the changes in style dictionary 4 with the new transitive transformers some calculated android sizing were not resolving and causing some syntax errors in the file. This is maybe a bit of a brute force method using eval but it should work well enough for what we currently need.

Before you would see calculations in the android kotlin size tokens for example `10.dp * 0.8 * 1.5`. Now it should all be resolved like so `12.dp`.

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

Check with android team that it's working, make sure build releases correctly.


braddialpad:
Merging this in so I can test.

PR #544, opened 2024-10-25
fix(emoji-text-wrapper): DLT-2176 invalid regular expression error

ninamarina:
# fix(emoji-text-wrapper): DLT-2176 invalid regular expression error

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket
[DLT-2176]

## :book: Description
Fixes the error `Error: SyntaxError: Invalid regular expression: /(X)/g: Nothing to repeat` when entering *️⃣ emoji on a message.
I fixed it by using directly the `emojiPattern` regex instead of the `findEmojis` function.

## :bulb: Context
The error appeared in prod: https://dialpad.atlassian.net/browse/DP-115263

## :pencil: Checklist

For all PRs:

For all Vue changes:

### Previous
image-20241025-192541

### Now
image

[DLT-2176]: https://dialpad.atlassian.net/browse/DLT-2176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dialtone-vue2/common/emoji/index.js
+++ b/packages/dialtone-vue2/common/emoji/index.js
@@ -196,12 +196,3 @@ export function filterValidShortCodes (shortcodes) {
   const filtered = shortcodes ? shortcodes.filter(code => shortcodeToEmojiData(code)) : [];
   return new Set(filtered);
 }
-
-// Finds every emoji in slot text
-// removes duplicates
-// @returns {string[]}
-export function findEmojis (textContent) {
-  const matches = [...textContent.matchAll(emojiRegex)];
-  const emojis = matches.length ? matches.map(match => match[0]) : [];
-  return new Set(emojis);
-}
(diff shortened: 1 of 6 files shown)


juliodialpad:
If the error is within findEmojis function then why not just fixing the function itself instead of going around it?

Is it being used elsewhere?

ninamarina:
> If the error is within findEmojis function then why not just fixing the function itself instead of going around it?
> 
> Is it being used elsewhere?

I think the problem is that findEmojis is already finding matches with the same regex, and then we are doing it again and that's when the invalid regex error occurs. This way we are running the regex only once to find emojis. 

It's not being used anywhere else and I should delete it if we decide this is the correct solution.

braddialpad (approved):
I think this should work.. kind of a weird one alright.

PR #549, opened 2024-10-30
feat(emoji-picker): DP-113112 add shift key info on selected-emoji event

bianca-artola-dialpad:
# add(emoji-picker): DP-113112 add shift key info on selected-emoji event

Obligatory GIF

- [x] Feature

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DP-113112

## :book: Description
This pull request enhances the selected-emoji event by including information about key events. With this update, we can now identify whether an emoji was selected using the Shift + Enter keys.

## :bulb: Context
For going into GA with the new message input component, we need to allow users to add an emoji to the input field with Shift + Enter without closing the emoji picker. To enable this functionality on the product side, it's essential to detect when an emoji is selected using the Shift + Enter key combination.

### DIFF
--- a/packages/dialtone-vue2/components/emoji_picker/modules/emoji_selector.vue
+++ b/packages/dialtone-vue2/components/emoji_picker/modules/emoji_selector.vue
@@ -44,7 +44,7 @@
             :ref="`emojiRef-${indexTab}`"
             type="button"
             :aria-label="emoji.name"
-            @click="$emit('selected-emoji', emoji)"
+            @click="event => handleEmojiSelection(emoji, event)"
             @focusin="$emit('highlighted-emoji', emoji)"
             @focusout="$emit('highlighted-emoji', null)"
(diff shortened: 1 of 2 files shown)


bianca-artola-dialpad:
Firespotter changes: https://github.com/dialpad/firespotter/pull/48999

ninamarina:
Looks good! Do you also need it to add the shiftKey info on click? You would have to add it also to the click event in that case

braddialpad:
> Looks good! Do you also need it to add the shiftKey info on click? You would have to add it also to the click event in that case

Yes I think we would also need this.

bianca-artola-dialpad:
> Looks good! Do you also need it to add the shiftKey info on click? You would have to add it also to the click event in that case

good catch, should be working now. thank you!

ninamarina (approved):

braddialpad (approved):
Great!

PR #557, opened 2024-10-31
fix(component): DLT-2121 move all styling to dialtone-css from vue

braddialpad:
# fix(component): DLT-2121 move all styling to dialtone-css from vue

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2121

## :book: Description

Moved all remaining CSS styling from core components into dialtone-css.

## :bulb: Context

We want all our styling in one place

## :pencil: Checklist

For all PRs:

For all CSS changes:

## :crystal_ball: Next Steps

Do the same thing on recipes

## BREAKING CHANGE CLASSNAME LIST

We will likely have to do a small find/replace migration to fix code overriding these classes. d-item-layout and d-list-item are the highest risk. I would say usage of any of the others should be rare.

Note if this will also include all sub elements such as `d-list-item--focusable`

dt-item-layout -> d-item-layout (and all sub elements)
dt-list-item -> d-list-item (and all sub elements)
enter-active -> d-collapsible__enter-active
leave-active -> d-collapsible__leave-active
dt-empty-list-item -> d-combobox__empty-list-item
dt-description-list -> d-description-list
dt-list-separator -> d-list-item-separator
mention-suggestion-name -> d-mention-suggestion__name
dt-suggestion-list__container -> d-suggestion-list__container
dt-rich-text-editor -> d-rich-text-editor
skeleton-placeholder -> d-skeleton-placeholder


packages/dialtone-css/lib/build/less/components/skeleton.less
```diff
@@ -17,15 +17,19 @@
 //  ============================================================================
 //  @   SKELETON CONTAINER
 //  ----------------------------------------------------------------------------
+
+// The --placeholder-from-color and --placeholder-to-color
+// custom properties can be set on the parent class of the
+// placeholder to control the animation colors.
 .skeleton-placeholder {
```
francisrupert: Namespace with `d-` maybe? Would that be a breaking change to have to update for non-Vue use?

packages/dialtone-css/lib/build/less/components/skeleton.less
```diff
@@ -17,15 +17,19 @@
...
   --placeholder-from-color: hsla(var(--dt-color-black-900-hsl) / 0.2);
   --placeholder-to-color: hsla(var(--dt-color-black-900-hsl) / 0.1);
 
   display: flex;
   width: 100%;
-  background: var(--placeholder-from-color);
+  background: var(--placeholder-from-color, var(--dt-color-black-300));
```
francisrupert: Let's make the fallback color `var(--dt-color-surface-bold-opaque)`. Same for the others below.

francisrupert (commented):
Awesome. A few basic comments. I haven't gone through all them yet, though first one I noticed still with `dt-` classes is Item Layout (with List Item).

packages/dialtone-css/lib/build/less/components/skeleton.less
braddialpad: What about the `--placeholder-to-color`? currently falls back to `var(--dt-color-black-100)`

packages/dialtone-css/lib/build/less/components/skeleton.less
braddialpad: Yeah, but we've already done a few those already 🤷. Updated

braddialpad:
Looking into percy issues, a couple small problems found there.

packages/dialtone-css/lib/build/less/components/skeleton.less
francisrupert: Ah yes, use `var(--dt-color-surface-moderate-opaque)`. That doesn't map to the current Black 100, but it's a less jarring animating shift. 

Also, remove the...
```
style="
  ...
  --placeholder-from-color: ...;
  --placeholder-to-color: ...;
  ...
"
```
...from the `## Prefabricated examples`.

packages/dialtone-css/lib/build/less/components/skeleton.less
```diff
@@ -17,38 +17,42 @@
...
 //  @   SKELETON CONTAINER
 //  ----------------------------------------------------------------------------
-.skeleton-placeholder {
+
+// The --placeholder-from-color and --placeholder-to-color
+// custom properties can be set on the parent class of the
+// placeholder to control the animation colors.
+.d-skeleton-placeholder {
   --placeholder-from-color: hsla(var(--dt-color-black-900-hsl) / 0.2);
   --placeholder-to-color: hsla(var(--dt-color-black-900-hsl) / 0.1);
```
francisrupert:
```suggestion
  --placeholder-from-color: var(--dt-color-surface-bold-opaque);
  --placeholder-to-color: var(--dt-color-surface-moderate-opaque);
```
Actually, maybe they should be here and not even need a fallback anywhere? I haven't tested it, but off top of my head I can't figure a situation where a fallback kicks in.

packages/dialtone-css/lib/build/less/components/skeleton.less
francisrupert: Maybe belay that, look at [this other comment](https://github.com/dialpad/dialtone/pull/557#pullrequestreview-2409386555) first.

But still do the removal of `style="..."` comment above.

packages/dialtone-css/lib/build/less/components/skeleton.less
braddialpad: Yeah don't really think we need the fallback... removed it

packages/dialtone-css/lib/build/less/components/skeleton.less
braddialpad: Alright, should be good.

braddialpad:
Alright, all remaining issues should be fixed. Also added a blog post for the breaking change since it could affect anyone overriding component classes.

packages/combinator/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Dialtone vue component combinator",
   "scripts": {
     "lint": "eslint --ext .js,.vue --ignore-path .gitignore --fix src",
-    "test": "vue-cli-service test:unit --timeout 100000 --colors --recursive --glob *.test.js src"
+    "test:old": "vue-cli-service test:unit --timeout 100000 --colors --recursive --glob *.test.js src"
```
juliodialpad, line 7: Why are we doing this? combinator tests were messing up with the other tests?

juliodialpad (approved):
Looks good! this is a pretty nice cleanup.

We should move [core_scroller](https://github.com/dialpad/dialtone/blob/d824647d1ab64453f8276c7c939ddd78ab36af20/packages/dialtone-vue3/components/scroller/modules/core_scroller.vue) styles too, right?

Other than that, this is good to merge

packages/combinator/package.json
braddialpad: yeah if you just run `nx test` it runs all "test" scripts including this one, and they are broken because they haven't been migrated yet.

braddialpad:
> We should move [core_scroller](https://github.com/dialpad/dialtone/blob/d824647d1ab64453f8276c7c939ddd78ab36af20/packages/dialtone-vue3/components/scroller/modules/core_scroller.vue) styles too, right?

Thanks, moved this as well.

PR #574, opened 2024-11-27
chore: NO-JIRA fix icon json format issues, fix tokens gradle add

braddialpad:
# chore: NO-JIRA fix icon json format issues, fix tokens gradle add

Obligatory GIF

- [x] Other (chore)

## :book: Description

Fixed two issues continually causing problems with our releases

the dialtone icons gulpfile was performing some formatting / sorting on the icons json file on build. Because this was not done on commit often it would format and change the json file on release causing "unclean working tree". Now it will run on commit instead via lint-staged

the release-ci.config.js for dialtone-tokens was doing a git add -A so it was possible for it to add "all uncommited files" rather than just the ones it generated changes to. Updated this to only add and commit the specific files it made changes to for better safety.


lint-staged.config.cjs
```diff
@@ -0,0 +1,16 @@
+module.exports = {
+  '*.{js,mjs,cjs,vue}': [
+    'eslint --fix',
+    'vitest related --run',
+  ],
+  '**/dialtone-icons/src/keywords-icons.json': (filenames) => 'nx run dialtone-icons:build --skip-nx-cache',
```
juliodialpad: Not sure this is going to work, does this automatically add the changes to the commit after the command is run?

- I think we could detect changes on `packages/dialtone-icons/src/*.json` instead to include icons and illustrations json files.

- Why the command is receiving a function parameter instead of being just an array like the others? Whe're not using the parameter anyways.

juliodialpad (changes requested):
Why the rename to `lint-staged.config.cjs`?

braddialpad:
> Why the rename to `lint-staged.config.cjs`?

It needs to be javascript instead of json to support function syntax

> Not sure this is going to work, does this automatically add the changes to the commit after the command is run?

Yep thats what lint staged does.

> I think we could detect changes on packages/dialtone-icons/src/*.json instead to include icons and illustrations json files.

Updated

> Why the command is receiving a function parameter instead of being just an array like the others? Whe're not using the parameter anyways.

lint-staged will automatically add the file names of the changed files as parameters to the written command, which in the case of running the nx build command it causes an error. By using the functional syntax you can run a command without passing it any filename parameters

juliodialpad (approved):
Nice! thanks for the fix

PR #582, opened 2024-12-09
fix(attachment-carousel): css cleanup

braddialpad:
# fix(attachment-carousel): css cleanup

- [x] Fix

## :book: Jira Ticket

## :book: Description


packages/dialtone-css/lib/build/less/recipes/attachment_carousel.less
```diff
@@ -3,7 +3,7 @@
   width: var(--dt-space-1000);
   max-height: 100px; /* stylelint-disable-line meowtec/no-px */
 
-  .d-recipe-attachment-carousel__media-list {
+  &__media-list {
```
juliodialpad: I was keeping them like `.d-recipe-attachment-carousel__media-list` as it's a lot simplier to find those classes when you're debugging or trying to fix something quickly.

Might seem too verbose, but it's a good practice I read somewhere and made sense for me.

juliodialpad: Just need to move them to the right place of course

braddialpad: I feel like this is one of those debated topics in CSS. writing the whole class name gives it better searchability, but using `&` makes it easier to see what is related.

I made the restructure, let me know if that's what you were thinking.

juliodialpad (approved):
Yes, this is more likely what I was looking for, thanks!

PR #584, opened 2024-12-09
fix(callbar_button): DLT-2122 css cleanup

braddialpad:
# fix(callbar_button): css cleanup

- [x] Fix

## :book: Jira Ticket

## :book: Description


iropolo (approved):

packages/dialtone-css/lib/build/less/recipes/callbar_button.less
```diff
@@ -27,7 +27,7 @@
 .d-recipe-callbar-button--active,
 .d-recipe-callbar-button--active:hover {
   .base-button__icon {
-    color: var(--primary-color);
+    color: var(--dt-theme-mention-color-background); // best replacement for --primary-color we have at the moment
```
iropolo, line 30: why is that?

PR #586, opened 2024-12-09
fix(leftbar_row): DLT-2122 css cleanup

braddialpad:
# fix(leftbar_row): css cleanup

- [x] Fix

## :book: Jira Ticket

## :book: Description


packages/dialtone-css/lib/build/less/recipes/leftbar_row.less
```diff
@@ -342,7 +342,7 @@
   animation-fill-mode: both;
 }
 
-@keyframes opacity-pulsate {
```
braddialpad, line 345: Strange this was defined 3 times in a row.. There wasn't a reason for that right?

braddialpad:
Put a couple other tiny little fixes for message_input and top_banner_info in with this PR.

iropolo (approved):

iropolo:
Actions failing for some reason...

PR #592, opened 2024-12-12
feat: DLT-2236 update emoji set

iropolo:
# Update emoji set

## :book: Description

Emoji set updated to 9.0 -> https://blog.joypixels.com/joypixels-9-0-now-available/

test -> https://dialtone.dialpad.com/vue/deploy-previews/pr-592/?path=/story/components-emoji-picker--default


braddialpad (approved):
Tested out emoji's that are only available in emoji 15.1. Seems to work nicely!

Is there a way you can automate this, or at least share the process for doing this? Problem is if this happens and you are not here we cannot update it.

iropolo:
> Tested out emoji's that are only available in emoji 15.1. Seems to work nicely!
> 
> Is there a way you can automate this, or at least share the process for doing this? Problem is if this happens and you are not here we cannot update it.

I follow the instructions here packages/dialtone-vue2/components/emoji_picker/README.md
I will update later today or next week this document to be more clear.

PR #593, opened 2024-12-13
feat: DLT-2291 add rich text formatting

braddialpad:
### Rich Text Editor

- code extension for inline code
- UI dialog for editing links
- bubble menu for editing links directly when clicking on them in the editor
- the component will now output in all different formats with every change (json, html, text) under json-input, html-input, text-input events. The regular `input` event still outputs in the format selected via outputFormat. This is for v-sync/v-model purposes.

### Message Input

- Added all rich text formatting icon buttons (bold, italic, link etc) to top of message input component
- Default output to json rather than text
- Add link related props
- Implement lastActiveNodes to fix complex issue of nested lists showing both types as active when really it should only be the closest node.
- use deepEqual to get an accurate comparison when using json data.
- Added single boolean prop "richText" to toggle between "rich text" mode and "text only mode" for dialpad to dialpad vs dialpad to phone messages. Disabling this will hide the rich text buttons and disable the rich text extensions.
- Added props to set i18n text for each rich text button

It's very important to be able to release this without breaking existing functionality in the product and also in the editor component. I have tested this version of the component in the current master after making some small changes to props and such. It seems to be working okay, but I will run it through some QA testing.

Apologies that this PR is a bit large. I'm going to blame it on this being a hackathon project and not going through the normal channels 😁


packages/dialtone-css/lib/build/less/components/rich-text-editor.less
```diff
@@ -30,11 +37,28 @@
...
+  // Left-Border on all buttons except the first one (separators)
+  button + button {
+    border-left: 1px solid var(--dt-color-border-default);
+  }
+
+  button {
+    padding: var(--dt-space-300) var(--dt-space-450);
+  }
+}
+
```
francisrupert: I provided feedback with @francotirabasso on this and some other things, but the key thing: 

I say this should be a Popover (or at least emulate Popover style) and just use three straight clear buttons. 

Though of course confirm with Franco before doing so.

The result would be effectively...

```html
<dt-popover class="d-p4">
  <dt-stack direction="row" gap="200">
    <dt-button kind="muted" importance="clear">Edit</dt-button>
    <dt-button kind="muted" importance="clear">Open link</dt-button>
    <dt-button kind="danger" importance="clear">Remove</dt-button>
  </dt-stack>
</dt-popover>
```

https://github.com/user-attachments/assets/4e6b053b-4d80-4fa8-bdd6-2547d88fdad5

packages/dialtone-css/lib/build/less/components/rich-text-editor.less
```diff
@@ -4,6 +4,13 @@
     background: var(--dt-color-surface-secondary);
   }
 
+  code:not(.d-rich-text-editor__code-block > code) {
+    padding: var(--dt-space-200) var(--dt-space-300);
+    color: var(--dt-color-purple-400);
+    background-color: var(--dt-color-purple-100);
+    border-radius: var(--dt-size-200);
```
francisrupert:
```suggestion
    padding: var(--dt-space-200);
    color: var(--dt-color-foreground-warning);
    background-color: var(--dt-color-surface-secondary-opaque);
    border-radius: var(--dt-size-radius-200);
    border: var(--dt-size-border-100) solid var(--dt-color-border-subtle);
    font: var(--dt-typography-code-md);
```

image

francisrupert:
For `.d-rich-text-editor__code-block` (not the inline one) add:
```
font: var(--dt-typography-code-md);
```

(I'm too lazy to find it in the file changes diff)

image

francisrupert:
Line break issue when when converting multiple lines of text to codeblock

https://github.com/user-attachments/assets/624749b4-18dd-4955-a16e-cc745a103d3f

francisrupert:
When there are multiple links, it's difficult to move text cursor around. 

https://github.com/user-attachments/assets/73f502bf-7510-4790-ba27-5747344ae7c4

francisrupert:
- Buttons should only be functional if focus is in compose field.
- No button should be shown as active when focus is not in compose field.

https://github.com/user-attachments/assets/ae2dba74-95f3-43cb-bdd7-8398936e3274

francisrupert:
A nested list of a different type should only reflect that type as active. Currently it shows as being both number and bullet. Technically true, though ideally the active one should only reflect that type even if nested.

https://github.com/user-attachments/assets/e5f71528-f6f0-4039-b3f2-11bd0e35b8a7

packages/dialtone-css/lib/build/less/components/rich-text-editor.less
```diff
@@ -30,11 +37,28 @@
     blockquote {
       margin-left: 0;
       padding-left: var(--dt-space-400);
-      border-left: var(--dt-size-border-300) solid var(--dt-color-foreground-muted-inverted);
+      border-left: var(--dt-size-border-300) solid var(--dt-color-foreground-secondary-inverted);
```
francisrupert:
```suggestion
      border-left: var(--dt-size-border-300) solid var(--dt-color-border-subtle);
```

francisrupert:
Each action definitely needs a Tooltip, including keyboard shortcut if it has one.

@francotirabasso, which treatment do you like? The first one uses [Keyboard Shortcut component](https://dialtone.dialpad.com/components/keyboard-shortcut.html). The second one is just tertiary color text. 

image

francisrupert:
@braddialpad, @francotirabasso,

I wonder if we should provide keyboard accessibility to the link options when the cursor is within the link. Gmail example attached:

https://github.com/user-attachments/assets/8c223bee-330e-4408-82a2-b75548558493

braddialpad:
Thanks for the feedback!

braddialpad:
Finally ready for review!!!

packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
```diff
@@ -1,15 +1,263 @@
...
+          />
+        </template>
+      </dt-recipe-message-input-button>
+
+      <dt-recipe-message-input-button
+        :aria-label="strikeButtonOptions.ariaLabel"
+        :tooltip-text="strikeButtonOptions.tooltipText"
+        :keyboard-shortcut-text="strikeButtonOptions.keyboardShortcutText"
+        data-qa="strikethrough"
+        :is-active="$refs.richTextEditor?.editor?.isActive('strike') && isFocused"
```
ninamarina: extract this to a computed like this:
```
isNodeActive(type) {
  return (
    $refs.richTextEditor?.editor?.isActive(type) && this.isFocused
  );
}
```

packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
```diff
@@ -1,15 +1,263 @@
...
+      >
+        <template #icon>
+          <dt-icon-strikethrough
+            class="d-recipe-message-input__button-stack-icon"
+            size="200"
+          />
+        </template>
+      </dt-recipe-message-input-button>
+      <div class="d-recipe-message-input--button-group-divider" />
+      <dt-popover
```
ninamarina: This popover can be extracted to a separate component to avoid making this template so long

packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
```diff
@@ -1,15 +1,263 @@
...
+                </dt-button>
+              </dt-stack>
+            </dt-stack>
+          </dt-stack>
+        </template>
+      </dt-popover>
+
+      <div class="d-recipe-message-input--button-group-divider" />
+
+      <dt-recipe-message-input-button
```
ninamarina: Seems like this `dt-recipe-message-input-button` is repeated a lot, can the buttons be defined in an array and then loop through that array to render them?

packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
braddialpad: wouldn't work well in a computed because refs are not reactive, however I think a method should work.

packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
braddialpad: yeah this component in general is too much code. I think putting all this into an array wouldn't be that nice to work with given there are some exceptional differences between the buttons, but I think what I'll do is take this entire "topbar" and make it it's own component.

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
```diff
@@ -1,19 +1,64 @@
...
+      :editor="editor"
+      :should-show="bubbleMenuShouldShow"
+      :tippy-options="tippyOptions"
+      style="visibility: visible;"
+    >
+      <div class="d-popover__dialog">
+        <dt-stack
+          direction="row"
+          gap="200"
+        >
```
francisrupert: * Edit link popover needs padding of at least 2, maybe 4. 
* May not need gap at all since all three buttons don't have a border.

Consult with @francotirabasso on that kind of tiny UI detail.

image

braddialpad: @francotirabasso any guidance here?

braddialpad:
@ninamarina I split the topbar and link popover out into two separate components. Ready for re-review.

ninamarina:
🎉 I really like this! Looks much cleaner with all the component separated.

One thing I noticed, the buttons are working, but they are not showing the selected state:

messageinput

packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.vue
```diff
@@ -695,10 +853,78 @@ export default {
...
+        this.initLinkDialog();
+      } else {
+        this.hideLinkBubbleMenu = false;
+        this.$refs.richTextEditor?.focusEditor();
+      }
+    },
+
+    // eslint-disable-next-line complexity
+    handleTopbarClick (type) {
+      switch (type) {
```
ninamarina: Maybe here instead of a switch you could use a map, so that it's easier to modify later:

```
const commands = {
    bold: 'toggleBold',
    italic: 'toggleItalic',
    ...
  };
  
  const editor = this.$refs.richTextEditor?.editor;
  if (editor && commands[type]) {
      editor.chain().focus().commands[type]().run();
  }
  ```

braddialpad: can't execute a function via a string in this way, but I found an alternative that works well.

braddialpad:
> 🎉 I really like this! Looks much cleaner with all the component separated.
> 
> One thing I noticed, the buttons are working, but they are not showing the selected state:
> 

Thanks, fixed this.

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
braddialpad: Set to 4 for now, unless suggested otherwise

ninamarina:
> > 🎉 I really like this! Looks much cleaner with all the component separated.
> > One thing I noticed, the buttons are working, but they are not showing the selected state:
> 
> Thanks, fixed this.

Maybe I am doing something wrong but I still not see it working. I tried vue 2 and vue 3 in the preview.

braddialpad:
> Maybe I am doing something wrong but I still not see it working. I tried vue 2 and vue 3 in the preview.

Hmm can you try hard refresh? And if that doesn't work tell me exactly which buttons you are pressing.

ninamarina:
> > Maybe I am doing something wrong but I still not see it working. I tried vue 2 and vue 3 in the preview.
> 
> Hmm can you try hard refresh? And if that doesn't work tell me exactly which buttons you are pressing.
> 
>  [  ](https://private-user-images.githubusercontent.com/64808812/407842343-2859c4d6-b2ba-4c6b-9548-c87721d7d020.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzgxNzU5MjMsIm5iZiI6MTczODE3NTYyMywicGF0aCI6Ii82NDgwODgxMi80MDc4NDIzNDMtMjg1OWM0ZDYtYjJiYS00YzZiLTk1NDgtYzg3NzIxZDdkMDIwLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTAxMjklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUwMTI5VDE4MzM0M1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTA2MDc2ZjAyMmRmMDYyMDRjYjU5OTUzNGU2MWNhN2JkNjFlZjk1MmUzNThhZGU4NzE4YWRhMThhNDI3ODJiYzMmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.2FsKbxMX-38MrEqYM2aR3dSICz-RqPwaeeM9RbL6pZA) [ ](https://private-user-images.githubusercontent.com/64808812/407842343-2859c4d6-b2ba-4c6b-9548-c87721d7d020.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzgxNzU5MjMsIm5iZiI6MTczODE3NTYyMywicGF0aCI6Ii82NDgwODgxMi80MDc4NDIzNDMtMjg1OWM0ZDYtYjJiYS00YzZiLTk1NDgtYzg3NzIxZDdkMDIwLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTAxMjklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUwMTI5VDE4MzM0M1omWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTA2MDc2ZjAyMmRmMDYyMDRjYjU5OTUzNGU2MWNhN2JkNjFlZjk1MmUzNThhZGU4NzE4YWRhMThhNDI3ODJiYzMmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.2FsKbxMX-38MrEqYM2aR3dSICz-RqPwaeeM9RbL6pZA)

It worked in incognito! thank you

ninamarina (approved):

PR #597, opened 2024-12-16
feat(hovercard): DP-119841 hovercard fixes for Fast Connect

ninamarina:
# Hovercard fixes for Fast Connect

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket
[DP-119841]

## :book: Description
Some features needed for Fast Connect, and a bug fix.

- Adds "open" prop to Hovercard, to allow to programmatically open / close the hovercard, works like the "open" prop for the Popover component

- Fixes a bug that happened when hovering mentions that are close.
To fix this I removed the special behavior for the timer, that when changing from one hovercard to another would reduce the delay. I think it doesn't have a noticeable impact for the user and it was causing the hovercard to change to another one when passing through a mention, making it imposible to access the buttons for this case:

| Before | Now |
|------|-----|
| mentions-prev | mentions |

- Adds `enterDelay` and `leaveDelay` params

## :bulb: Context
This is for the Launchathon of Fast Connect: https://www.figma.com/design/gOHiwlDi8bkaV2rkDeR0NP/Fast-Connect?node-id=4921-60460&p=f&t=8uDpFPXJuc88aZzI-0

[DP-119841]: https://dialpad.atlassian.net/browse/DP-119841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dialtone-vue2/components/popover/popover.vue
juliodialpad, line 1: I don't like using 'hovercard' for naming stuff inside `popover` component makes me feel like we're mixing two components, Hovercard should be extending the popover, but we shouldn't need to modify the popover itself

juliodialpad (dismissed):
Hovercard changes look good, please check my comment about the popover changes.

juliodialpad:
Don't forgot to update the docs on the [HoverCard component page](https://dialtone.dialpad.com/deploy-previews/pr-597/components/hovercard.html)

packages/dialtone-vue2/components/popover/popover.vue
braddialpad: I think it should be possible to pass in listeners from the hovercard component and then use v-sync / v-model to manage the "open" prop. Let me know if there is some blocker I am not seeing.

braddialpad (approved):
I'm okay with you merging this and then fixing it later since this works, and is probably a blocker for your launch-a-thon team, but I agree with @juliodialpad that it would be ideal to have all hovercard logic in the hovercard component and we should try to do that. I realize popover is a very complicated component and it might take a while.

ninamarina:
> I'm okay with you merging this and then fixing it later since this works, and is probably a blocker for your launch-a-thon team, but I agree with @juliodialpad that it would be ideal to have all hovercard logic in the hovercard component and we should try to do that. I realize popover is a very complicated component and it might take a while.

I created this ticket and added it to the next sprint: https://dialpad.atlassian.net/browse/DLT-2245. Thanks!

packages/dialtone-vue2/components/popover/popover.vue
ninamarina: Absolutely agree. I will do those changes in this ticket  https://dialpad.atlassian.net/browse/DLT-2245

PR #615, opened 2025-01-22
refactor: DLT-1450 custom item layout

juliodialpad:
# Refactor components using item-layout to remove custom styling

Obligatory GIF

- [x] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1450

## :book: Description

- Added `unstyled` prop to components using `dt-recipe-item-layout`

## :bulb: Context

- Components using `<dt-item-layout>` needed to make a lot of overrides in order to achieve the desired behavior.
- On #160 we introduced an `unstyled` prop to `<dt-item-layout>` to avoid the previously mentioned issues, this migrates our base components using `<dt-item-layout>` to use the unstyled version of it and add any custom css into its own classes, avoiding the overriding hell we had previously.

## :pencil: Checklist

For all PRs:

For all Vue changes:

For all CSS changes:

Note: Made this change Vue 2 only on purpose to make it easier to review, I'll migrate to vue 3 once approved and will request a quick review after that just to verify that the changes were migrated correctly.

## :crystal_ball: Next Steps

- Give QA a heads up once this makes it to beta so they can keep an eye open on any change that might be related to this.
- Verify and migrate DtItemLayout usages on Dialpad
- Remove the `unstyled` prop and the custom stylings on DtItemLayout


braddialpad:
Some unintended changes on percy due to this.

packages/dialtone-css/lib/build/less/components/item-layout.less
```diff
@@ -20,91 +20,56 @@
   font-size: var(--dt-font-size-200);
   line-height: var(--dt-font-line-height-300);
 
-  > .d-item-layout--content {
+  > .d-item-layout__content {
```
braddialpad, line 23: This naming was definitely wrong, but we'll have to be careful about breaking any existing CSS overrides in product

braddialpad (commented):
Will test a bunch of example cases myself tomorrow. This looks great, but is also a bit of a scary change!

packages/dialtone-css/lib/build/less/components/item-layout.less
juliodialpad: I was thinking a migration will still be needed anyway due to some changes on paddings, the overrides they were doing and to be able to remove the `unstyled` prop from the <item-layout>

I will take care of this migration, I can release this to a beta version and make the migration before merging this to staging.

packages/dialtone-vue2/recipes/conversation_view/feed_item_row/feed_item_row.test.js
```diff
@@ -126,7 +126,7 @@ describe('DtFeedItemRow tests', () => {
         propsData = {
           ...basePropsData,
           showHeader: true,
-          avatarImageUrl: 'https://i1.sndcdn.com/avatars-000181324408-652e57-t500x500.jpg',
+          avatarImageUrl: '/avatar1.png',
```
braddialpad, line 129: Is this supposed to be an actual test image?

braddialpad (approved):
Did component comparison on the changed components. Looks good by me. Remember to do the migration, and also to do vue 3

packages/dialtone-vue2/recipes/conversation_view/feed_item_row/feed_item_row.test.js
juliodialpad: It does nothing on that test actually, just to avoid a warning maybe, but that file actually exists, the previous link wasn't working correctly tho.

braddialpad: ah yeah, I should have commented the .story rather than the test. I just mean the image shows up like this:

Screenshot 2025-02-03 at 10 47 21 AM

we should put in some sort of actual placeholder image.

juliodialpad: Ah right, forgot about the base, will update to append the base URL

PR #616, opened 2025-01-23
fix(docs): NO-JIRA the class output in divide width docs was backwards

braddialpad:
# fix(docs): NO-JIRA the class output in divide width docs was backwards

border-right and border-left was reversed when compared to the actual css output.
see: https://dialtone.dialpad.com/utilities/borders/divide-width.html#classes


DestaDialpad (approved):
Thanks for fixing this

PR #631, opened 2025-02-12
fix(tokens): DLT-2348 android font sizes were dp instead of sp

johnp-dialpad:
# fix(tokens) DLT-2348: Fix typography font sizes being dp instead of sp for android

Obligatory GIF

- [X] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2348

## :book: Description

Typography font size tokens were being resolved to `dp` instead of `sp` which is the correct type to use for font size.

## :bulb: Context

For some reason, font sizes in typography tokens are deemed as `dimension` type in tokens-studio which meant that the transform to turn px into dp was being applied to them. The fix is to check the name instead of the type in the filter.

## :pencil: Checklist

For all PRs:

- [X] I have ensured no private Dialpad links or info are in the code or pull request description (Dialtone is a public repo!).
- [X] I have reviewed my changes.
- [X] I have added all relevant documentation.
- [X] I have considered the performance impact of my change.

## :crystal_ball: Next Steps

Unit tests to ensure font size tokens are in sp. Perhaps by checking the name ends with 'fontSize'

### DIFF
--- a/packages/dialtone-tokens/dialtone-transforms.js
+++ b/packages/dialtone-tokens/dialtone-transforms.js
@@ -6,7 +6,7 @@
 
 import Color from 'tinycolor2';
 
-const SIZE_IDENTIFIERS = ['fontSizes', 'sizing', 'borderWidth', 'borderRadius', 'blur', 'spread', 'x', 'y', 'dimension'];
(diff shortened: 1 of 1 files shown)


braddialpad (approved):
Hmm might be something that should be fixed on the tokens studio side eventually, but I think this is a good fix for now.

PR #635, opened 2025-02-17
feat: DP-128277 footer slot for DtDropdown

tiagox:
# DP-128277 footer slot for DtDropdown

Obligatory GIF

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket

DP-128277

## :book: Description

Add the ability to provide a footer section into a Dropdown. The `#footer` will be the footer of the underlying `DtPopover` in the `DtDropdown`.

## :pencil: Checklist

For all PRs:

- [ ] I have reviewed my changes.

For all Vue changes:

For all CSS changes:

image

## :link: Sources

Figma link for the feature: https://www.figma.com/design/dkjx25f3qXIJeTYbDBq6O3/%5BDP-116475%5D-Agent-log-in-%2F-log-out?node-id=5408-48888&m=dev

### DIFF
--- a/packages/dialtone-vue2/components/dropdown/dropdown_default.story.vue
+++ b/packages/dialtone-vue2/components/dropdown/dropdown_default.story.vue
@@ -18,10 +18,7 @@
     @highlight="$attrs.onHighlight"
     @opened="$attrs.onOpened"
   >
-    <template
-      slot="anchor"
-      slot-scope="{ attrs }"
-    >
+    <template #anchor="{ attrs }">
       <div
         v-if="$attrs.anchor"
         v-html="$attrs.anchor"
@@ -33,10 +30,7 @@
         Click to open
       </dt-button>
     </template>
-    <template
-      slot="list"
-      slot-scope="{ close }"
-    >
+    <template #list="{ close }">
       <div
         v-if="$attrs.list"
         v-html="$attrs.list"
@@ -51,6 +45,13 @@
       >
         {{ item.name }}
       </dt-list-item>
+    </template>
+    <template #footer="{ close }">
+      <div class="d-px12">
+        <dt-button @click="close">
+          Dropdown footer
+        </dt-button>
+      </div>
     </template>
   </dt-dropdown>
 </template>
--- a/packages/dialtone-vue2/components/dropdown/dropdown_default.story.vue
+++ b/packages/dialtone-vue2/components/dropdown/dropdown_default.story.vue
@@ -46,12 +46,11 @@
         {{ item.name }}
       </dt-list-item>
     </template>
-    <template #footer="{ close }">
-      <div class="d-px12">
-        <dt-button @click="close">
-          Dropdown footer
-        </dt-button>
-      </div>
+    <template #footer>
+      <div
+        v-if="$attrs.footer"
+        v-html="$attrs.footer"
+      />
     </template>
   </dt-dropdown>
 </template>
--- a/packages/dialtone-vue2/components/dropdown/dropdown_variants.story.vue
+++ b/packages/dialtone-vue2/components/dropdown/dropdown_variants.story.vue
@@ -11,15 +11,12 @@
         :transition="$attrs.transition"
         :open="$attrs.open"
       >
-        <template slot="anchor">
-          <dt-button>
+        <template #anchor="{ attrs }">
+          <dt-button v-bind="attrs">
             {{ variant }} aligned dropdown
           </dt-button>
         </template>
-        <template
-          slot="list"
-          slot-scope="{ close }"
-        >
+        <template #list="{ close }">
           <dt-list-item
             v-for="(item) in items"
             :key="item.id"
@@ -45,10 +42,7 @@
         :transition="$attrs.transition"
         @opened="$attrs.onOpened"
       >
-        <template
-          slot="anchor"
-          slot-scope="{ attrs }"
-        >
+        <template #anchor="{ attrs }">
           <div
             v-if="$attrs.anchor"
             v-html="$attrs.anchor"
@@ -60,10 +54,7 @@
             with sections
           </dt-button>
         </template>
-        <template
-          slot="list"
-          slot-scope="{ close }"
-        >
+        <template #list="{ close }">
           <dt-list-item-group
             :list-class="$attrs.listClass"
           >
@@ -106,10 +97,7 @@
         :transition="$attrs.transition"
         @opened="$attrs.onOpened"
       >
-        <template
-          slot="anchor"
-          slot-scope="{ attrs }"
-        >
+        <template #anchor="{ attrs }">
           <div
             v-if="$attrs.anchor"
             v-html="$attrs.anchor"
@@ -121,10 +109,7 @@
             with sections and headings
           </dt-button>
         </template>
-        <template
-          slot="list"
-          slot-scope="{ close }"
-        >
+        <template #list="{ close }">
           <dt-list-item-group
             heading-class="d-py4 d-px8 d-fw-semibold d-c-default"
             :list-class="$attrs.listClass"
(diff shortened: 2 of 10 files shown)


packages/dialtone-vue2/components/dropdown/dropdown_variants.story.vue
```diff
@@ -162,6 +162,60 @@
...
+          slot="footer"
+          slot-scope="{ close }"
+        >
+          <div class="d-px12">
+            <dt-button @click="close">
+              Dropdown footer
+            </dt-button>
+          </div>
+        </template>
+      </dt-dropdown>
```
juliodialpad: A little clean up suggested, we don't really need all of those variables as the `variants` stories are meant to be `static` examples, we only keep the `:open="$attrs.open"` as it is needed for visual testing.

```suggestion
      <dt-dropdown
        class="d-mr8"
        :open="$attrs.open"
      >
        <template
          slot="anchor"
          slot-scope="{ attrs }"
        >
          <dt-button
            v-bind="attrs"
          >
            with footer
          </dt-button>
        </template>
        <template
          slot="list"
          slot-scope="{ close }"
        >
          <dt-list-item
            v-for="(item) in items"
            :key="item.id"
            role="menuitem"
            :navigation-type="$attrs.navigationType"
            @click="close"
          >
            {{ item.name }}
          </dt-list-item>
        </template>
        <template
          slot="footer"
          slot-scope="{ close }"
        >
          <div class="d-px12">
            <dt-button @click="close">
              Dropdown footer
            </dt-button>
          </div>
        </template>
      </dt-dropdown>
```

packages/dialtone-vue3/components/dropdown/dropdown_variants.story.vue
```diff
@@ -156,6 +156,40 @@
           </dt-list-item-group>
         </template>
       </dt-dropdown>
+
+      <dt-dropdown
+        class="d-mr8"
+        :placement="variant"
+        :padding="$attrs.padding"
+        :navigation-type="$attrs.navigationType"
+        :transition="$attrs.transition"
```
juliodialpad:
```suggestion
```

packages/dialtone-vue3/components/dropdown/dropdown_variants.story.vue
```diff
@@ -156,6 +156,40 @@
...
+        :open="$attrs.open"
+        @highlight="$attrs.onHighlight"
```
juliodialpad:
```suggestion
```

packages/dialtone-vue3/components/dropdown/dropdown_variants.story.vue
```diff
@@ -156,6 +156,40 @@
...
+      >
+        <template #anchor>
+          <dt-button>
+            with footer
+          </dt-button>
+        </template>
```
juliodialpad: Keep Vue 2 and Vue 3 code as close as possible

```suggestion
        <template #anchor="{ attrs }">
          <dt-button v-bind="attrs">
            with footer
          </dt-button>
        </template>
```

juliodialpad (changes requested):
Thanks for contributing! this is a really nice first time contribution, just a small changes requested.

1. Please make sure the slot works on the default story, add it to `dropdown_default.story.vue` and set it on `dropdown.stories.js` too please. You can refer to `DtPopover` component stories or send me a message directly if needed.

 Thanks!

tiagox:
> Keep Vue 2 and Vue 3 code as close as possible

With this in mind I've replaced the deprecated `slot=""` as a prop for the shorthand syntax in Vue2.

Also, I believe I have addressed your comments:

- Footer added to the default story.
- `footer` slot added in prop definition for live editing in Storybook.

packages/dialtone-vue2/components/popover/popover.vue
```diff
@@ -595,6 +595,10 @@ export default {
       // there is no aria-label and the labelledby should point to the anchor.
       return this.ariaLabelledby || (!this.ariaLabel && getUniqueString('DtPopover__anchor'));
     },
+
+    hasFooter () {
+      return this.$slots.footerContent || (this.$scopedSlots.footerContent && this.$scopedSlots.footerContent());
+    },
```
braddialpad: Hmm I don't believe this will be reactive because it's referencing $slots. Instead of a computed make this a method and call it from the template.

packages/dialtone-vue2/components/dropdown/dropdown_default.story.vue
```diff
@@ -52,6 +46,13 @@
         {{ item.name }}
       </dt-list-item>
     </template>
+    <template #footer="{ close }">
+      <div class="d-px12">
+        <dt-button @click="close">
+          Dropdown footer
+        </dt-button>
+      </div>
+    </template>
```
braddialpad: For this one you can just go

```
<template #footer>
  <div v-if="$attrs.footer" v-html="$attrs.footer" />
</template>
```

This allows the user to pass in whatever html tags they want to it from storybook (empty by default).

packages/dialtone-vue2/components/popover/popover.vue
braddialpad, line 1: Do we also need to do this in the vue 3 popover?

braddialpad (changes requested):
Thanks for the PR!

packages/dialtone-vue2/components/popover/popover.vue
tiagox: I'm not sure. This was introduced because in Vue2, the `#footer` slot from the Dropdown wasn't working at all.

@ninamarina can you elaborate on this?

ninamarina: Exactly, the way we check for slot content in Vue 3 is different, so the problem was only in Vue 2.

braddialpad: sounds good suspected that might be the case. Just wanted to make sure we didn't miss copying something to vue 3.

packages/dialtone-vue2/components/popover/popover.vue
tiagox: I've moved this to the methods section and it's not being reactive either. If I set the `footer` in storybook, then I need to make the component to re-create (going to another page and back to the dropdown) to see the new footer finally rendered.

I'm not sure why setting the slot `#footerContent` is not showing it in `$slots.footerContent`.

braddialpad: hmmm just a sec going to take a look myself.

tiagox: Here is the interaction

[scoped-slot-re-rendering.webm](https://github.com/user-attachments/assets/ce89cbc8-6424-4710-b415-945024f9db9c)

only applicable for Vue2. Vue3 works fine.

braddialpad: It doesn't even work on the popover component on the currently live storybook, so I think that's where the problem is https://dialtone.dialpad.com/vue/index.html?path=/story/components-popover--default

braddialpad: Okay so I've confirmed this issue, it is definitely a bit strange... If you don't initialize the slots with a value they just never update when entered in the UI, if you initialize them with anything, even an empty string, they will then update reactively, so I added this to "popover.stories.js"

```
args: {
    headerContent: '',
    footerContent: '',
    content: '',
  },
```

It's weird because props do not work this same way.

There is probably a better, more global fix we can do for this but it's really nothing to do with this PR. I pushed a fix for this case, you should be good to go.

Thanks for struggling through that with us.

braddialpad (approved):
Looks good thanks!

PR #639, opened 2025-02-19
fix: NO-JIRA custom item layout

juliodialpad:
# Fix custom item layout

Obligatory GIF

- [x] Fix

## :book: Description

- Fixed missing font-size
- Fixed missing min-width

## :bulb: Context

Errors reported on product

## :pencil: Checklist

For all PRs:

For all Vue changes:

For all CSS changes:

### DIFF
--- a/packages/dialtone-css/lib/build/less/components/list-item.less
+++ b/packages/dialtone-css/lib/build/less/components/list-item.less
@@ -37,6 +37,7 @@
   gap: var(--dt-space-400);
   min-height: calc(var(--dt-size-550) + var(--dt-size-300));
   padding: var(--dt-space-300) var(--dt-space-400);
+  font-size: var(--dt-font-size-200);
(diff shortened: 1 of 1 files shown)


braddialpad (approved):
Looks fine, my only concern is that there are potentially more CSS properties like this that we might have missed.

juliodialpad:
> Looks fine, my only concern is that there are potentially more CSS properties like this that we might have missed.

Don't think there are many like this, I omitted them intentionally as they seemed to make no change as we can see on the visual tests but they were actually doing something elsewhere.

braddialpad:
> Don't think there are many like this, I omitted them intentionally as they seemed to make no change as we can see on the visual tests but they were actually doing something elsewhere.

Yeah, I think because the default font size is set to 12 instead of 15 on ubervoice

PR #650, opened 2025-02-26
fix(button): DLT-2310 fix colors for circle button

ninamarina:
# Fix colors for circle button

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket
[DLT-2310]

## :book: Description
In the circle button, the colors for border, background and text where incorrect.

**Variant Outlined, Default**
Previous:
image

Now:
image

**Variant Clear, Default**
Previous:
image

Now: 
image

[DLT-2310]: https://dialpad.atlassian.net/browse/DLT-2310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dialtone-css/lib/build/less/components/button.less
+++ b/packages/dialtone-css/lib/build/less/components/button.less
@@ -226,20 +226,7 @@
 .d-btn--circle {
   --button-padding-x: var(--button-padding-y-md);
   --button-padding-y: var(--button-padding-y-md);
-  --button-color-text: var(--dt-action-color-foreground-muted-default);
   --button-border-radius: var(--dt-size-radius-circle);
-
-  &:hover:not([disabled]) {
-    --button-color-text: var(--dt-action-color-foreground-muted-hover);
-    --button-color-background: var(--dt-action-color-background-muted-hover);
-  }
-
-  &:active:not([disabled]),
-  &.d-btn--active:not([disabled]),
-  &.d-btn--active:active:not([disabled]) {
-    --button-color-text: var(--dt-action-color-foreground-muted-active);
-    --button-color-background: var(--dt-action-color-background-muted-active);
-  }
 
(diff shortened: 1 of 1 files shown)


francisrupert (changes requested):
@ninamarina, Hmm, we're gonna have to be careful here. How did this come up? The default Circle button was always intentionally presented as Muted even without it explicitly being prop'ed as such. It should be quite rare, if at all, that a circle button is purple. 

I consider this a breaking change. We don't want all current product uses of the default circle button to suddenly go purple.

I would rather all current use of the default Circle button in product first be updated to `muted`. That won't visually change anything as is since they're already muted.

As a result, when **this** update makes it to production it won't change anything since those current circle buttons will have already been muted.

ninamarina:
> of the default Circle button in product first be updated to `muted`. That won't visually change anything as is since they're already muted.

Makes sense. This was reported by @fede-dp although I don't know exactly where they needed to use the purple button. @fede-dp can you provide more information?

fede-dp:
I think this was requested by @lau-dialpad and @axlew was working on it, because we have that variables in the Figma library, attaching image here ⬇️ 
Screenshot 2025-02-27 at 5 39 14 PM

fede-dp:
nevermind! this design change to have rectangular buttons, here https://www.figma.com/design/HX4jW6W91hwGMgBXEXRqbZ?node-id=4850-6698#1116347684 we can close this one out, sorry Nina 🙏🏼

PR #655, opened 2025-02-28
feat: DLT-2350 expand color css utility classes

juliodialpad:
# Expand color CSS utility classes

Obligatory GIF

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2350

## :book: Description

- Updated dialtone-generators to generate every color that exists on `tokens-base-light.json`
- Generated `.d-bc-ai` manually because a linear-gradient border is a special case.

## :bulb: Context

- We had some regex that matched the color combinations that could exist but that's not necessary as we could just find the prefix of the colors and that way if we add a new color it'd be generate automatically without needing to update.

## :pencil: Checklist

For all PRs:

For all CSS changes:


braddialpad (approved):
Nice, much cleaner

PR #657, opened 2025-02-28
ci(icon): DLT-2384 android icons package

johnp-dialpad:
# Android icons package

Obligatory GIF

- [ ] Documentation
- [X] Build system
- [X] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2384

## :book: Description

- Added android gradle project to the dialtone-icons project.
- Added [Valkyrie CLI](https://github.com/ComposeGears/Valkyrie/tree/main?tab=readme-ov-file#cli-tool) binary
- Added build process to the dialtone-icons:build target
- The build process: Use valkyrie to convert SVG icons to Compose ImageVector format
- The publish process uses the same pattern as android tokens

Note: Illustrations are not included because these cannot be converted to android format automatically using valkyrie (illustrations reference color tokens in the svg).

Note: I tried adding Homebrew github actions step to workflows to install valkyrie, but it proved to be awkward because it needs to be added to a lot of workflows in case they ever run the dialtone-icons build. It would also required devs to have homebrew installed on their machine when running a dialtone-icons build. Adding the binary to the repo is easier as it just works and doesn't require updates to workflows or devs local environment.

## :bulb: Context

Android does not support SVG files. This means developers are manually importing icons and converting them to the appropriate format.
This PR creates a `dialtone.dialpad.icons` android project that can be published similar to android dialtone tokens. The package will contain all the current icons in the correct format for android.

## :pencil: Checklist

For all PRs:

- [X] I have ensured no private Dialpad links or info are in the code or pull request description (Dialtone is a public repo!).
- [X] I have reviewed my changes.
- [X] I have considered the performance impact of my change.

## :link: Sources

https://github.com/ComposeGears/Valkyrie/tree/main?tab=readme-ov-file#cli-tool


braddialpad (approved):
This all seems to be working perfectly! thanks for the change.

I just have one question, I was a bit confused about the output location of the kotlin files after processing, I was expecting them to be in "dist" because they were files generated by the build, however they were in `android/src/kotlin/icons`. I can see why you put them here because `.kt` files are technically uncompiled source code files, however it might be nice to have all "build generated files" in one place just for simplicity, what do you think? Forgive me if I am breaking some sort of android standard here 😄.

johnp-dialpad:
> This all seems to be working perfectly! thanks for the change.
> 
> I just have one question, I was a bit confused about the output location of the kotlin files after processing, I was expecting them to be in "dist" because they were files generated by the build, however they were in `android/src/kotlin/icons`. I can see why you put them here because `.kt` files are technically uncompiled source code files, however it might be nice to have all "build generated files" in one place just for simplicity, what do you think? Forgive me if I am breaking some sort of android standard here 😄.

Thanks for the review. I was following the same pattern where currently the generated `.vue` icons and illustrations are added to the dialtone-icons/src folder and are git ignored. I can move the `.kt` files to dist though if you like. Would you prefer them in dialtone-icons/dist or dialtone-icons/android/dist ?

braddialpad:
> Thanks for the review. I was following the same pattern where currently the generated `.vue` icons and illustrations are added to the dialtone-icons/src folder and are git ignored. I can move the `.kt` files to dist though if you like. Would you prefer them in dialtone-icons/dist or dialtone-icons/android/dist ?

Hmm you know what you're right, this is more similar to the way the other icons are built. Let's just leave it the way it is for now and we can change it on our side if necessary. I wouldn't want to change one without changing the other.

braddialpad (approved):

PR #666, opened 2025-03-11
ci(release-notification): NO-JIRA notify on stable releases only

juliodialpad:
# Update release notification to notify on stable releases only

Obligatory GIF

- [x] CI

## :book: Description

- Limited the tags the workflow listens to.

## :bulb: Context

Having the release messages about alpha/beta/rebrand releases is noisy and don't make a lot of sense.

## :pencil: Checklist

For all PRs:

## :link: Sources

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags

### DIFF
--- a/.github/workflows/release-notification.yml
+++ b/.github/workflows/release-notification.yml
@@ -3,7 +3,7 @@ name: Release notification
 on:
   push:
     tags:
-      - dialtone/v*
+      - 'dialtone/v[0-9]+.[0-9]+.[0-9]+'
 
 env:
   RELEASE_URL: ${{ format('https://github.com/dialpad/dialtone/releases/tag/{0}', github.ref_name) }}


.github/workflows/release-notification.yml
```diff
@@ -3,7 +3,7 @@ name: Release notification
 on:
   push:
     tags:
-      - dialtone/v*
+      - 'dialtone/v[0-9]+.[0-9]+.[0-9]+'
```
braddialpad, line 6: Is this regex? If so you'll need to escape the dot right?

juliodialpad: It is glob pattern, the reference link with the examples is at the bottom of the PR description.

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags

braddialpad: Alright, yeah it's fine then

braddialpad (approved):

PR #653, opened 2025-02-27
fix(message-input): DP-131613 fix pasting links when no rich-text

bianca-artola-dialpad:
# [fix(message-input): DP-131613 Fix pasting links when no rich-text](https://github.com/dialpad/dialtone/commit/e993dec8b2684f0e88c81d214fb50e56da38c7f3)

## Obligatory GIF (super important!)

Obligatory GIF

## :hammer_and_wrench: Type Of Change

These types will increment the version number on release:

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DP-131613

## :book: Description

Clients reported an issue where, due to certain settings in Microsoft Edge, links were not being copied correctly when pasted into the new message input field.

Previous Behavior:
[View video](https://github.com/user-attachments/assets/db9abe15-e2b7-42a7-aeae-b42e1fc467ca)

This update resolves the issue, allowing users to copy and paste these links without any problems. Notably, this problem only affected the message input field when the link extension was disabled. When the link extension was enabled, the issue did not occur.

Updated Behavior (Fix Applied):
[View Screenshot](https://github.com/user-attachments/assets/ef0729cb-d527-4640-bd5a-b9bebd8f96ab)

PR #674, opened 2025-03-19
fix(message-input): DP-134543 fix pasting behavior

bianca-artola-dialpad:
# fix(message-input): DP-134543 fix pasting behavior

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :bulb: Context

We are receiving reports saying that the paste functionality is not working fine on the message input so we are planning to revert the chnages that were made as part of [this pull request](https://github.com/dialpad/dialtone/pull/653) and for fixing the original issue we'll start using on product side the link extension.

### DIFF
--- a/packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.stories.js
+++ b/packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.stories.js
@@ -155,6 +155,5 @@ export const WithCustomExtensions = {
     allowUnderline: false,
     allowCode: false,
     allowCodeblock: false,
-    link: false,
   },
 };
--- a/packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.stories.js
+++ b/packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.stories.js
@@ -155,5 +155,21 @@ export const WithCustomExtensions = {
     allowUnderline: false,
     allowCode: false,
     allowCodeblock: false,
+    link: false
+  },
+};
+
+export const OnlyLinkExtension = {
+  render: (argsData) => createRenderConfig(DtRichTextEditor, DtRichTextEditorDefaultTemplate, argsData),
+  args: {
+    allowBlockquote: false,
+    allowBold: false,
+    allowBulletList: false,
+    allowItalic: false,
+    allowStrike: false,
+    allowUnderline: false,
+    allowCode: false,
+    allowCodeblock: false,
+    link: true
   },
 };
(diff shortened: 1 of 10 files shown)


bianca-artola-dialpad:
Previews are not being updated with the latest changes, pushed a commit to retriger

braddialpad:
Hey regarding your comment in the chat:

Screenshot 2025-03-20 at 12 20 18 AM

Are you able to inspect the clipboard data when copying a link from edge like this? what is the metadata that causes it to be pasted as a named link rather than the address? If we can figure this out we can likely fix this problem.

bianca-artola-dialpad:
> Hey regarding your comment in the chat:
> 
> Screenshot 2025-03-20 at 12 20 18 AM
> 
> Are you able to inspect the clipboard data when copying a link from edge like this? what is the metadata that causes it to be pasted as a named link rather than the address? If we can figure this out we can likely fix this problem.

yeah good point. I added these logs:

```
      console.log(event.clipboardData);
      console.log("Text:", event.clipboardData.getData("text/plain")); // Plain text
      console.log("HTML:", event.clipboardData.getData("text/html"));  // HTML formatted text (if available)
      console.log("Files:", event.clipboardData.files);               // List of files (if any)
```

This is what happened when pasting an edge link:
Screenshot 2025-03-20 at 9 32 01 AM

vs pasting a link from any other place:
Screenshot 2025-03-20 at 9 32 06 AM


but i don't think we can do it something with that as we also get text/html when pasting for example a piece of code. 

I have another proposal that's a patch fix, but should fix both issues for now and will give us more time to work on this feature that will be a great thing to have. 
The original issue was with edge links -> we can identify those links with a regex and also we can also do the special handlePaste when the text/html section from the clipboardData is not null which we saw is the case for edge. Then, for all the other cases, we can just return and go into the normal paste behavior (in that way, we will avoid the new issues with pasting that were generated after my changes, like losing formating or not working okey for pasting code). I'm thinking on something like this:

```
handlePaste: (_, event) => {
            // When having link and customLink props we should maintain default paste behavior
            if (!this.link && !this.customLink) {
              const regex = /^https?:\/\//;
              const pastedContent = event.clipboardData.getData('text');

              if (!regex.test(pastedContent) || !event.clipboardData.getData('text/html')) {
                console.log('No URL or not HTML');
                return false;
              }

              this.editor.chain().focus().insertContent(pastedContent).run();
              return true; // Prevent the default paste behavior
            }

            return false; // Allow the default paste behavior
          },
```

I know it's no clean but it will give us more time to launch the link into the input with all the cases covered and tested.
I will raise a new pr with this small change to fix both issues and then I will continue thinking on how to fix this

braddialpad (approved):

braddialpad (changes requested):
Let's hold off on merging this one for now.

PR #675, opened 2025-03-19
feat: DLT-2419 hideEdges + watcher on active page prop

sanjeevdialpad:
# feat: hideEdges for lazy loading pagination and watcher on active page prop

test

test

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2419

## :book: Description

activePage prop is not reactive since its assigned to data on init, so we cannot control the pagination from outside this component which is needed sometimes.

Sometimes you may need to hide start and end page number buttons when moving in between.
This prop will be used to hide the first and last page buttons when not near the edges.
This is useful when your backend does not support offset and you can only use cursor based pagination.

## :pencil: Checklist

For all PRs:

For all Vue changes:

### DIFF
--- a/packages/dialtone-vue2/components/pagination/pagination.vue
+++ b/packages/dialtone-vue2/components/pagination/pagination.vue
@@ -187,32 +187,51 @@ export default {
         return this.range(1, this.totalPages);
       }
 
-      const start = this.maxVisible - 1;
-      const end = this.totalPages - start + 1;
+      let start = this.maxVisible - 1;
+      let end = this.totalPages - start + 1;
+
+      // if hideEdges is true, modify the start and
+      // end to account for the hidden pages
+      if (this.hideEdges) {
+        start = start + 1;
+        end = end - 1;
+      }
 
       if (this.currentPage < start) {
-        if (this.hideEdges) {
-          return [...this.range(1, start + 1), '...'];
+        const pages = [...this.range(1, start), '...'];
+        if (!this.hideEdges) {
+          // add last page to the end
+          pages.push(this.totalPages);
         }
-        return [...this.range(1, start), '...', this.totalPages];
+        return pages;
       }
 
       if (this.currentPage > end) {
-        if (this.hideEdges) {
-          return ['...', ...this.range(end - 1, this.totalPages)];
+        const pages = ['...', ...this.range(end, this.totalPages)];
+        if (!this.hideEdges) {
+          // add first page to the beginning
+          pages.unshift(1);
         }
-        return [1, '...', ...this.range(end, this.totalPages)];
+        return pages;
       }
 
       // rounding to the nearest odd according to the maxlength to always show the page number in the middle.
       const total = this.maxVisible - (3 - this.maxVisible % 2);
       const centerIndex = Math.floor(total / 2);
-      const left = this.currentPage - centerIndex;
-      const right = this.currentPage + centerIndex;
+      let left = this.currentPage - centerIndex;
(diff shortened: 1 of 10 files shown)


packages/dialtone-vue2/components/pagination/pagination_variants.story.vue
```diff
@@ -38,6 +38,20 @@
...
+    <div class="d-m32">
+      <p class="d-my16 d-fs-200 d-fw-bold">
+        Hide edges
+      </p>
+      <dt-pagination
+        :total-pages="10"
+        :active-page="5"
+        :aria-label="'pagination with separator on both sides'"
+        :prev-aria-label="'previous'"
+        :next-aria-label="'next'"
```
braddialpad: double quoting like this technically works but is kinda weird, just remove the `:`

```
aria-label="pagination with separator on both sides"
prev-aria-label="previous"
...
```

packages/dialtone-vue2/components/pagination/pagination.vue
```diff
@@ -142,6 +142,16 @@ export default {
...
+
+    /**
+     * Sometimes you may need to hide start and end page number buttons when moving in between.
+     * This prop will be used to hide the first and last page buttons when not near the edges.
+     * This is useful when your backend does not support offset and you can only use cursor based pagination.
+     */
+    hideEdges: {
+      type: Boolean,
+      default: false,
+    },
```
braddialpad, line 154: Should add a test for the hide edges case.

packages/dialtone-vue2/components/pagination/pagination.vue
```diff
@@ -181,10 +191,16 @@ export default {
       const end = this.totalPages - start + 1;
 
       if (this.currentPage < start) {
+        if (this.hideEdges) {
+          return [...this.range(1, start + 1), '...'];
```
braddialpad: instead of modifying start at this level can we change the variable assignment? Add a comment because this is a bit confusing.

```
let start = this.maxVisible;
let end = this.totalPages - start;

if (!this.hideEdges) {
  start = start - 1;
  end = end + 1;
}

File: packages/dialtone-vue2/components/pagination/pagination.vue [hunk 4]
```diff
@@ -181,10 +191,16 @@ export default {
...
+        }
         return [...this.range(1, start), '...', this.totalPages];
       }
 
       if (this.currentPage > end) {
+        if (this.hideEdges) {
+          return ['...', ...this.range(end - 1, this.totalPages)];
+        }
         return [1, '...', ...this.range(end, this.totalPages)];
```
braddialpad: Build the array instead of returning two nearly identical cases

```
pages = ['...', ...this.range(end, this.totalPages)];
if (!this.hideEdges) {
  pages.unshift(1);
}
return pages;
```

braddialpad (changes requested):

packages/dialtone-vue2/components/pagination/pagination.vue
sanjeevdialpad: added

sanjeevdialpad:
@braddialpad made all the changes!

braddialpad (approved):
Looks great thanks!

PR #676, opened 2025-03-20
fix(rich-text-editor): DP-134543 patchfix paste

bianca-artola-dialpad:
# fix(rich-text-editor): DP-134543 patchfix paste

Obligatory GIF

- [x] Fix

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DP-134569
https://dialpad.atlassian.net/browse/DP-134543

## :book: Description

The initial issue that was reported was an issue with links on Edge. With some investigation, we realized that when copying and pasting links from edge brings some html data inside that makes links not working as the other ones. 

**Technical details**

I added these logs to see the behavior:
```
      console.log(event.clipboardData);
      console.log("Text:", event.clipboardData.getData("text/plain")); // Plain text
      console.log("HTML:", event.clipboardData.getData("text/html"));  // HTML formatted text (if available)
      console.log("Files:", event.clipboardData.files);               // List of files (if any)
```

This is what happened when pasting an edge link:
Screenshot 2025-03-20 at 9 32 01 AM

vs pasting a link from any other place:
Screenshot 2025-03-20 at 9 32 06 AM

but i don't think we can do something with that as we also get text/html when pasting for example a piece of code. 

The original issue was with edge links -> we can identify those links with a **regex** and also we can also do the special handlePaste when the text/html section from the clipboardData is not null which we saw is the case for edge. Then, for all the other cases, we can just return and go into the normal paste behavior (in that way, we will avoid the new issues with pasting that were generated after my changes, like losing formatting or not working okey for pasting code).

I know it's no clean but it will give us more time to launch the link into the input with all the cases covered and tested.

## :bulb: Context
We are receiving reports saying that the paste functionality is not working fine on the message input; this was caused by https://github.com/dialpad/dialtone/pull/653. We initially think on reverting those changes and using the link extension instead, but as we noticed some unexpected behavior with it I think adding this patch fix will be the easiest and most quick patch fix we can do to fix both errors. 

Link PR for future: https://github.com/dialpad/dialtone/pull/674

Let me know what you think team!

### DIFF
--- a/packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
+++ b/packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
@@ -671,7 +671,25 @@ export default {
           handlePaste: (_, event) => {
             // When having link and customLink props we should maintain default paste behavior
             if (!this.link && !this.customLink) {
+              const regex = /^https?:\/\//;
+
+              if (!event?.clipboardData) {
+                return false;
+              }
               const pastedContent = event.clipboardData.getData('text');
+
+              // Check if the pasted content is a valid URL (starting with http:// or https://)
+              // If it's not a URL, allow the default paste behavior
+              if (!regex.test(pastedContent)) {
+                return false;
+              }
+
+              // If `text/html` is missing from clipboard data, it's a plain link
+              // In this case, allow the default paste behavior
+              if (!event.clipboardData.getData('text/html')) {
+                return false;
+              }
+
               this.editor.chain().focus().insertContent(pastedContent).run();
               return true; // Prevent the default paste behavior
             }
--- a/packages/dialtone-vue3/components/rich_text_editor/rich_text_editor.vue
+++ b/packages/dialtone-vue3/components/rich_text_editor/rich_text_editor.vue
@@ -673,7 +673,25 @@ export default {
           handlePaste: (_, event) => {
             // When having link and customLink props we should maintain default paste behavior
             if (!this.link && !this.customLink) {
+              const regex = /^https?:\/\//;
+
+              if (!event?.clipboardData) {
+                return false;
+              }
               const pastedContent = event.clipboardData.getData('text');
+
+              // Check if the pasted content is a valid URL (starting with http:// or https://)
+              // If it's not a URL, allow the default paste behavior
+              if (!regex.test(pastedContent)) {
+                return false;
+              }
+
+              // If `text/html` is missing from clipboard data, it's a plain link
+              // In this case, allow the default paste behavior
+              if (!event.clipboardData.getData('text/html')) {
+                return false;
+              }
+
               this.editor.chain().focus().insertContent(pastedContent).run();
               return true; // Prevent the default paste behavior
             }


packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
```diff
@@ -671,7 +671,25 @@ export default {
...
+              if (!regex.test(pastedContent)) {
+                return false;
+              }
+
+              // If `text/html` is missing from clipboard data, it's a plain link
+              // In this case, allow the default paste behavior
+              if (!event.clipboardData.getData('text/html')) {
+                return false;
+              }
+
```
braddialpad, line 692: I'm actually wondering if we should just ALWAYS take the text property from clipboard when our input is running in non-rich-text mode. Anyways that's a bit more dangerous so we can do this first.

braddialpad (approved):
I think this should work well

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
```diff
@@ -671,7 +671,25 @@ export default {
           handlePaste: (_, event) => {
             // When having link and customLink props we should maintain default paste behavior
             if (!this.link && !this.customLink) {
+              const regex = /^https?:\/\//;
```
braddialpad, line 674: actually just one additional question.. how will this behave if there is a link pasted alongside text.. for example:

hello [Google](www.google.com) test123

braddialpad: Just tried it.. pastes as text.. which I suppose is fine since the input doesn't support actual links

bianca-artola-dialpad: oh yeah when copying pasting from here it pastes as text, but tested with edge and it pasted the link 

https://github.com/user-attachments/assets/85071356-ee10-41ae-a640-237ea35949f6


I guess this is okey? so many different examples I'm confused lol

braddialpad: yeah no, I was finding it confusing too but it's a totally different case.

When copying www.google.com from edge it's like this:

```
text copied: www.google.com

in paste event:
text: www.google.com
html: <a href=www.google.com>Google</a>
```

When copying a link like I posted above it would be something above

```
text copied: hello Google test123

in paste event:
text: hello Google test123
html: hello <a href=www.google.com>Google</a> test123
```

in the case from edge it actually CHANGES the text you have copied in the html to be something different than what you copied. In the second case it at least is still pasting exactly what you copied (without the link of course).

braddialpad:
@bianca-artola-dialpad Do you know what the plan is with this fix, if it's going to be a prod CP we'll have to handle the release a bit differently.

bianca-artola-dialpad:
> @bianca-artola-dialpad Do you know what the plan is with this fix, if it's going to be a prod CP we'll have to handle the release a bit differently.

Waiting for Andrea response. Will update here

bianca-artola-dialpad:
> > @bianca-artola-dialpad Do you know what the plan is with this fix, if it's going to be a prod CP we'll have to handle the release a bit differently.
> 
> Waiting for Andrea response. Will update here

She said we can go with the regular flow for release, so if we deploy a version and update it on firespotter before Tuesday should be okay. I'll merge this, thank you Brad

braddialpad:
> She said we can go with the regular flow for release, so if we deploy a version and update it on firespotter before Tuesday should be okay. I'll merge this, thank you Brad

thanks! let's make sure to get some good QA testing on it so we don't continue creating regressions haha.

PR #683, opened 2025-03-26
ci: NO-JIRA refactor publish workflows

juliodialpad:
# Refactor publish workflows

Obligatory GIF

- [x] CI

## :book: Description

- Refactor publish workflows into separate files
- Fix a build issue due to the recent addition of the android dialtone-icons package being built on the main dialtone-icons build.

## :bulb: Context

- While I was working on publishing a beta version of tokens realized that android and iOS publishing should be running from a separate workflows as they have different running conditions, such as production only release (iOS) and only tokens/icons (Android). I think with this changes, it'd be more maintainable and it's easier to read and follow up what's going on with those workflows. 

- Building the android icons is not necessary for our main build as we're not using them, separated that built to `build:android` tasks to make it clear you're building android specific projects that might require external dependencies like JAVA and the android SDK

## :pencil: Checklist

For all PRs:

### DIFF
--- a/.github/workflows/publish-android.yml
+++ b/.github/workflows/publish-android.yml
@@ -1,0 +1,120 @@
+name: Publish Android
+
+on:
+  workflow_dispatch:
+    inputs:
+      package:
+        description: 'Select the package you want to publish'
+        required: true
+        default: 'dialtone-tokens'
+        type: 'choice'
+        options:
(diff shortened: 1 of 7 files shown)


braddialpad (approved):
Definitely makes sense to do this, thanks!

PR #688, opened 2025-04-01
fix: DLT-2398 synchronize class attrs on vue 3 components

braddialpad:
# fix: synchronize class attrs on vue 3 components

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2398

## :book: Description

Migrated our class attributes to match the way it works in vue 2 if config option is enabled.

Vue 3 only.

## :bulb: Context

Due to the breaking change [INSTANCE_ATTRS_CLASS_STYLE](https://v3-migration.vuejs.org/breaking-changes/attrs-includes-class-style.html) in vue 3, the class and style attributes are not applied to the same elements they were in vue 2.

Any component that uses `inheritAttrs: false` will now include class and style attributes in `$attrs` unlike vue 2 which always applied them to the root element. This means in order to have our components behave the same in vue 2 and vue 3 we need to make a small modification to directly apply the class to the root element in the case of inheritAttrs: true.

This could potentially be a breaking change for applications already using the vue 3 library and applying class attributes to inheritAttrs components so I included this behaviour as a configuration option that you can set on init more for migration purposes.

Already tested on the vue 3 migration branch and it does seem to work.

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

Apply config option in ubervoice vue 3 migration.

## :link: Sources

https://v3-migration.vuejs.org/breaking-changes/attrs-includes-class-style.html


juliodialpad (commented):
As you're removing the `class` and `style` attributes doesn't that mean that you should include on all those components a `:style="$attrs.style"` attribute? even if it's empty, otherwise, if any of those components have a style, it'd be ignored, right?

braddialpad:
> As you're removing the `class` and `style` attributes doesn't that mean that you should include on all those components a `:style="$attrs.style"` attribute? even if it's empty, otherwise, if any of those components have a style, it'd be ignored, right?

Yeah true, if there was one. I didn't do this because we don't generally use inline styles and in fact consider it a bad practice to do so.

juliodialpad:
> > As you're removing the `class` and `style` attributes doesn't that mean that you should include on all those components a `:style="$attrs.style"` attribute? even if it's empty, otherwise, if any of those components have a style, it'd be ignored, right?
> 
> Yeah true, if there was one. I didn't do this because we don't generally use inline styles and in fact consider it a bad practice to do so.

I understand, but as we don't want to do breaking changes, we should include them as they might cause issues if someone used it on product.

Or at least we should put a comment somewhere saying that we remove style attribute on purpose on those components.

braddialpad:
> > > As you're removing the `class` and `style` attributes doesn't that mean that you should include on all those components a `:style="$attrs.style"` attribute? even if it's empty, otherwise, if any of those components have a style, it'd be ignored, right?
> > 
> > Yeah true, if there was one. I didn't do this because we don't generally use inline styles and in fact consider it a bad practice to do so.
> 
> I understand, but as we don't want to do breaking changes, we should include them as they might cause issues if someone used it on product.
> 
> Or at least we should put a comment somewhere saying that we remove style attribute on purpose on those components.

Yeah doesn't hurt to include them I guess..

braddialpad:
Added style as requested.

juliodialpad (approved):
Looking good, thanks!

packages/dialtone-vue3/components/avatar/avatar.vue
```diff
@@ -3,6 +3,7 @@
     :is="clickable ? 'button' : 'div'"
     :id="id"
     :class="avatarClasses"
+    :style="$attrs.style"
```
noellelevydialpad, line 6: why does this file have inheritAttrs false? If it didn't have that what you are doing would be default behavior (plus some other attrs could be inherited)

packages/dialtone-vue3/recipes/leftbar/callbox/callbox.vue
```diff
@@ -1,7 +1,8 @@
 <template>
   <div
     data-qa="dt-recipe-callbox"
-    class="d-recipe-callbox"
+    :class="[$attrs.class, 'd-recipe-callbox']"
```
noellelevydialpad, line 4: also would be default behavior without inheritattrs false

packages/dialtone-vue3/recipes/leftbar/contact_centers_row/contact_centers_row.vue
```diff
@@ -168,6 +170,7 @@ export default {
   data () {
     return {
       labelWidth: 'auto',
+      removeClassStyleAttrs,
```
noellelevydialpad: this is minor, but vue will try to make this reactive in `data`. its more of a `methods` member

packages/dialtone-vue3/common/utils/index.js
```diff
@@ -202,6 +205,26 @@ export const extractVueListeners = (attrs) => {
   return Object.fromEntries(listeners);
 };
 
+/*
+  Only will apply changes if the config option configVue2StyleClassAttrs is set to true. It is false by default.
```
noellelevydialpad, line 226: I would argue this PR is still a breaking change, because if you do not opt into enabling this, the class and style attrs are still added to the root element.

noellelevydialpad (changes requested):
I think the util function should be used in a computed that splits out the style and class attrs if enabled. As it stands, if you don't opt into emulating the vue2 behavior, the class and style gets added in both places.

packages/dialtone-vue3/components/avatar/avatar.vue
braddialpad: 100% agree, I will do this in a separate PR though since i'll have to do it on vue 2 and vue 3.

packages/dialtone-vue3/common/utils/index.js
braddialpad: I'm not sure that's true because this is only used on components where inheritAttrs: false. I will do some tests to confirm that though..

packages/dialtone-vue3/recipes/leftbar/contact_centers_row/contact_centers_row.vue
braddialpad: good point... for some reason it never occurred to me to put an imported function into `methods:` even though that completely makes sense

packages/dialtone-vue3/common/utils/index.js
braddialpad: Oh nevermind I see what you mean, because we added

```
:class="$attrs.class"
:style="$attrs.style"
```

to the root for all cases... Yeah I'll fix that.

braddialpad:
Made requested fixes, will put the two I mentioned in another PR

noellelevydialpad (approved):

PR #696, opened 2025-04-07
feat(localization): DLT-2452 setup i18n

juliodialpad:
# Setup i18n

Obligatory GIF

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2452

## :book: Description

- Set up i18n library
- Test localization on the Popover component
- Added @dialpad:registry config to all the workflows that need it to be able to install private `@dialpad/i18n-services` and `@dialpad/i18n-vue2` packages

## :bulb: Context

- We're starting the efforts to get Dialtone-vue to be i18n compatible

## :pencil: Checklist

For all PRs:

For all Vue changes:

If new component:

- [x] I am exporting new plugin from the index.js in the root (either `packages/dialtone-vue2` or `packages/dialtone-vue3`).

## :crystal_ball: Next Steps

- Incrementally test on other components

Screenshot 2025-04-06 at 11 24 31 p m
Screenshot 2025-04-06 at 11 23 56 p m
Screenshot 2025-04-06 at 11 24 17 p m

## Sources:

[Dialpad i18n documentation](https://github.com/dialpad/web-clients/blob/main/packages/i18n-vue2/README.md)


packages/dialtone-vue2/components/popover/popover_header_footer.vue
```diff
@@ -28,7 +28,7 @@
       importance="outlined"
       kind="muted"
       circle
-      v-bind="closeButtonProps"
+      v-bind="$ta('CLOSE_BUTTON', { component: 'popover' })"
```
ninamarina: Do we still want to support other button props than ariaLabel?
Is the `component` key used here?

juliodialpad: > Do we still want to support other button props than ariaLabel?

I don't think that we used that closeButtonProps for anything besides passing `aria-label` but we can keep it in case it's needed.

> Is the component key used here?

That was a leftover, will remove it, thanks for catching that!

packages/dialtone-vue2/localization/dp-DP.ftl
braddialpad, line 1: So I guess all our localization files would go out as part of the bundle right? Does it only load languages you actually need to use AKA tree shaking? Just wondering how this needs to be managed on the consumer side.

packages/dialtone-vue2/components/popover/popover_default.story.vue
```diff
@@ -106,5 +142,13 @@ export default {
...
     },
   },
+
+  methods: {
+    $t,
+
+    async setLocale (locale) {
+      await setI18N({ preferredLocale: locale });
+    },
+  },
```
braddialpad: How would a consumer set the language within dialtone?

braddialpad (commented):
Nice! This is exciting!

packages/dialtone-vue2/localization/dp-DP.ftl
juliodialpad: We'll need to check if the i18n is tree-shaking capable, as we're using a function coming from their `i18n-services` package to `bundle` the translations. But I think it is not capable of tree shaking at the moment.

packages/dialtone-vue2/components/popover/popover_default.story.vue
juliodialpad: I will need to test a bit more once this is merged, but I think it'd change automatically if the locale changes on the client side, you can read more in the following link:

https://github.com/dialpad/web-clients/blob/main/packages/i18n-vue2/README.md#users-locale-selection-criteria

packages/dialtone-vue2/components/popover/popover_default.story.vue
braddialpad, line 1: Before merging can we make this an alternate example outside of popover_default? or make some sort of new translation test component? IDK up to you, just don't want it to replace the default popover example with this

packages/dialtone-vue2/components/popover/popover_header_footer.vue
braddialpad: I think we should actually remove closeButtonProps entirely. The other purpose of it would be to set any other related accessibility attributes or other HTML attributes, but we can now set what we need on the button itself directly within the component. closeButtonProps actually gives developer TOO much flexibility. We want things like this to be more standardized by the design system.

braddialpad:
Need to transfer to vue 3 before merging as well?

packages/dialtone-vue2/components/popover/popover_default.story.vue
juliodialpad: For sure, I created it on popover for testing purposes only, I'll check if I can add it to storybook globally or just add another example if it's a lot of work, will create another PR to merge into this one so it's easier to review

juliodialpad:
> Need to transfer to vue 3 before merging as well?

I'll create another PR with the vue 3 version of this after removing the popover example to avoid making this PR too complex to review.

braddialpad (approved):

PR #703, opened 2025-04-09
ci(i18n): DLT-2451 setup i18n workflows

juliodialpad:
# Setup i18n workflows

Obligatory GIF

- [x] CI

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2451

## :book: Description

- Added workflows to automatically push and pull translations from smartling

> Note: Branched out of https://github.com/dialpad/dialtone/tree/dlt-2452-setup-i18n to be able to test the actions.

## :bulb: Context

We're setting up i18n on dialtone-vue

## :pencil: Checklist

For all PRs:

## :link: Sources

- https://docs.google.com/document/d/1RZVbFwDqZ4qUQ3C65y9_iUAn6KZ4yfuTPyLIx8qDeyQ/edit?tab=t.0 see step 10.

### DIFF
--- a/.github/actions/setup-environment/action.yml
+++ b/.github/actions/setup-environment/action.yml
@@ -1,0 +1,22 @@
+name: Setup Environment
+description: Common steps  set up the environment for tasks
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install pnpm
+      uses: pnpm/action-setup@v3
+
+    - name: Set up Node.js
+      uses: actions/setup-node@v4
+      with:
+        node-version-file: ".nvmrc"
+        cache: "pnpm"
+        registry-url: "https://npm.pkg.github.com"
+        scope: "dialpad"
+      env:
+        NODE_AUTH_TOKEN: ${{ github.token }}
+
+    - name: Install dependencies
+      shell: bash
+      run: pnpm install --frozen-lockfile
--- a/.github/workflows/pull-translations.yml
+++ b/.github/workflows/pull-translations.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - staging
-      - setup-i18n-actions
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/pull-translations.yml
+++ b/.github/workflows/pull-translations.yml
@@ -1,0 +1,115 @@
+name: Pull from translation service
+
+on:
+  push:
+    branches:
+      - staging
+      - setup-i18n-actions
+  workflow_dispatch:
+
+jobs:
+  pull-translations:
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        shell: bash -o pipefail {0}
+        working-directory: ./packages/dialtone-vue2
+
+    env:
+      SMARTLING_BASE_URL: "https://api.smartling.com"
+      SMARTLING_AUTH_PATH: "/auth-api/v2/authenticate"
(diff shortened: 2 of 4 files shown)


braddialpad (approved):
Did you get this from WCF team or did you write it yourself? This seems like something should be standardized as part of the translation library or at least some sort of packaged github action or script.

How does this work in terms of timing? what if we release a version but the translations have not been completed by smartling yet?

juliodialpad:
> Did you get this from WCF team or did you write it yourself? This seems like something should be standardized as part of the translation library or at least some sort of packaged github action or script.

Took this from WCF team, made a minor adjustments to make it work with our versions and to be similar to how our other workflows work. I agree this could be an action written by the WCF team, that would simplify the usage of it, but for now, I think this will work correctly.

> How does this work in terms of timing? what if we release a version but the translations have not been completed by smartling yet?

That's a good point...

The way Smartling works is:

1. We update the en-US.ftl file locally, that's the source of truth.
2. Once that file reaches staging, it'd be pushed to Smartling.
3. Then you need to go to Smartling and create a "Job" to start the translation of the phrases.
4. Once the translations are ready they'll be available for download
5. On every push to staging, manually or whenever we want to set it up; an action will try to pull the translations from Smartling, create a commit and a PR with the updated files.
6. We need to manually review and merge the PR to make sure everything is OK.
7. The translations are now available in Dialtone-vue so we can publish a new version containing all the translations.

The process is semi-automated.

I don't think Dialtone is going to use Smartling a lot, we'll mostly translating static content so we're going to be using smartling a lot at the beginning, but it become rarely used after a few iterations. 

Anyway I think if the translation is missing, it fallbacks to the default locale `en-US` so no errors expected on that case. 
Am I right? @dlovero-dialpad

dlovero-dialpad:
> > Did you get this from WCF team or did you write it yourself? This seems like something should be standardized as part of the translation library or at least some sort of packaged github action or script.
> 
> Took this from WCF team, made a minor adjustments to make it work with our versions and to be similar to how our other workflows work. I agree this could be an action written by the WCF team, that would simplify the usage of it, but for now, I think this will work correctly.
> 
> > How does this work in terms of timing? what if we release a version but the translations have not been completed by smartling yet?
> 
> That's a good point...
> 
> The way Smartling works is:
> 
> 1. We update the en-US.ftl file locally, that's the source of truth.
> 2. Once that file reaches staging, it'd be pushed to Smartling.
> 3. Then you need to go to Smartling and create a "Job" to start the translation of the phrases.
> 4. Once the translations are ready they'll be available for download
> 5. On every push to staging, manually or whenever we want to set it up; an action will try to pull the translations from Smartling, create a commit and a PR with the updated files.
> 6. We need to manually review and merge the PR to make sure everything is OK.
> 7. The translations are now available in Dialtone-vue so we can publish a new version containing all the translations.
> 
> The process is semi-automated.
> 
> I don't think Dialtone is going to use Smartling a lot, we'll mostly translating static content so we're going to be using smartling a lot at the beginning, but it become rarely used after a few iterations.
> 
> Anyway I think if the translation is missing, it fallbacks to the default locale `en-US` so no errors expected on that case. Am I right? @dlovero-dialpad

That's correct, it will fall back to en-US or the one you specified as fallback on the API

EDIT: All the other things you wrote are also correct :-]

braddialpad:
Sounds good thanks, probably won't be a huge issue for us.

PR #711, opened 2025-04-15
fix: NO-JIRA vue 3 v-model corrections

braddialpad:
# fix: NO-JIRA vue 3 v-model corrections

Obligatory GIF

- [x] Fix

## :book: Description

Migrated input components to use the correct conventions for v-model as specified in https://v3-migration.vuejs.org/breaking-changes/v-model.html#_3-x-syntax

- on input / select components modelValue maps to "value"
- on checkbox / radio / toggle components modelValue maps to "checked"

## :bulb: Context

Noticed during vue 3 migration of Dialpad that the way we were using v-model in vue 3 was not the standard/recommended way of using it. This will be a possible breaking change for vue 3 consumers if they were referencing value / checked props directly rather than using v-model so I've written a blog post for it.

## :pencil: Checklist

For all PRs:

For all Vue changes:

## :crystal_ball: Next Steps

Update in vue 3 migration branch

## :link: Sources

https://v3-migration.vuejs.org/breaking-changes/v-model.html#_3-x-syntax

### DIFF
--- a/packages/dialtone-vue3/components/radio_group/radio_group.vue
+++ b/packages/dialtone-vue3/components/radio_group/radio_group.vue
@@ -16,7 +16,7 @@ export default {
      * A provided value for the radio group
      * @model value
      */
-    value: {
+    modelValue: {
       type: [String, Number],
       default: '',
     },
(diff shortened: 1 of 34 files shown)


packages/dialtone-vue3/components/radio_group/radio_group.vue
```diff
@@ -16,7 +16,7 @@ export default {
      * A provided value for the radio group
      * @model value
      */
-    value: {
+    modelValue: {
```
ninamarina, line 19: Is there a reason why in checkbox group we are using `selectedValues` instead of `modelValue` or `value`?

braddialpad: I also asked this question... no I think it is totally wrong but I don't want to fix it in this PR because then I'd also have to break vue 2.

ninamarina: oh, that makes sense

ninamarina (approved):

PR #724, opened 2025-04-21
feat(localization): DLT-2461 refactor components

juliodialpad:
# Refactor components to use localization

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2461

## :book: Description

- Migrated first set of componentes, will break this into multiple PRs to ease the review process as there might be a lot of changes happening.

## :bulb: Context

We're implementing localization on Dialtone.

## :pencil: Checklist

For all PRs:

For all Vue changes:


packages/dialtone-vue2/components/banner/banner.stories.js
juliodialpad, line 1: Separated banner from notice as they might have different props so it's easier to maintain them separately than overriding props here.

packages/dialtone-vue2/components/combobox/combobox.stories.js
```diff
@@ -3,14 +3,14 @@ import { createRenderConfig } from '@/common/storybook_utils';
 import DtCombobox from './combobox.vue';
 
 import DtComboboxDefaultTemplate from './combobox_default.story.vue';
-import { LABEL_SIZES } from './combobox_constants';
+import { COMBOBOX_LABEL_SIZES } from './combobox_constants';
```
juliodialpad, line 6: Renamed to match our specifications about prefixing the constants with the component name.

packages/dialtone-vue2/common/mixins/localization.js
braddialpad, line 1: Would this work better as a composable for vue 3?

packages/dialtone-vue2/components/combobox/combobox_default.story.vue
```diff
@@ -60,8 +60,8 @@
 </template>
 
 <script>
+import DtCombobox from './combobox.vue';
```
braddialpad, line 63: Why did we change to this style of import only here?

packages/dialtone-vue2/components/combobox/combobox_empty-list.vue
```diff
@@ -17,7 +17,7 @@
 </template>
 
 <script>
-import { DtListItem } from '../list_item';
+import { DtListItem } from '@/components/list_item';
```
braddialpad, line 20: Do we need to change these imports? I think we used relative imports in the first place because we had some weird cyclical dependency issue when using imports this way. I'm not against changing it if it works now because I think it is a more consistent way to import.

packages/dialtone-vue2/components/datepicker/modules/month-year-picker.vue
```diff
@@ -71,26 +71,26 @@
...
       as="nav"
+      class="d-datepicker__nav"
       direction="row"
       gap="200"
-      class="d-datepicker__nav"
     >
       <dt-tooltip
-        :message="nextMonthLabel"
+        :fallback-placements="['top-end', 'auto']"
+        :message="$t('NEXT_MONTH')"
```
braddialpad: Should we prefix this with component name DATEPICKER_NEXT_MONTH? or is it not necessary?

packages/dialtone-vue2/common/mixins/localization.js
juliodialpad: Yes, I did it as a mixin to be compatible with Vue 2.6, but will create a composable for Vue 3

packages/dialtone-vue2/components/combobox/combobox_default.story.vue
juliodialpad: I prefer to import local stuff with relative imports to know that the component is within the same directory.

But it's just a preference, the way we have storybook right now, would work with both styles

packages/dialtone-vue2/components/combobox/combobox_empty-list.vue
juliodialpad: The same comment as above, if the component is within the same directory, I use relative import, otherwise I prefer to use the '@/' prefix to know that it is external.

That helped me to fix a lot of issues while working on tree-shaking and recently with the dependency issues.

packages/dialtone-vue2/components/datepicker/modules/month-year-picker.vue
juliodialpad: Yes, thanks for pointing out that, we should follow the same rules as our constants.

braddialpad:
Some error happening in percy, guessing maybe that won't be fixed until the migration is fully complete?

juliodialpad:
> Some error happening in percy, guessing maybe that won't be fixed until the migration is fully complete?

Pretty strange, that component is not even using any localization, will look into it

juliodialpad:
> Some error happening in percy, guessing maybe that won't be fixed until the migration is fully complete?

It was failing on the base `i18n/brase-branch` branch, it is working correctly now, that's why there was a difference.

juliodialpad:
Merging this one, It's getting difficult to continue working on localization without this changes in.

braddialpad:
It's good, thanks!

PR #725, opened 2025-04-21
fix(checkbox, radio): NO-JIRA description slot not being reactive

ninamarina:
# fix(checkbox, radio): NO-JIRA description slot not being reactive

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket
No jira, we found out about this with a test failing in Firespotter: https://github.com/dialpad/firespotter/pull/56110

## :book: Description
Since `$slots` isn't reactive, it doesn't re render the component when passing the `description` slot. Moving the logic back inline instead of in a computed property makes it work again.

### DIFF
--- a/packages/dialtone-vue2/components/checkbox/checkbox.vue
+++ b/packages/dialtone-vue2/components/checkbox/checkbox.vue
@@ -25,12 +25,12 @@
       </div>
     </label>
     <div
-      v-if="hasDescriptionOrMessages"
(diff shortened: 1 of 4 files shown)


braddialpad (approved):
Oh yeah, makes sense because $slots isn't reactive. Good fix thanks!

PR #732, opened 2025-04-25
feat(editor): DP-115099 recipe editor support for focusing action bar buttons using arrow keys

anarravula-dialpad:
# Digital Email Users Would Prefer to have Focus Enabled on a single Top Action Bar Button on the Recipe Editor

Obligatory GIF

- [ ] Fix
- [ ] Feature
- [x] Performance Improvement
- [ ] Refactor

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DP-115099

## :book: Description

When users are using the email editor they should be able to press tabs to navigate from To Address, Subject, one action bar button and then to the recipe editors text area. However, currently when they press tab they go from To => Subject => Recipe Editor Actions (Formatting, Quick Replies, Bullet Points, etc) and then finally to the text area. Additionally, the arrow keys will wrap around the focus when they get to the end of the row of the buttons. When users focus out of the action bar and refocus, they get right back to where they were previously focused in the action bar.

## :bulb: Context

Purpose of the changes is to make the editor only have a single focusable top action bar button at a time. Users can however use the left and right arrow keys to shift the focus of an action bar button. 

Users were complaining about using too many tab presses to get to the text area to type

As a result of these changes users can focus into the text area in two button presses when they use tab to change focus instead of having to tab their way through the top action bar buttons

## :pencil: Checklist

For all PRs:

For all Vue changes:

https://github.com/user-attachments/assets/6e282e02-fc90-4867-908e-62cbce0adb2b


braddialpad:
Hi Aditya, thanks for the PR! Do you need an urgent fix here? If not I think we would prefer if the action bar was still focusable but instead of tabbing through each action bar item you would instead use arrow keys to navigate through each item, and pressing tab from here would change focus to the editor.

anarravula-dialpad:
> Hi Aditya, thanks for the PR! Do you need an urgent fix here? If not I think we would prefer if the action bar was still focusable but instead of tabbing through each action bar item you would instead use arrow keys to navigate through each item, and pressing tab from here would change focus to the editor.

@braddialpad I like your suggestion a lot! I proposed this to the omni channel team and they are onboard! I would like to clarify some of the behavior. Do you want me to remember where the last focus was on the action bar item when we refocus back. For example:

1. If a user initially lands on the action bar for focus, the focus is on the first item of the action bar. 
2. When the user presses right arrow they can focus on the second action bar item. 
3. When they press tab they go to the text input area. 
4. When they press shift tab their focus goes back to the action bar items

At this point do you want the focus to go back to the first action bar item or do you want the focus to go back to where they pressed tab from (second action bar item).

braddialpad:
> At this point do you want the focus to go back to the first action bar item or do you want the focus to go back to where they pressed tab from (second action bar item).

I think saving the last focus location is definitely a better experience!

anarravula-dialpad:
> > At this point do you want the focus to go back to the first action bar item or do you want the focus to go back to where they pressed tab from (second action bar item).
> 
> I think saving the last focus location is definitely a better experience!

@braddialpad Added support for shifting focus with wrapping around support on the action bar with arrow keys. It also remembers where the focus was last when we blur and focus next or previous by keep the last focused element's tabindex to 0 while leaving the rest of the buttons to -1. When the page loads the default button the focus lands on when the action bar is focused is the first button (quick replies). Updated the tests to test wrap around and shift right and left. Updated the video with the new behavior and updated the PR title and description. It is ready for you re-review.

braddialpad (commented):
Looks and functions really well. Couple of questions before I approve.

Currently the focus jumps to the editor after selecting an action bar button via pressing enter. Is this how we want it to behave?

Do we want it to be possible to focus the action bar by pressing "arrow up" on the top line of the editor? I only mention this because there are some cases where pressing "shift + tab" in the editor will make a change to the formatting rather than changing focus, such as when you have a bullet list. Not a blocker for this PR as it's quite an uncommon edge case, maybe a future improvement.

anarravula-dialpad:
> Looks and functions really well. Couple of questions before I approve.
> 
> Currently the focus jumps to the editor after selecting an action bar button via pressing enter. Is this how we want it to behave?
> 
> Do we want it to be possible to focus the action bar by pressing "arrow up" on the top line of the editor? I only mention this because there are some cases where pressing "shift + tab" in the editor will make a change to the formatting rather than changing focus, such as when you have a bullet list. Not a blocker for this PR as it's quite an uncommon edge case, maybe a future improvement.

@braddialpad 
1. I would like the behavior where the focus jumps to the editor after selecting the action bar button via pressing enter. It sort of gives users the familiar feel of gmail where the behavior is similar. I think this JIRA was initially opened because users were expecting the editor to behave similar to the email clients they are already used to. Ofc we aren't promising the exact same behavior but we would like to keep is fairly similar if we can.
2. Good catch! I haven't even considered bullet list and other special cases. I think it should definitely be a future improvement. I have create a JIRA to track and prioritize this: https://dialpad.atlassian.net/browse/DP-140444

braddialpad (approved):
great thanks!

PR #748, opened 2025-05-07
fix(ivr-node): NO-JIRA fix connector color in dark mode

braddialpad:
Just a tiny fix on ivr_node to make the connector color dark mode compatible.


ninamarina (approved):
thanks!

ninamarina:
I noticed the same should be done for node with top connector:

image

braddialpad:
> I noticed the same should be done for node with top connector:

Good find thanks, it was in the story.vue

PR #768, opened 2025-05-20
fix(hovercard): DP-142048 remove hovercard when anchor is removed from the DOM

ninamarina:
# fix(hovercard): DP-142048 remove hovercard when anchor is removed from the DOM 

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket
[DP-142048]

## :book: Description
The closing of the hovercard was done on `mouseleave`, but there are cases in which the event is not triggered but we want to close the hovercard as in the example from [DP-142048]. In this fix I added an observer to detect if the anchor is removed from the DOM, and in that case, set `hovercardOpen` to false.

## :pencil: Checklist

For all PRs:

For all Vue changes:

### Before

May-20-2025 16-34-45

### With this fix

May-20-2025 15-39-28

[DP-142048]: https://dialpad.atlassian.net/browse/DP-142048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DP-142048]: https://dialpad.atlassian.net/browse/DP-142048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dialtone-vue2/components/hovercard/hovercard.vue
+++ b/packages/dialtone-vue2/components/hovercard/hovercard.vue
@@ -1,6 +1,7 @@
 <template>
   <dt-popover
     :id="id"
+    ref="popover"
     :open="hovercardOpen"
     :placement="placement"
     :content-class="contentClass"
@@ -213,6 +214,8 @@ export default {
   data () {
     return {
       hovercardOpen: this.open,
+      anchorEl: null,
+      observer: null,
       inTimer: null,
       outTimer: null,
     };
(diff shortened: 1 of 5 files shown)


packages/dialtone-vue2/components/hovercard/hovercard.vue
```diff
@@ -228,6 +231,33 @@ export default {
     },
   },
 
+  mounted () {
+    this.$nextTick(() => {
+      this.anchorEl = this.$refs.popover?.$refs?.anchor;
```
braddialpad: This refers to the wrapper around the anchor element as opposed to the anchor itself. In the rightbar case you posted, it would probably still work, but what if the user passes a button into the slot and then hides the button using v-if, would it work in that case?

ninamarina: It won't work in that case, you are right. I was thinking about this case in which the route changes so the whole thing is removed from the DOM, but I can update the fix to consider that case

braddialpad (approved):
Thanks!

PR #774, opened 2025-05-25
feat(toggle): NO-JIRA add wrapperClass vue3 only

braddialpad:
# feat(toggle): add wrapperClass vue3 only

- [x] Feature

## :book: Description

add wrapperClass to toggle

## :bulb: Context

class attrs changed from vue 2 to vue 3, they now apply to the input element instead of the root, so we need a way to apply classes to the wrapper.

### DIFF
--- a/packages/dialtone-vue3/components/toggle/toggle.stories.js
+++ b/packages/dialtone-vue3/components/toggle/toggle.stories.js
@@ -12,6 +12,7 @@ export const argsData = {
   modelValue: false,
   onChange: action('change'),
   labelClass: 'd-mr6',
+  wrapperClass: '',
 };
 
 // Prop Controls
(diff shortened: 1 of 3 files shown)


juliodialpad (commented):
Why do we need a `wrapperClass` if we're already inheriting the class via `addClassStyleAttrs`?

braddialpad:
> Why do we need a `wrapperClass` if we're already inheriting the class via `addClassStyleAttrs`?

addClassStyleAttrs only works if you enable the global option configVue2StyleClassAttrs which makes vue 3 work like vue 2. It's good for migration purposes but not really something we'd want everyone to use going forward.

juliodialpad:
> addClassStyleAttrs only works if you enable the global option configVue2StyleClassAttrs which makes vue 3 work like vue 2. It's good for migration purposes but not really something we'd want everyone to use going forward.

Completely forgot about it was opt-in, that makes sense.

juliodialpad (approved):

PR #796, opened 2025-06-13
fix(rich-text-editor): DLT-2486 add pasteRichText prop

braddialpad:
# fix(rich-text-editor): DLT-2486 add pasteRichText prop

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2486

## :book: Description

- Added a new "pasteRichText" prop which if enabled will only paste plain text

- Added a bunch of new copy / paste related tests since we are constantly having so many copy paste issues with rich-text-editor.

- Fixed issues with whitespace using `preserveWhitespace: full` option

## :bulb: Context

The root of this problem is that the TipTap setContent command accepts both plain text as well as html strings and there is no way to tell it specifically whether it's one or the other. Even if you copy literal html tags from a code editor it will still parse them. When we are using the input in non-rich-text mode we don't want it to parse the HTML because it causes problems with formatting.

With rich text enabled this is not an issue because the formatting is handled by all the extensions we have enabled for rich text. When pasteRichText is enabled it will only ever paste literal text and not parse any tags solving this issue.

## :pencil: Checklist

For all PRs:

For all Vue changes:

## :crystal_ball: Next Steps

Validate in product


ninamarina:
I see it working correctly for `richText:false`, when `richText` is enabled I still see the pasted text in a single line:
image

Is this expected?

braddialpad:
> I see it working correctly for `richText:false`, when `richText` is enabled I still see the pasted text in a single line: image
> 
> Is this expected?

Thanks for finding that edge case, should be fixed now.

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
```diff
@@ -702,48 +711,48 @@ export default {
...
+            const transformedHtml = html
+              // Convert standalone br tags to hard breaks
+              .replace(/<br\s*\/?>/gi, '<br />')
+              // Preserve line breaks at end of paragraphs and divs
+              .replace(/(<\/(?:p|div)>)\s*\n/gi, '$1<br />')
+              // Convert newlines followed by text to br tags
+              .replace(/\n(?=\S)/g, '<br />');
+
+            // Then apply the original transformation
+            return transformedHtml.replace(/(<\/\w+>)((<br \/>)+)/g, '$2$3$1');
```
juliodialpad, line 755: This custom HTML transformations might cause issues in the future, but for now, they're apparently working correctly, let's just keep an eye on this.

juliodialpad (approved):
Looking good! The edge case reported by Nina now works correctly, couldn't find any other issue, thanks!

packages/dialtone-vue2/components/rich_text_editor/rich_text_editor.vue
braddialpad: Yeah... I would rather not have them but couldn't find any better way.

PR #811, opened 2025-06-23
feat(localization): temp revert i18n

braddialpad:
Temporarily reverting i18n so I can make a non-blocking release without it. Will unrevert right after the release.


juliodialpad:
It is not necessary, just opened a PR on firespotter with i18n working correctly.

braddialpad:
Closing this since tico found a proper fix

PR #812, opened 2025-06-25
fix(select): DP-147182 remove icon z-index

ninamarina:
# fix(select): DP-147182 remove icon z-index

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket
[DP-147182]

## :book: Description
The z-index for the icon in the Select component was set to 25, which makes it display on top of the overlay. Removing the z-index doesn't seem to break anything. 

[DP-147182]: https://dialpad.atlassian.net/browse/DP-147182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


braddialpad:
some weird flakiness with the visual tests...I'm just going to try re-running them.

ninamarina:
> some weird flakiness with the visual tests...I'm just going to try re-running them.

yeah, maybe I need to wait for those components to load

braddialpad:
> yeah, maybe I need to wait for those components to load

percy is supposed to be smart enough to do that on its own..

braddialpad (approved):
Change is reasonable, I think setting an explicit z-index here causes more problems than it solves, but I do feel like this could break some weird case on the product side. I'll approve but lets make sure we get visual tests passing at least.

ninamarina:
> Change is reasonable, I think setting an explicit z-index here causes more problems than it solves, but I do feel like this could break some weird case on the product side. I'll approve but lets make sure we get visual tests passing at least.

There they passed. Yes, that z-index is suspicious, I can't imagine why such a high z-index would be needed.

PR #815, opened 2025-06-30
docs(breakpoints): NO-JIRA improve documentation to include example

ninamarina:
General improvement of the [Breakpoints](https://dialtone.dialpad.com/deploy-previews/pr-815/utilities/responsive/breakpoints.html) section in the docs. Adds setup instructions for the `@dialpad/postcss-responsive-variations` plugin. Adds example of `postcss.config.js` so users can copy paste.

### DIFF
--- a/apps/dialtone-documentation/docs/utilities/responsive/breakpoints.md
+++ b/apps/dialtone-documentation/docs/utilities/responsive/breakpoints.md
@@ -59,24 +59,6 @@ To use [PostCSS](https://www.npmjs.com/package/postcss) with our custom plugin
 [@dialpad/postcss-responsive-variations](https://www.npmjs.com/package/@dialpad/postcss-responsive-variations),
 install it and configure your PostCSS setup.
 
-## Usage
-
-<code-well-header>
-  <div class="d-fl-center d-w100p d-m8 d-p16 d-bgc-moderate d-bar4 d-ta-center">This is visible on all screens</div>
-  <div class="d-fl-center d-w100p d-m8 d-p16 d-bgc-moderate d-bar4 d-ta-center d-d-none xl:d-d-block">This is visible on extra large screens</div>
-  <div class="d-fl-center d-w100p d-m8 d-p16 d-bgc-moderate d-bar4 d-ta-center d-d-none lg:d-d-block">This is visible on large screens</div>
-  <div class="d-fl-center d-w100p d-m8 d-p16 d-bgc-moderate d-bar4 d-ta-center d-d-none md:d-d-block">This is visible on medium screens</div>
-  <div class="d-fl-center d-w100p d-m8 d-p16 d-bgc-moderate d-bar4 d-ta-center d-d-none sm:d-d-block">This is visible on small screens</div>
-</code-well-header>
-
-```html
-<div>...</div>
-<div class="d-d-none xl:d-d-block">...</div>
-<div class="d-d-none lg:d-d-block">...</div>
-<div class="d-d-none md:d-d-block">...</div>
-<div class="d-d-none sm:d-d-block">...</div>
-```
-
 ## Configuration
 
 ### Step 1: Install the Plugin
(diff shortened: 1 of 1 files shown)


apps/dialtone-documentation/docs/utilities/responsive/breakpoints.md
```diff
@@ -21,12 +71,75 @@ This plugin takes the breakpoints and the classes you need to have responsive va
...
+```
+
+### Configuration Options
+
+**`breakpoints`** (optional): Array of breakpoint objects. Each object has:
+
+- `prefix`: The prefix to use (e.g., `'sm\:'`)
+- `mediaQuery`: The CSS media query (e.g., `'(min-width: 480px)'`)
+
+**`classes`**: Array of class selectors (strings or regex patterns) that should get responsive variations.
```
braddialpad, line 107: It feels a bit like this should be moved above the examples showing breakpoints and classes. Not really a big deal, up to you if you prefer it.

apps/dialtone-documentation/docs/utilities/responsive/breakpoints.md
```diff
@@ -21,12 +71,75 @@ This plugin takes the breakpoints and the classes you need to have responsive va
...
+  '.d-d-block',
+
+  // Regex for multiple related classes
+  /\.d-d-(flex|none|block)$/,     // Matches .d-d-flex, .d-d-none, .d-d-block
+  /\.d-p[0-9]+$/,                 // Matches .d-p8, .d-p16, .d-p24, etc.
+  /\.d-m[tblr]?[0-9]*$/,          // Matches .d-m8, .d-mt16, .d-ml24, etc.
+];
+```
+
+**Result**: The plugin will generate responsive versions like `sm:d-d-block`, `md:d-p16`, `lg:d-mt8`, etc.
```
braddialpad, line 158: Let's mention to be careful of performance / CSS size here, every class added here will generate 4 new classes in your CSS bundle so we should mention to try to only add ones that are actually used.

braddialpad (commented):
Looks good Nina!

apps/dialtone-documentation/docs/utilities/responsive/breakpoints.md
ninamarina: Yes, I agree 👍

ninamarina:
@francisrupert tagging you because you were interested in this one, in case you want to take a look

francisrupert:
Change these...
image
...to read as...

```
Visible on **all** screens
Visible only on screens wider than **extra-large** breakpoint
Visible only on screens wider than **large** breakpoint
Visible only on screens wider than **medium** breakpoint
Visible only on screens wider than **small** breakpoint
```

francisrupert:
Somewhere within the `How Responsive Variations Work` section there should be some language addressing why `min-width` is the default recommendation, i.e. a "mobile-first" approach rather than a "designing downward" approach, possibly even referencing https://dialtone.dialpad.com/about/whats-new/posts/2025-5-6.html.

francisrupert:
The only way to appreciate the demo on the page is to resize the doc site, which is a little awkward. This is probably a doc improvement to log, but a more specific demo would be cool. [Something like this](https://usda-fsa.github.io/fsa-design-system/components/buttons/#responsive) with `<iframe>`.

https://github.com/user-attachments/assets/ac13533b-2a9b-49db-96cf-e1912fadeb02

francisrupert (approved):
This is great, thanks for the attention to detail!

ninamarina:
> The only way to appreciate the demo on the page is to resize the doc site, which is a little awkward. This is probably a doc improvement to log, but a more specific demo would be cool. [Something like this](https://usda-fsa.github.io/fsa-design-system/components/buttons/#responsive) with `<iframe>`.
> 
>  rwd.mov

Will log a ticket for this one

PR #825, opened 2025-07-15
fix(modal): NO-JIRA prevent losing focus when clicking inside the modal

ninamarina:
# fix(modal): NO-JIRA prevent losing focus when clicking inside the modal

When working on [this PR](https://github.com/dialpad/firespotter/pull/60607) I came across these small issues in the modal component.

1- When clicking inside the modal, the focus would go to the background
2- The parameter for Vue transition `after-enter` callback is the element, not the event, so I removed this condition `(event.target === event.currentTarget)`. It was always true because both `event.target` and `event.currentTarget` are undefined always (check https://vuejs.org/guide/built-ins/transition)

Also added `await` for async functions.

## Before
Open the modal and click inside of it (for example in the title)
modal-click

## With this fix
modal-click-pr

### DIFF
--- a/packages/dialtone-vue2/components/modal/modal.vue
+++ b/packages/dialtone-vue2/components/modal/modal.vue
@@ -343,8 +343,14 @@ export default {
         ...this.$listeners,
 
         click: event => {
-          if (!this.closeOnClick) return;
-          (event.target === event.currentTarget) && this.close();
+          // Handle backdrop clicks for closing modal
+          if (this.closeOnClick && event.target === event.currentTarget) {
+            this.close();
+          } else if (this.show && event.target !== event.currentTarget) {
+            // Ensure focus stays within modal when clicking inside it
+            this.handleModalClick(event);
+          }
+
           this.$emit('click', event);
         },
 
@@ -361,9 +367,9 @@ export default {
           this.$emit('keydown', event);
         },
 
-        'after-enter': event => {
+        'after-enter': async () => {
           this.$emit('update:show', true);
-          (event.target === event.currentTarget) && this.setFocusAfterTransition();
+          await this.setFocusAfterTransition();
         },
       };
     },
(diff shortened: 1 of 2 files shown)


packages/dialtone-vue2/components/modal/modal.vue
```diff
@@ -422,6 +428,20 @@ export default {
...
+      const focusableElements = this._getFocusableElements();
+
+      // If the clicked element is not focusable, ensure focus stays in modal
+      if (focusableElements.length && !focusableElements.includes(clickedElement)) {
+        // Check if current active element is still within the modal
+        if (!focusableElements.includes(document.activeElement)) {
+          this.focusFirstElement();
+        }
+      }
+    },
```
braddialpad, line 444: I am sure what you have done here works, but I am sort of confused why we have to do this. When you click the modal it renders a backdrop overlay (class `.d-modal`) that prevents clicking or scrolling on any of the content behind the modal. Are they somehow overriding this in the reported case?

braddialpad: ohhh i wonder if it's because of our custom scrollbar in the docsite.

braddialpad: seems like it doesn't allow scrolling just pressing tab

braddialpad (approved):
Don't fully understand why the click was going "through" the modal to the background in the first place, but this change looks okay.

packages/dialtone-vue2/components/modal/modal.vue
ninamarina: > ohhh i wonder if it's because of our custom scrollbar in the docsite.

It does happen in some parts of the product too, if you go to https://dialpadbeta.com/coachinghub and click "edit" and then click the title of the modal and start pressing tab, this happens.
modal-click-beta

ninamarina: > seems like it doesn't allow scrolling just pressing tab

yes, no scrolling, only pressing tab will focus elements on the background

ninamarina:
> Don't fully understand why the click was going "through" the modal to the background in the first place, but this change looks okay.

I think it's because when clicking in a part of the modal, the element that had focus loses the focus, but there was no focus management from the Modal itself in that case, only the click handler to close it if the click was on the backdrop. So it was up to the browser to decide who gets the focus, that's why in some cases, not all, the focus would go to the background.

PR #832, opened 2025-07-21
feat(icon): DLT-2619 deprecate ai notes

fede-dp:
# feat: DLT-2619 deprecate ai notes icon

a-cartoon-duck-holding-a-notebook-and-a-pencil

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2619

## :pencil: Checklist

For all PRs:


fede-dp:
this makes me change the .husky/pre-commit to commit, not sure why 🤷🏼‍♂️  but i can't commit without it

francisrupert (changes requested):
Please update PR title to include scope. I've called this out far too often in your PRs, and it remains inconsistent.

In this case, it should be `feat(icon): DLT-2619 deprecate ai notes`. 

Ask if the reason for this is unclear.

packages/dialtone-icons/src/svg/icons/brand/ai-notes.svg
braddialpad, line 1: Are we looking to "deprecate" or just "remove". Usually deprecate means it still exists in the library but is not advertised in documentation. In this case we are fully removing it so it no longer exists. This is a breaking change also if consumers are already using this icon.

.husky/pre-commit
```diff
@@ -8,7 +8,7 @@ if ! git diff --staged --quiet packages/dialtone-icons/src; then
   pnpm nx run dialtone-icons:build;
   # node packages/dialtone-vue2/localization/sync-icons.js;
 
-  git packages/dialtone-icons/src/keywords-*;
+  git add packages/dialtone-icons/src/keywords-*;
```
braddialpad: Wait did you change this manually or did it just automatically happen?

packages/dialtone-icons/src/svg/icons/brand/ai-notes.svg
fede-dp: the main idea is that the icon it's only used here → https://dialpad.atlassian.net/browse/DP-141354?focusedCommentId=1319825 so i remove it and delete it from the library, will be better to 'hide it' for now and delete this later down the road?

.husky/pre-commit
fede-dp: this is what i was reffering to here → https://github.com/dialpad/dialtone/pull/832#issuecomment-3097045705 not sure why but can't commit without that change 🤷🏼‍♂️ we can delete if not necessary

fede-dp: could revert that this morning! so we are good 👍🏼 
do you think we should also keep the icon hidden? @braddialpad

braddialpad: confirmed with tico this was a mistake he made in a prior PR. don't worry about changing it though, he will fix it.

packages/dialtone-icons/src/svg/icons/brand/ai-notes.svg
braddialpad: Since we've agreed to prevent any breaking changes unless for major releases, we should probably just hide it and mark it to be fully removed for the next major release.

francisrupert: A little confused... why a new Draft PR, https://github.com/dialpad/dialtone/pull/839? I agree we shouldn't outright remove the icon, and just suppress it from appearing on doc site.

So you can just update this PR to still have the SVG, `dialtone-icons/src/svg/icons/brand/ai-notes.svg`, and still remove it from `packages/dialtone-icons/src/keywords-icons.json` to prevent it from appearing as an option.

juliodialpad:
Closing this PR as #839 is already approved.

PR #849, opened 2025-07-30
fix(tab): DP-148500 tabs not loading instantly

iropolo:
# fix(tab): DP-148500 tabs not loading instantly

- [x] Fix

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DP-148500

## :book: Description

`updateSelected` will be triggered on  `updated` instead of  `beforeUpdate`.

## :bulb: Context

tab-group children are not getting updated, making this.tabs missing tab causing undefined.
Changing  `beforeUpdate` to `updated` seems to fix it.
Fix proposed by: @abhishekjaindialpad


ninamarina:
I see in the docs that it's not a good idea to update the state inside the `updated()` hook since it might cause an infinite loop: https://vuejs.org/api/options-lifecycle.html#updated
Is the `updateSelected()` in the `mounted` hook not being called?

iropolo:
> I see in the docs that it's not a good idea to update the state inside the `updated()` hook since it might cause an infinite loop: https://vuejs.org/api/options-lifecycle.html#updated Is the `updateSelected()` in the `mounted` hook not being called?

Instead of create the bug ticket I just create the contribution of @abhishekjaindialpad , this comes from Triage.

If you have the time to take it a look, go ahead. Seems is not urgent.
I will try to give it a chance later today or tomorrow.

abhishekjaindialpad:
> > I see in the docs that it's not a good idea to update the state inside the `updated()` hook since it might cause an infinite loop: https://vuejs.org/api/options-lifecycle.html#updated Is the `updateSelected()` in the `mounted` hook not being called?
> 
> Instead of create the bug ticket I just create the contribution of @abhishekjaindialpad , this comes from Triage.
> 
> If you have the time to take it a look, go ahead. Seems is not urgent. I will try to give it a chance later today or tomorrow.

If it takes time to decide, I suggest we should make changes here, return 0, dont return -1, since -1 should never be a case. 

```
    getFocusedTabIndex () {
      return this.tabs.findIndex((context) =>
        this.focusId ? context.tabId === `${this.focusId}` : context.isSelected);
    },
```

iropolo:
Okay, Im not sure if this will fix the title of the bug ticket `App Tabs (content) not Loading Instantly` but this will probably fix the issue reported in this comment https://dialpad.atlassian.net/browse/DP-148500?focusedCommentId=1354870


I guess I understand what @abhishekjaindialpad suggest with His comment 

> If it takes time to decide, I suggest we should make changes here, return 0, dont return -1, since -1 should never be a case.

Seems like our function `getFocusedTabIndex` throw `-1` if there is no match (expected result of findIndex function). And this result in the issue because it cannot be a valid case. `this.tabs[-1]` 

Screenshot 2025-08-01 at 4 36 38 PM


results in -> 

image

because this function fails ->

```
isSameTabClicked () {
      const tab = this.tabs[this.getFocusedTabIndex()];
      return this.provideObj.selected === tab.panelId;
    },
```


Instead change it to 0.

image


I suggest this:

```
    getFocusedTabIndex () {
      const index = this.tabs.findIndex((context) =>
        this.focusId ? context.tabId === `${this.focusId}` : context.isSelected
      );

      return index === -1 ? 0 : index;
    },
```


Did I understand it right?
This is a hotfix.
The main issue is that this.tabs is not being updated at the time this is being triggered.


Screenshot 2025-08-01 at 4 52 45 PM

abhishekjaindialpad:
> Okay, Im not sure if this will fix the title of the bug ticket `App Tabs (content) not Loading Instantly` but this will probably fix the issue reported in this comment https://dialpad.atlassian.net/browse/DP-148500?focusedCommentId=1354870
> 
> I guess I understand what @abhishekjaindialpad suggest with His comment
> 
> > If it takes time to decide, I suggest we should make changes here, return 0, dont return -1, since -1 should never be a case.
> 
> Seems like our function `getFocusedTabIndex` throw `-1` if there is no match (expected result of findIndex function). And this result in the issue because it cannot be a valid case. `this.tabs[-1]`
> 
> Screenshot 2025-08-01 at 4 36 38 PM
> results in ->
> 
> image
> because this function fails ->
> 
> ```
> isSameTabClicked () {
>       const tab = this.tabs[this.getFocusedTabIndex()];
>       return this.provideObj.selected === tab.panelId;
>     },
> ```
> 
> Instead change it to 0.
> 
> image
> I suggest this:
> 
> ```
>     getFocusedTabIndex () {
>       const index = this.tabs.findIndex((context) =>
>         this.focusId ? context.tabId === `${this.focusId}` : context.isSelected
>       );
> 
>       return index === -1 ? 0 : index;
>     },
> ```
> 
> Did I understand it right? This is a hotfix. The main issue is that this.tabs is not being updated at the time this is being triggered.
> 
> Screenshot 2025-08-01 at 4 52 45 PM

Yes this is what I meant and I agree with code change, we can't keep prod issue open for long period of time and with two weeks of release cycle, it is more critical that it reaches Monday or Tuesday beta cut.

```
    getFocusedTabIndex () {
      const index = this.tabs.findIndex((context) =>
        this.focusId ? context.tabId === `${this.focusId}` : context.isSelected
      );

      return index === -1 ? 0 : index;
    },
```

iropolo:
Im testing it.

Previously:

https://github.com/user-attachments/assets/56ea55c4-c7d7-4ad8-b854-df458e506e3d



With hotfix (does not block the user navigation):

https://github.com/user-attachments/assets/368f0adb-ebd9-4177-95b4-7bd223e6ed1f

abhishekjaindialpad (approved):

PR #860, opened 2025-08-05
fix(rich-text-editor): DLT-2690 custom emojis not working

iropolo:
# fix(rich-text-editor): DLT-2690 custom emojis not working

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2690

## :book: Description

Seems like our implementation on rich text editor component get outdated.
Thats why we get the reported error.

What Im doing in this PR is improving the implementation and sync the custom emoji data with the expected from Backend

Since this is an ongoing feature we will probably continue doing changes.

## :bulb: Context

We are working on the custom emoji feature.

This will give the possibility to add jpeg/png/gif emoji to our emoji picker.

When we select an emoji, it need to be rendered, we do it in:

packages/dialtone-vue2/components/rich_text_editor/extensions/emoji/EmojiComponent.vue

https://github.com/user-attachments/assets/896c9315-85dd-4010-8caf-83aff0257ba0

## On image load error:
Similar as what is DtEmoji does. We are showing a skeleton until the image load.
If the image fails, we remove it with 
`event.target.parentNode.remove();` to avoid polluting the input.

https://github.com/user-attachments/assets/6a743612-94f4-4072-be78-ef4a3650e872

## :crystal_ball: Next Steps

DtEmoji -> emoji.vue component does not support custom emojis. It only receives a `code` prop that it will use to find the emoji in our `emojiJsonLocal`

I updated some tests
- `common/emoji.test.js`
- `components/emoji/emoji.test.js` 
- `emoji_text_wrapper.test.js`

but all these should be removed along with a clean up of `common/emoji/index.js`.

At this moment we are using custom emojis in

- emoji_selector-> in emoji_picker component
- EmojiComponent.vue -> in rich_text_editor component (to render the emoji in message_input)

emoji_selector does not use DtEmoji because if I remember correctly it has a performance issue.

EmojiComponent.vue uses DtEmoji to handle our local emojis and has a different implementation to handle custom emojis (similiar to emoji_selector but with a skeleton)

I would like to somehow unify this in one component/handler.
I expect that DtEmoji will support custom emojies, or maybe we need to create a different component as DtCustomEmoji to take care of it then we can use it in our different implementations.

I use this space to avoid go further in changes in this PR since it is a 'quick fix' to unlock work to @bianca-artola-dialpad and team.

Possible upcoming tickets im thinking we can generate.
Spike -> Investigate how to unify the implementation of emoji with custom emoji
Story -> Clean up of custom emoji on emoji files. (and probably implement new ones like tests)

----
I will add the vue3 implementation after the changes approval.

### DIFF
--- a/packages/dialtone-vue2/components/rich_text_editor/extensions/emoji/EmojiComponent.vue
+++ b/packages/dialtone-vue2/components/rich_text_editor/extensions/emoji/EmojiComponent.vue
@@ -4,15 +4,25 @@
     class="d-d-inline-block d-va-bottom d-lh0"
   >
     <dt-emoji
+      v-if="node.attrs.code"
       size="500"
       :code="node.attrs.code"
     />
+    <img
+      v-else
+      class="d-icon d-icon--size-500"
+      :alt="node.attrs.name"
+      :aria-label="node.attrs.name"
+      :title="node.attrs.name"
+      :src="getImgSrc(node.attrs)"
+      @error="handleImageError"
+    >
   </node-view-wrapper>
 </template>
 
 <script>
 import { nodeViewProps, NodeViewWrapper } from '@tiptap/vue-2';
-
+import { CDN_URL } from '@/components/emoji_picker/emoji_picker_constants';
 import { DtEmoji } from '@/components/emoji';
 
 export default {
@@ -23,5 +33,27 @@ export default {
   },
 
   props: nodeViewProps,
+
+  computed: {
+    CDN_URL () {
+      return CDN_URL;
+    },
+  },
+
+  methods: {
+    getImgSrc: function (emoji) {
+      // TODO Update json structure to have a property for custom emojis and avoid using date_added
+      if (emoji.image) { // if custom emoji
+        return emoji.image;
+      } else { // if regular emoji
+        return this.CDN_URL + emoji.code + '.png';
+      }
+    },
+
+    handleImageError: function (event) {
+      console.log('test error emoji');
+      // event.target.parentNode.style.display = 'none';
+    },
+  }
 };
 </script>
--- a/packages/dialtone-vue2/components/rich_text_editor/extensions/emoji/EmojiComponent.vue
+++ b/packages/dialtone-vue2/components/rich_text_editor/extensions/emoji/EmojiComponent.vue
@@ -8,51 +8,60 @@
       size="500"
       :code="node.attrs.code"
     />
-    <img
-      v-else
-      class="d-icon d-icon--size-500"
-      :alt="node.attrs.name"
-      :aria-label="node.attrs.name"
-      :title="node.attrs.name"
-      :src="getImgSrc(node.attrs)"
-      @error="handleImageError"
-    >
+
+    <template v-else>
+      <dt-skeleton
+        v-if="showSkeleton"
+        :offset="0"
+        class="d-icon d-icon--size-500"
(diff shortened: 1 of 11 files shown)


packages/dialtone-vue2/components/rich_text_editor/extensions/emoji/EmojiComponent.vue
```diff
@@ -4,24 +4,65 @@
...
+    <template v-else>
+      <dt-skeleton
+        v-if="showSkeleton"
+        :offset="0"
+        class="d-icon d-icon--size-500"
+        :shape-option="{ shape: 'circle', size: '100%' }"
+      />
+
+      <img
+        v-show="!showSkeleton"
```
bianca-artola-dialpad, line 21: Just curiosity, but does this not work with `v-else`?

bianca-artola-dialpad (approved):
thanks for the quick fix 🙏🏽

packages/dialtone-vue2/components/rich_text_editor/extensions/emoji/EmojiComponent.vue
iropolo: yeah, seems intuitive but it will not work with `v-else` because it is not in the same compilation time, what Im doing is showing the skeleton whatever is the result of the image loading (because I don't know if the image is valid or not). AND rendering the image to trigger the load, then it will show/hide the skeleton and the image.

tldr: we are doing it in `emoji.vue` the difference but here Im using `v-if` to not polluting the message input

iropolo: Im not sure this will be needed, because the image is already being loaded in the emoji-picker, so the source should be always right.
If the emoji-picker cannot load the image, it will not show it so you cannot select it.

I copied the `handleImageError` from there so.. I believe we are doing it twice unnecessarily, so all this skeleton behaviour is just to replicate what we do with the normal emoji.

But yeah... I mean... we can let it there.

bianca-artola-dialpad:
Hey team, I'm wondering if we can have this reviewed/merged. We are aiming for code completion of custom emojis in two weeks, so I want to unblock the work as much as possible.
Thank you and sorry 🙏🏽

packages/dialtone-vue2/components/rich_text_editor/extensions/emoji/EmojiComponent.vue
```diff
@@ -4,24 +4,65 @@
...
+  },
+
+  methods: {
+    handleImageLoad () {
+      this.showSkeleton = false;
+    },
+
+    handleImageError: function (event) {
+      this.showSkeleton = false;
+      event.target.parentNode.remove();
```
braddialpad, line 64: We just want nothing to show at all on image error? That seems like confusing behaviour

braddialpad (approved):
This seems like it needs further improvement as you have mentioned in the description, but this should be fine to merge to unblock.

packages/dialtone-vue2/components/rich_text_editor/extensions/emoji/EmojiComponent.vue
iropolo: yeah, as I said it is very unlikely situation.

iropolo:
I will prepare vue3 update in a different PR somewhile today or next week.

PR #840, opened 2025-07-28
feat(color): DLT-2594 shell design tokens

jcarter-dp:
# feat(color): DLT-2594 shell design tokens

## :hammer_and_wrench: Type Of Change

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2594

## :book: Description

Creating the `shell` design tokens and handful of themes, e.g. Sunflower, Plum, High Desert, etc. 

## :pencil: Checklist

For all PRs:

- [x] I have ensured no private Dialpad links or info are in the code or pull request description (Dialtone is a public repo!).
- [x] I have reviewed my changes.
- [ ] I have added all relevant documentation.
- [x] I have considered the performance impact of my change.

## :crystal_ball: Next Steps

- Test shell tokens in the beacon prototype
- Apply to Figma Components (i.e. top bar / left bar in App Components)
- Update CSS for (e.g. "Leftbar" Recipes, Unread Pill, etc.)
- Update Presence CSS, including updating Presence within Leftbar rows w/ shell override.

## :camera: Screenshots / GIFs

Samples...

### Light
https://github.com/user-attachments/assets/0d58d916-c7d1-4a23-ab16-7c77cdbc9d04

### Dark
https://github.com/user-attachments/assets/6fdf9bd9-9ce7-40d1-80b7-8d909a858225

PR #867, opened 2025-08-12
feat(themes): DLT-2702  shell design tokens

juliodialpad:
# Generate new themes

Obligatory GIF

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2702

## :book: Description

- Added a script to automatically generate themes based on `tokens/themes` directories this way if we add or remove themes from Figma they'll be automatically synced 🪄
- Removed manually created themes and added a .gitignore file to ignore every file under `themes` except of `config.js` and `debug.js` this way we don't ship a lot of duplicated files that can be generated on the fly.
- Renamed `ceruleo-default` theme to `ceruleo-light`.
- Updated `build-ios` script to copy all generated themes.
- Updated `vite.config.js` to automatically find `*.js` entries under `postcss` and `themes` directories.

## :bulb: Context

After #840 we need to verify that all the new added themes are being exported correctly

## :pencil: Checklist

For all PRs:


apps/dialtone-documentation/docs/.vuepress/baseComponents/tokens/constants.js
```diff
@@ -14,11 +14,19 @@ export const BRANDS = [
   { value: 'tmo', label: 'T-Mobile' },
   { value: 'expressive', label: 'Expressive' },
   { value: 'expressive-sm', label: 'Expressive Small' },
```
francisrupert: Remove `expressive`, or at least comment out for now.

juliodialpad: Just from the docs? or completely remove the output in the tokens?

francisrupert: From doc site for now.

jcarter-dp:
@francisrupert  Should the shell tokens be placed as a different category like "themes" instead of "components"? I would think shell tokens are kind of its own category cause its specific purpose is theming.

jcarter-dp:
Could we change the name here so that theme is mode and brand is theme?
Screenshot 2025-08-19 at 1 24 00 PM

jcarter-dp:
@juliodialpad There is a large list of shell tokens missing. I notice we have only action, presence and mention shell tokens. But I'm not seeing the other tokens like base, foreground, surface and border shell tokens in there.

jcarter-dp:
> @juliodialpad There is a large list of shell tokens missing. I notice we have only action, presence and mention shell tokens. But I'm not seeing the other tokens like base, foreground, surface and border shell tokens in there.

Nevermind, the tokens are in there..I just see that they aren't listed in the tokens page here:

juliodialpad:
> @juliodialpad There is a large list of shell tokens missing. I notice we have only action, presence and mention shell tokens. But I'm not seeing the other tokens like base, foreground, surface and border shell tokens in there.

I hid those intentionally on that table, let me know if you want all of them to be displayed there.

braddialpad (commented):
This seems like a significant breaking change right? should be merged to next? or do we want to increment a major version for this?

francisrupert:
> This seems like a significant breaking change right? should be merged to next? or do we want to increment a major version for this?

Technically I don't think it is breaking. The current `var(--dt-theme...)` tokens remain available for backwards-compatibility, and the key addition is introducing new tokens with a namespace of `var(--dt-shell...)` which was https://github.com/dialpad/dialtone/pull/840.

Those `var(--dt-theme...)` CSS Variables in DT Components/Recipes will next be updated by @jcarter-dp to the `var(--dt-shell...)` ones in a follow-up PR; and another for in-product use (i.e. https://github.com/search?q=repo%3Adialpad%2Ffirespotter+dt-theme-&type=code). After that, all `var(--dt-theme...)` tokens would be removed.

That's at least the perspective of @jcarter-dp and I, and I'll let @juliodialpad verify if the changes he listed in the Description (e.g. "Added a script to automatically generate themes...") isn't breaking.

braddialpad:
> Technically I don't think it is breaking. The current `var(--dt-theme...)` tokens remain available for backwards-compatibility, and the key addition is introducing new tokens with a namespace of `var(--dt-shell...)` which was

Ah okay, my mistake, thought we were removing the old --dt-theme tokens

braddialpad (approved):
Looks good

juliodialpad:
> That's at least the perspective of @jcarter-dp and I, and I'll let @juliodialpad verify if the changes he listed in the Description (e.g. "Added a script to automatically generate themes...") isn't breaking.

It's not a breaking change, it generates the same themes as before and adds the new ones automatically.

juliodialpad:
I'll do this in another PR, had to change a lot of things to keep this naming consistent across the code and not only displaying that, also I'm trying to enable theme switching on the docsite.

> Could we change the name here so that theme is mode and brand is theme? Screenshot 2025-08-19 at 1 24 00 PM

PR #868, opened 2025-08-12
fix(emoji): NO-JIRA replace internal input value when adding custom emojis

bianca-artola-dialpad:
# fix(emoji): NO-JIRA replace internal input value when adding custom emojis

Obligatory GIF

- [x] Fix

## :book: Jira Ticket
No jira

## :book: Description
On the product side we need to update internally the input to :custom_emoji: instead of url


braddialpad:
Can we get the custom emoji storybook example working again? @iropolo could you please help bianca with this.

Screenshot 2025-08-12 at 1 22 24 PM

braddialpad (changes requested):
custom emoji correctly display if I add them via the picker, but if I type in the shortname they do not display in the input, and also do not display in the suggestions.

Screenshot 2025-08-12 at 1 25 52 PM

Screenshot 2025-08-12 at 1 27 04 PM

iropolo:
@braddialpad Custom emoji on suggestion list wasn't a previous behavior and it is not contemplated at this stage. Not sure how it will be done for now since it is a new feature.

You can test the message_input with :baby_yoda: :blinkingguy: and :awkward_seal:, the component will catch the string inside `:` and find if there is an emoji to show. (Like how it was done before)

@bianca-artola-dialpad please check if the PR looks good.

braddialpad:
> You can test the message_input with :baby_yoda: :blinkingguy: and :awkward_seal:, the component will catch the string inside `:` and find if there is an emoji to show. (Like how it was done before)

Yep works now thanks, really nice change!

> @braddialpad Custom emoji on suggestion list wasn't a previous behaviour and it is not contemplated at this stage. Not sure how it will be done for now since it is a new feature.

Feels kind of incomplete without this doesn't it? It seems pretty simple to do, just needed to use `import { emojisIndexed } from '@dialpad/dialtone-emojis';` instead of referencing `emojisIndexed` directly.

I pushed a change to add this, let me know if that's okay.

bianca-artola-dialpad:
Just tested, and I don't see any issue with these new changes
Screenshot 2025-08-13 at 4 26 38 PM
Screenshot 2025-08-13 at 4 25 28 PM

Great improvement having it on the shortcut list, thanks! 👏🏽  
I will retest this with [this PR](https://github.com/dialpad/firespotter/pull/58321) once we have it but we should be good

braddialpad:
Just added some sorting logic as well, since custom emojis were always showing above native emojis.

packages/dialtone-vue2/common/custom-emoji.json
```diff
@@ -4,20 +4,32 @@
     "content_type": "image/gif",
     "date_added": 1730918926288,
     "image": "https://storage.googleapis.com/uv-beta_custom_emojis/5646620347596800/blinkingguy",
-    "name": "blinkingguy"
+    "name": "blinkingguy",
+    "unicode_output": "3a-62-6c-69-6e-6b-69-6e-67-67-75-79-3a",
```
braddialpad, line 8: @iropolo what is this unicode_output? is it just some random identifier?

braddialpad (commented):
We should port this to vue 3 before merging.

packages/dialtone-vue2/common/custom-emoji.json
iropolo: I guess, this comes from `stringToCodePoints` on product side implementation 
cc. @bianca-artola-dialpad

braddialpad: I don't think it means anything valuable since a custom emoji doesn't have a unicode character. Is it possible to remove or does that break everything?

iropolo: We used it in `packages/dialtone-vue2/components/rich_text_editor/extensions/emoji/emoji.js` to render it.
Seems like is rich text editor stuff needed. Not sure how needed it is.

iropolo:
> 
> Feels kind of incomplete without this doesn't it? It seems pretty simple to do, just needed to use `import { emojisIndexed } from '@dialpad/dialtone-emojis';` instead of referencing `emojisIndexed` directly.
> 
> I pushed a change to add this, let me know if that's okay.

Looks good for Dialtone, not sure how it will work for product side on how they will provide the list or searching.

iropolo:
> We should port this to vue 3 before merging.

okay should be quick I will add it after this PR approval to avoid continue doing changes, and we will merge it all together.

packages/dialtone-vue2/common/custom-emoji.json
braddialpad: Alright guessing it is used as a key/identifier then. The only problem is that for any custom emojis that are added on the product side they will also need this to have this, do we know if they have it?

braddialpad:
> Looks good for Dialtone, not sure how it will work for product side on how they will provide the list or searching.

Should work fine as long as the custom emoji list they provide has the same properties.

packages/dialtone-vue2/common/custom-emoji.json
iropolo: Seems like they generated it here based on the `name` `ubervoice/static/js/utils/custom_emoji_util.js` ([pr](https://github.com/dialpad/firespotter/pull/58321/files))

I used it to get the needed unicode_output

image

bianca-artola-dialpad: yeah anyway each name should be unique... I think I can get ride of this stringToCodePoints function if u agree

bianca-artola-dialpad:
> > Looks good for Dialtone, not sure how it will work for product side on how they will provide the list or searching.
> 
> Should work fine as long as the custom emoji list they provide has the same properties.

yeap we don't have extra logic on product side to render the emoji suggestion so it _should_ work good

packages/dialtone-vue2/common/custom-emoji.json
iropolo: I guess we are okay with this implementation

iropolo:
Working on vue3 version

bianca-artola-dialpad:
Just tested this one with the generated version and it works okay on product side!


https://github.com/user-attachments/assets/3502ce0f-c19b-4171-bbd8-3b2661c5891b

thanks

braddialpad:
> Just tested this one with the generated version and it works okay on product side!
> 
>  Screen.Recording.2025-08-14.at.1.28.37.PM.mov 
> thanks

nice! looks like we are good to go then

braddialpad (approved):

iropolo (approved):

PR #871, opened 2025-08-13
docs(toggle): DLT-2623 fix vue 3 toggle variants

braddialpad:
Simple fix for vue 3 toggle variants in storybook. They were still using checked when they should have been using model-value.

Documentation fix only


juliodialpad (approved):
Looks good, thanks!

PR #872, opened 2025-08-15
refactor(css): DLT-1876 css logical properties

francisrupert:
# Refactor to CSS Logical Properties

a96c0b57528d6cdceff4c0a5c68ed1e6

- [x] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1055

## :book: Description

Convert all the DT things to CSS Logical Properties. Intentionally targeting `next`. Would use [Beacon](https://dialpad.github.io/beacon-app/#/inbox) to test.

## :pencil: Checklist

For all PRs:

For all CSS changes:


plan--css-logical-properties.md
```diff
@@ -0,0 +1,822 @@
...
+node scripts/migrate-to-logical.js --dry-run --path="utilities/spacing.less"
+
+# Apply changes with backup
+node scripts/migrate-to-logical.js --path="utilities/" --backup
+
+# Generate report only
+node scripts/migrate-to-logical.js --report-only --path="components/"
+```
+
+#### 2. Stylelint Rules Implementation
```
francisrupert: I held back from following through on this part of the plan since it should probably be a separate task and PR in and of itself. 

Still wouldn't mind reactions to it though. It's 90% generated by Claude Code after the actual migration of the files listed in this PR.

braddialpad:
Just was thinking about how hard it's going to be to switch my brain over to this, and found this awesome stylelint rule: https://github.com/yuschick/stylelint-plugin-logical-css

braddialpad:
The fact percy found no visual changes is a really good sign

plan--css-logical-properties.md
juliodialpad: The plan is functional and will need some minor adjustments/verifications, but seems reasonable to me.

juliodialpad (commented):
This is a pretty huge change which will not only include Dialtone, but all products in order to be successfully implemented.

Do we want to keep utility classes like `d-ml-*, d-mr-*`? I'd prefer to rename them to the logical way too, something like `d-ms-*, d-me-*` I think tailwindCSS is already doing so and seems like the way to go for me.

plan--css-logical-properties.md
francisrupert: I've extracted that out into a separate task, https://dialpad.atlassian.net/browse/DLT-2723.

francisrupert:
> Just was thinking about how hard it's going to be to switch my brain over to this, and found this awesome stylelint rule: https://github.com/yuschick/stylelint-plugin-logical-css

Heck yeah. I extracted out that portion of the temporary plan file and speculated on how that might be done in a distinct Jira task, https://dialpad.atlassian.net/browse/DLT-2723, including making use of that plugin.

francisrupert:
> Do we want to keep utility classes like `d-ml-*, d-mr-*`? I'd prefer to rename them to the logical way too, something like `d-ms-*, d-me-*` I think tailwindCSS is already doing so and seems like the way to go for me.

Definitely a target, though of course not as part of this task. To start with, I don't really mind if a CSS Utility is named in a physical way, `d-pl`, while its property is logical `padding-inline-start`. 

Perhaps next supporting both initially, e.g `.d-pl, .d-pis { ... }`, followed by a planned massive search/replace migration, and then retire them outright.

juliodialpad (approved):
This is a huge PR that'd be difficult to carefully review every single line, mostly a search-replace task and I couldn't find any issue at a glance so this look good IMO, thanks!

francisrupert:
> This is a huge PR that'd be difficult to carefully review every single line, mostly a search-replace task and I couldn't find any issue at a glance so this look good IMO, thanks!

Thanks @juliodialpad! Merging now. Since this will be on `next`, we'll keep an eye out for anything off. But I'm feeling real good about the result, especially w/ Percy's results. I'm also gonna do a local test of using it in ubervoice to more fully gut-check it.

Next up is [DLT-2723](https://dialpad.atlassian.net/browse/DLT-2723) for automating it – linting, pre-commit hook, etc. 

[DLT-2723]: https://dialpad.atlassian.net/browse/DLT-2723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #873, opened 2025-08-18
feat(message-input): DLT-2714 add scroll-bottom-reached event on input

bianca-artola-dialpad:
# feat(message-input): DLT-2714 add scroll-bottom-reached event on input

Obligatory GIF

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket
https://dialpad.atlassian.net/browse/DLT-2714

## :book: Description

This pull request adds the ability to notify when they have scrolled to the bottom of the emoji picker in the message input.

## :bulb: Context

When using the message input component on the Firespotter side, some use cases require the emoji picker provided by Dialtone (which doesn't include a Giphy picker). In these situations, we need to know when a user has reached the end of the custom emoji list so we can continue fetching more emojis.

The scroll-bottom-reached event will only be emitted on the picker for cases where it is handled separately from the input.

### DIFF
--- a/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.stories.js
+++ b/packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.stories.js
@@ -191,7 +191,7 @@ export const argTypesData = {
     },
   },
 
-  onScrollBottomReached: {
+  onEmojiScrollBottomReached: {
     table: {
       disable: true,
     },
@@ -243,7 +243,7 @@ export const argsData = {
   onNoticeClose: action('notice-close'),
   onSkinTone: action('skin-tone'),
   onCancel: action('cancel'),
-  onScrollBottomReached: action('scroll-bottom-reached'),
+  onEmojiScrollBottomReached: action('emoji-scroll-bottom-reached'),
 };
 
 // Story Collection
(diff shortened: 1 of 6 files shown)


bianca-artola-dialpad:
Unrelated to this PR, but I noticed that the _scroll_bottom_reached_ event is also triggered when custom emojis aren't present. Was this intentional? Maybe for a future improvement, we could adjust it to only fire when custom emojis are visible. Any thoughts @iropolo?

packages/dialtone-vue2/recipes/conversation_view/message_input/message_input.stories.js
```diff
@@ -190,6 +190,12 @@ export const argTypesData = {
       disable: true,
     },
   },
+
+  onScrollBottomReached: {
```
braddialpad: We should change the name of this to relate to emoji in some way, `onEmojiScrollBottomReached` since it's possible for the message input itself to be scrollable as well.

bianca-artola-dialpad: sure done

braddialpad (approved):

PR #874, opened 2025-08-19
feat(css-utility): DLT-2715 add default color for border css utilities

francisrupert:
# Add default color to border utilities

image

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2715

## :book: Description

`d-ba` (and other `d-b` utilities) have a default border of `current`, inheriting the foreground color of the element. This is never intended, and always overridden with a `d-bc-{color}`. 

**Solution:** default the border color to `var(--dt-color-border-default)`. This is a likely smart default, and could still be overridden with something else, e.g. `var(--dt-color-border-critical)`.

## :pencil: Checklist

For all PRs:

For all CSS changes:

### DIFF
--- a/packages/dialtone-css/lib/build/less/utilities/borders.less
+++ b/packages/dialtone-css/lib/build/less/utilities/borders.less
@@ -22,11 +22,27 @@
 //  ============================================================================
 //  $$ DIRECTION
 //  ----------------------------------------------------------------------------
-.d-ba { border: var(--dt-size-100) solid !important; }
-.d-bt { border-top: var(--dt-size-100) solid !important; }
-.d-br { border-right: var(--dt-size-100) solid !important; }
-.d-bb { border-bottom: var(--dt-size-100) solid !important; }
-.d-bl { border-left: var(--dt-size-100) solid !important; }
+.d-ba {
+  border-color: var(--dt-color-border-default) !important;
+  border-style: solid !important;
+  border-width: var(--dt-size-border-100) !important;
+}
+
+.d-bt {
+  border-top: var(--dt-size-border-100) solid var(--dt-color-border-default) !important;
+}
+
+.d-br {
+  border-right: var(--dt-size-border-100) solid var(--dt-color-border-default) !important;
+}
+
+.d-bb {
+  border-bottom: var(--dt-size-border-100) solid var(--dt-color-border-default) !important;
+}
+
+.d-bl {
+  border-left: var(--dt-size-border-100) solid var(--dt-color-border-default) !important;
+}
 
 .d-bx { .d-br(); .d-bl(); }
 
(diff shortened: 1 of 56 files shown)


packages/dialtone-css/lib/build/less/utilities/borders.less
francisrupert, line 1: This is the actual change to review. All other changes are just removing the redundant `d-bc-default` since it's already baked in for those doc uses.

packages/dialtone-css/lib/build/less/utilities/borders.less
```diff
@@ -22,11 +22,27 @@
...
 //  ----------------------------------------------------------------------------
-.d-ba { border: var(--dt-size-100) solid !important; }
-.d-bt { border-top: var(--dt-size-100) solid !important; }
-.d-br { border-right: var(--dt-size-100) solid !important; }
-.d-bb { border-bottom: var(--dt-size-100) solid !important; }
-.d-bl { border-left: var(--dt-size-100) solid !important; }
+.d-ba {
+  border-color: var(--dt-color-border-default) !important;
+  border-style: solid !important;
+  border-width: var(--dt-size-border-100) !important;
```
francisrupert, line 28: Each property must be separately declared since using shorthand likely may not be overridden by a CSS Utilitie, i.e.

```css
border: var(--dt-size-border-100) solid var(--dt-color-border-default) !important; // don't use shorthand in CSS Utilities
```

packages/dialtone-css/lib/build/less/utilities/borders.less
```diff
@@ -22,11 +22,27 @@
...
+  border-right: var(--dt-size-border-100) solid var(--dt-color-border-default) !important;
+}
+
+.d-bb {
+  border-bottom: var(--dt-size-border-100) solid var(--dt-color-border-default) !important;
+}
+
+.d-bl {
+  border-left: var(--dt-size-border-100) solid var(--dt-color-border-default) !important;
+}
```
juliodialpad, line 45: Shouldn't all of this properties be expanded too?

juliodialpad (approved):
My only concern is that it might break some borders on product if they were relying on the border inheriting by default the current color.

Aside from my comment, this looks good to me.

braddialpad:
> My only concern is that it might break some borders on product if they were relying on the border inheriting by default the current color.
> 
> Aside from my comment, this looks good to me.

Yes I am also worried about this.. Is this something we should include in next instead?

braddialpad:
> image

Such a perfect image for this, lol

packages/dialtone-css/lib/build/less/utilities/borders.less
francisrupert: Huh, could have sworn I did that. Anyway, done in 72a14580db9f943b8c626c6e3a57c6f218ad4b1c

francisrupert:
> Is this something we should include in next instead?

Actually yeah. Since I based it off of `staging` and I don't feel like pulling out the commits to make for a clean merge to `next`, perhaps I can wait to merge this one until after `next` is caught up with `staging`..?

francisrupert:
> Actually yeah. Since I based it off of `staging` and I don't feel like pulling out the commits to make for a clean merge to `next`, perhaps I can wait to merge this one until after `next` is caught up with `staging`..?

Changed to target `next` branch. Diff view is unchanged since it's caught up to `staging` now – thanks @juliodialpad

PR #875, opened 2025-08-19
fix(combobox-multi-select): NO-JIRA prevent losing focus on input

ninamarina:
# fix(combobox-multi-select): NO-JIRA prevent losing focus on input

- [x] Fix

## :book: Jira Ticket
Related to [DP-153208]

## :book: Description
This fixes the bug described in [this thread](https://dialpad.com/app/messages/agxzfnViZXItdm9pY2VyGAsSC1RleHRNZXNzYWdlGICAt7jvpesLDA).
The problem was that the listener was attached to the `keyup` event and moving the cursor / deleting a character happens on `keydown`. That's why when deleting the last character also the chip was focused and the input no longer on focus.

While working on this component I found other issues like:
- When hitting enter an element is created even if it's not part of the list and I am not sure that is intended, will investigate this
- In Vue 2, when selecting part of the text, a chip is created. This is due to the emitted event named `@select` being the same as the native event `select` that is fired when selecting text in the input. This is low priority since it doesn't happen in Vue 3.

## :pencil: Checklist

For all PRs:

For all Vue changes:

### Before
multiselect-bef

### Now
multiselect-now

[DP-153208]: https://dialpad.atlassian.net/browse/DP-153208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dialtone-vue2/components/chip/chip.test.js
+++ b/packages/dialtone-vue2/components/chip/chip.test.js
@@ -147,7 +147,7 @@ describe('DtChip Tests', () => {
 
     describe('When delete is pressed on a chip', () => {
       it('should emit close event', async () => {
-        await chip.trigger('keyup', { code: 'DELETE' });
+        await chip.trigger('keydown', { code: 'DELETE' });
 
         expect('close' in wrapper.emitted()).toBeTruthy();
       });
--- a/packages/dialtone-vue2/components/chip/chip.vue
+++ b/packages/dialtone-vue2/components/chip/chip.vue
@@ -164,6 +164,14 @@ export default {
      * @type {KeyboardEvent}
      */
     'keyup',
+
+    /**
+     * Native chip key down event
+     *
+     * @event keydown
+     * @type {KeyboardEvent}
+     */
+    'keydown',
   ],
 
   data () {
@@ -181,12 +189,16 @@ export default {
           if (this.interactive) this.$emit('click', event);
         },
 
-        keyup: event => {
+        keydown: event => {
           if (event.code?.toLowerCase() === 'delete') {
             this.onClose();
           } else {
-            this.$emit('keyup', event);
+            this.$emit('keydown', event);
           }
+        },
+
+        keyup: event => {
+          this.$emit('keyup', event);
         },
       };
     },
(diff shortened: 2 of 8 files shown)


braddialpad:
> While working on this component I found other issues like:

>When hitting enter an element is created even if it's not part of the list and I am not sure that is intended, will investigate this

This is actually just logic in the storybook for the purposes of example `combobox_multi_select_default.story.vue`. In actual usage the consumer decides what they want to do upon "select"

> In Vue 2, when selecting part of the text, a chip is created. This is due to the emitted event named @select being the same as the native event select that is fired when selecting text in the input. This is low priority since it doesn't happen in Vue 3.

good find... that's a pretty terrible bug. surprised it hasn't been discovered sooner

braddialpad (approved):
thanks!

ninamarina:
> > While working on this component I found other issues like:
> 
> > When hitting enter an element is created even if it's not part of the list and I am not sure that is intended, will investigate this
> 
> This is actually just logic in the storybook for the purposes of example `combobox_multi_select_default.story.vue`. In actual usage the consumer decides what they want to do upon "select"

I've investigated this. In our storybook it doesn't work as expected because `combobox.vue` doesn't emit the `select` event if nothing is highlighted in the list:

```
onEnterKey () {
      if (this.loading || this.emptyList) return;

      if (this.highlightIndex >= 0) {
        this.$emit('select', this.highlightIndex);

        if (this.clickOnSelect) {
          this.activeItemEl?.click();
        }
      }
    },
```

But there's a bug in which sometimes `this.highlightIndex` is not correctly updated, that's why it accidentally works sometimes in storybook Vue 2. In Vue 3, upon `enter` it deletes the entered text.

I was worried that the consumer was taking advantage of the bug, but they are correctly hooking into the `keyup.enter` event directly, that's why for them it works as expected: https://github.com/gluru/js-core/blob/bb1a18f03325cde353678d59535d37128009f1ff/services/dx-console/frontend/src/components/base-combobox/BaseCombobox.vue#L13C9-L13C21

maysamkf (approved):

braddialpad:
> But there's a bug in which sometimes `this.highlightIndex` is not correctly updated, that's why it accidentally works sometimes in storybook Vue 2. In Vue 3, upon `enter` it deletes the entered text.

Interesting.. we should fix this one.

PR #915, opened 2025-09-12
refactor(emoji-picker): DLT-2776 new prop showAddEmojiButton

iropolo:
# DLT-2776 new prop showAddEmojiButton

- [ ] Fix
- [ ] Feature
- [ ] Performance Improvement
- [x] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2776

## :book: Description

We update add-emoji button logic with a new prop `showAddEmojiButton`. Product can use this prop with the experiment `hasCustomEmojiExperiment` to still handle this behavior.

## :bulb: Context

Previously we show this button if there is some custom emoji to show.
But if there is no custom emoji, how the user will add one? 🤪

This was probably some old flag that we use to hide this feature.
But now we need to update it!

### DIFF
--- a/packages/dialtone-vue2/components/emoji_picker/emoji_picker.stories.js
+++ b/packages/dialtone-vue2/components/emoji_picker/emoji_picker.stories.js
@@ -65,6 +65,7 @@ export const argsData = {
   customEmojis,
   skinTone: 'Default',
   showPopover: false,
+  showAddEmojiButton: true,
 };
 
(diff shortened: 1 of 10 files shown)


braddialpad (approved):
great thanks!

bianca-artola-dialpad (approved):
thank you 🙏🏽 
I will update it on firespotter side

PR #937, opened 2025-09-26
fix: DLT-2784 wiz warnings

juliodialpad:
# Fix Wiz dependencies warnings

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2784

## :book: Description

- Updated dependency versions to the recommended ones on Wiz
- Removed dependencies that we're no longer needed

⚠️ I tried to test as much scenarios as possible and fixed them... But as many dependencies were updated and We're overriding some of them with version that might not be fully compatible, I'll suggest you test it thoroughly and keep an eye on this PR in case you need to rollback. ⚠️

## :bulb: Context

- Wiz raised warnings about dependencies with vulnerabilities.
- Many of those dependencies are dependencies or peerDependencies of old packages that we can't update, is hard or we missed some things to update (like storybook, stylelint, eslint)

## :pencil: Checklist

For all PRs:


iropolo (approved):
We need to be very careful about this update on following triage.
It could bring some hidden unwanted changes.

But at least we will have updated dependencies, it is always a good choice!

braddialpad (approved):
Don't really like the overrides, but guessing there is no other way.

ninarepetto (approved):

PR #941, opened 2025-10-03
feat(modal): DLT-2789 add appendTo prop

braddialpad:
# feat(modal): DLT-2789 add appendTo prop

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2789

## :book: Description

Added support for appendTo to modal

## :bulb: Context

Up to this point all our modals have been rendering inline, not great. Had a request for this from @AlexLamDialpad who was having some strange behaviour when triggering a modal in a table.

## :pencil: Checklist

For all PRs:

For all Vue changes:

For all CSS changes:

## :link: Sources

vue-simple-portal is needed for vue 2. In vue 3 it is built in with `<Teleport>`
https://github.com/LinusBorg/vue-simple-portal

### DIFF
--- a/packages/dialtone-vue2/components/modal/modal.stories.js
+++ b/packages/dialtone-vue2/components/modal/modal.stories.js
@@ -106,6 +106,12 @@ export const argTypesData = {
       type: 'boolean',
     },
   },
+  appendTo: {
+    options: [undefined, 'body', 'html'],
+    control: {
+      type: 'select',
+    },
+  },
 
   // Events
   'update:show': {
--- a/packages/dialtone-vue2/components/modal/modal.vue
+++ b/packages/dialtone-vue2/components/modal/modal.vue
@@ -1,116 +1,122 @@
 <template>
-  <dt-lazy-show
-    transition="d-zoom"
-    :show="show"
-    :class="[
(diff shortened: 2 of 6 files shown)


iropolo (approved):
Looks good!

francisrupert (approved):
Nice yeah. We've been doing `<Teleport>` [in Beacon](https://github.com/search?q=repo%3Adialpad%2Fbeacon-app%20teleport&type=code) for same reason.

So of course it should be a prop, though ideally `body` would be the default. Would that really be a breaking change? Like, what reason would DtModal **_not_** be rendered from `<body>`?

braddialpad:
> So of course it should be a prop, though ideally `body` would be the default. Would that really be a breaking change? Like, what reason would DtModal **_not_** be rendered from `<body>`?

It's the fact that prior to this it always rendered inline. In all cases modal SHOULD be rendered at body, I agree with that, however changing it from inline to body could make certain implementations break. TBH most of them would probably still work fine but I have enough experience with these types of changes to know that when we make a change to a component with a large amount of usage like this it always breaks something, somewhere.

PR #943, opened 2025-10-08
fix(popover): DLT-2795 add aria-hidden attribute on popover

iropolo:
# Add aria-hidden attribute on popover

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2795

## :book: Description

CSS styles need to have aria-hidden false attribute set to being applied.
We are adding this attribute to the HTML.

## :bulb: Context

This attribute is necessary to apply the correct CSS styles.
In this specific case we want to avoid interaction with elements below it, with an invisible overlay.

Previously without fix:

https://github.com/user-attachments/assets/5763e572-f828-4dc7-914a-1136989585da

With this fix:

https://github.com/user-attachments/assets/169b3937-7914-4084-8e0c-1e13d9ec9a59

### DIFF
--- a/packages/dialtone-vue2/components/popover/popover.vue
+++ b/packages/dialtone-vue2/components/popover/popover.vue
@@ -4,6 +4,7 @@
     <portal v-if="modal && isOpen">
       <div
         class="d-modal--transparent"
+        aria-hidden="false"
         @click.prevent.stop
       />
     </portal>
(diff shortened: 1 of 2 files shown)


francisrupert (approved):
Confirmed in DtDropdown too. Thanks!

After you merge, can you merge the `staging` result into `next` and publish `next` so it's available for where we're using the `next` release?

/c @hynes-dialpad

braddialpad (approved):

PR #938, opened 2025-09-29
feat(tokens): DLT-1977 add high contrast design tokens and theme

francisrupert:
# DLT-1977 add high contrast design tokens and theme

image

## :hammer_and_wrench: Type Of Change

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-1977

## :book: Description

> [!NOTE]
> this targets `next` branch 

**Created High Contrast Design Tokens**
- Implemented separate high contrast token files for light and dark modes (`tokens-contrast-high-light.css` and `tokens-contrast-high-dark.css`), and js equivalent. 

**Added Contrast as third dimension**
- Extended the existing theming system (Theme × Brand) w/ third `contrast` dimension, i.e. `data-dt-contrast="default||high"`.
- Users can independently control light/dark mode, brand theme (dp/tmo/aegean/etc), and contrast level (default/high) as separate concerns.

**Added Contrast Toggle to Documentation Site**
- Created a new contrast toggle button in the documentation navbar (hidden from rendering w/ `hidden` attribute). 
- Hold down `CMD + SHIFT` while clicking theme button (`CTRL + SHIFT` on Windows).
- Toggle works independently of other theme controls, and persists w/ localStorage.
- Add the `shell` design tokens to the Doc site wrapper. This only serves our purposes and likely won't be an option for visitors to DT site since the theme toggle control is purposely hidden. The rendering result for default `dp` theme is 97% identical to before.

**Expanded Theme Collection in Demo Site**
- Increased available themes from just Dialpad (dp) to 15+ themes, e.g. tmo, Aegean, 122, color-assistive, etc.
- All numbered themes (101-137) are available but commented out for performance – it's _really_ slow on first load. This goes on `next` branch, so not entirely concerned. Would be whittled down to fewer than a dozen once we select the themes we want to ship (including which T-Mobile likes or would like to add).

**Maintained Backward Compatibility**
- Preserved existing `setTheme(theme, rootNode)` API while adding optional contrast third parameter. 

## :pencil: Checklist

For all PRs:

- [x] I have ensured no private Dialpad links or info are in the code or pull request description (Dialtone is a public repo!).
- [x] I have reviewed my changes.
- [ ] I have added all relevant documentation.
- [ ] I have considered the performance impact of my change.

For all Vue changes:

- [ ] I have added / updated unit tests.
[...]

PR #945, opened 2025-10-10
feat(tokens): DLT-2802 DLT-2563 layered theming system and mode island component

francisrupert:
# Layered Theming System & Mode Island Component

lOXwCe

## 📋 Jira Ticket

* https://dialpad.atlassian.net/browse/DLT-2802
* https://dialpad.atlassian.net/browse/DLT-2563

## :book: Description

* Refactored theming system from 4 monolithic themes to layered architecture
* Light/dark mode switching via HTML attributes
* New Mode Island component for nesting light/dark sections (Vue 2 & Vue 3)
* Includes 50+ exploratory themes

The old theming system required loading complete 628KB CSS files for each theme (née brand), and swapping them entirely out to change modes. Performance bottleneck and made targeted multi-theme (shell) support impractical.

The new layered system treats DP as the base (since it's used 90%+ of the time) and applies other themes as small overrides. Mode switching primarily occurs merely via toggling HTML attributes instead of CSS file swaps, making it instant. A new "DtModeIsland" component for nested light/dark sections within the same page.

**Backward compatibility:** The legacy `setTheme()` API continues to work unchanged. Auto-detection determines whether legacy or layered format is being used, allowing teams to migrate at their own pace. Zero breaking changes.  **NEEDS TO BE TESTED AND CONFIRMED**

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| **Theme file size** | 628KB avg | 3.85KB avg | **95.7% smaller** |
| **Mode switch time** | 50-200ms | 0ms | **Instant** |
| **Mode switch data** | 1256KB | 0KB | **Attribute only** |
| **Available themes** | 4 | 51 | **12.75x more** |
| **Initial load (DP)** | 1256KB | 862KB | **31% smaller** |
| **All 51 themes** | 36,210KB | 1,058KB | **97% smaller** |

### 1. Layered Architecture

* **Dimension 1: Base+Brand** → `dp|tmo|aegean|101|137|etc`. (51 total)
* **Dimension 2: Mode:** → `light|dark`
* **Dimension 3: Contrast:** → `default|high`

Total: 200+ possible combinations

```text
Layer 1: tokens-core.css (50KB)           ← Load once
Layer 2: tokens-base-colors.css (185KB)   ← Load once
Layer 3: tokens-dp-colors.css (627KB)     ← Load once
Layer 4: themes/tokens-[name]-colors.css (~0.55KB) ← On demand
Layer 5: tokens-high-contrast.css (3.5KB) ← On demand
```

### 2. New APIs

```js
// Initialize once
initDialtoneTheme(coreTheme, brandTheme, 'light');

// Switch modes (0ms)
setMode('dark');

// Load theme override (0.55KB)
setBrand({ brand: { name: 'tmo', css: TmoColors } });

// Enable high contrast (3.5KB)
setContrast({ contrast: { name: 'high', css: HighContrast } });
```

### 3. New DtModeIsland Component

New component for nested light/dark sections:

```html
<dt-mode-island mode="dark">
  <p>Dark section</p>
  <dt-mode-island mode="inverted">
    <p>Light again</p>
  </dt-mode-island>
</dt-mode-island>
```

## Screencast

**Mode Islands (2:08)**
View at http://localhost:4000/scratch.html

https://github.com/user-attachments/assets/4a18b62f-7073-4727-9cd3-d3c8cc2d7096

**Presence example (00:22)**

https://github.com/user-attachments/assets/197c1d46-1505-4467-902e-02b2043a337b


francisrupert:
> [!NOTE]
> This is showing more changes than it actually is since I based it off https://github.com/dialpad/dialtone/pull/938, which blocks this one.

**UPDATE:** merged `next` into this since https://github.com/dialpad/dialtone/pull/938 was merged.

francisrupert:
Update: added Mode Island doc page:

https://github.com/user-attachments/assets/3e4386ea-61f9-45a1-8bca-69c9a9706c64

francisrupert:
Vue2 and Vue3 preview deploy failing, looking into it. But doc site at least is up:

* https://dialtone.dialpad.com/deploy-previews/pr-945/
* https://dialtone.dialpad.com/deploy-previews/pr-945/components/mode-island.html

francisrupert:
Fixed the dumb failed actions: there's no reason for Storybook to have a fallback to previous theme system. https://github.com/dialpad/dialtone/pull/945/commits/af1cdaed760df251726e42c521db907202e99f96

packages/dialtone-tokens/demo-layered-tokens.html
francisrupert, line 1: I'll probably delete this. Was earliest experimental file.

ninarepetto:
I have to look at this more because there are many changes... I see the approach is very different, all of the tokens are loaded and when changing theme only the html tag changes, was there a reason why we didn't use that approach before?
Also wanted to ask how did you make the theme file size 95.7% smaller

francisrupert:
> ...all of the tokens are loaded and when changing theme only the html tag changes, was there a reason why we didn't use that approach before?

I'm unsure why it wasn't done this before, @braddialpad might have insight there. The key is to understand that that for each `data-dt-theme` (formerly `data-dt-brand` e.g. DP, TMO, Sunflower, etc), both _modes_ are loaded and the `data-dt-mode` is only thing that changes when changing between system/dark/light. 

> Also wanted to ask how did you make the theme file size 95.7% smaller

Layered approach, loading in what's changed between themes/mode/contrast instead of literally swapping out the entire token set. For example: CSS Variables for most colors, Typography, Size/Space, etc don't change at all between themes (brand) and modes, so those are loaded once. Changing theme/mode/contrast only pulls them in as overrides. That's where the 90+% savings are because it doesn't have redundant tokens. 

From this PR's description:

```text
Layer 1: tokens-core.css                 (50KB)    ← Load once
Layer 2: tokens-base-colors.css          (185KB)   ← Load once
Layer 3: tokens-dp-colors.css            (627KB)   ← Load once
Layer 4: themes/tokens-[name]-colors.css (~0.55KB) ← On demand
Layer 5: tokens-high-contrast.css        (3.5KB)   ← On demand
```

Layers 1 through 3 are always present. 
1. Change theme (brand): swap out Layer 4
2. Change mode: instant, no loading
3. Change Contrast: toggle layer 5

braddialpad:
> I see the approach is very different, all of the tokens are loaded and when changing theme only the html tag changes, was there a reason why we didn't use that approach before?

We actually did do this before when we only had 2 themes. When we got more themes we were loading megabytes and megabytes of token data up front so it wasn't viable at all.

packages/dialtone-tokens/themes/config.js
```diff
@@ -24,6 +42,38 @@ export function setTheme (theme, rootNode = document.documentElement, contrastTh
...
+  }
+
+  // Load core tokens only once
+  if (theme.core && !coreTokensLoaded) {
+    _setStyleTag('dialtone-css-core', theme.core, rootNode);
+    coreTokensLoaded = true;
+  }
+
+  // Load base colors only once
+  if (theme.baseColors && !rootNode?.querySelector('#dialtone-css-base-colors')) {
```
braddialpad, line 137: Why do we use coreTokensLoaded to check for existence in the first one, but a querySelector here? Not a big deal, but feels inconsistent.

packages/dialtone-tokens/package.json
```diff
@@ -13,11 +13,18 @@
   "exports": {
     "./*.css": "./dist/css/*.css",
     "./*.less": "./dist/less/*.less",
+    "./dist/css/layered/*": "./dist/css/layered/*",
+    "./dist/css/layered/themes/*": "./dist/css/layered/themes/*",
+    "./dist/css/layered/contrast/*": "./dist/css/layered/contrast/*",
```
braddialpad: I don't think it's necessary to do this at all if the key and value are the exact same. This is just for creating "aliases"

apps/dialtone-documentation/docs/.vuepress/theme/client.js
```diff
@@ -172,127 +181,131 @@ async function importDocumentation (app) {
...
-      // '@dialpad/dialtone-tokens/themes/134-light.js': (await import('@dialpad/dialtone-tokens/themes/134-light')),
-      // '@dialpad/dialtone-tokens/themes/134-dark.js': (await import('@dialpad/dialtone-tokens/themes/134-dark')),
-      // '@dialpad/dialtone-tokens/themes/135-light.js': (await import('@dialpad/dialtone-tokens/themes/135-light')),
-      // '@dialpad/dialtone-tokens/themes/135-dark.js': (await import('@dialpad/dialtone-tokens/themes/135-dark')),
-      // '@dialpad/dialtone-tokens/themes/136-light.js': (await import('@dialpad/dialtone-tokens/themes/136-light')),
-      // '@dialpad/dialtone-tokens/themes/136-dark.js': (await import('@dialpad/dialtone-tokens/themes/136-dark')),
-      '@dialpad/dialtone-tokens/themes/137-light.js': (await import('@dialpad/dialtone-tokens/themes/137-light')),
-      '@dialpad/dialtone-tokens/themes/137-dark.js': (await import('@dialpad/dialtone-tokens/themes/137-dark')),
+    // Import all themes explicitly (Vite doesn't like dynamic imports)
+    const themeModules = await Promise.all([
```
braddialpad, line 187: Could probably import these dynamically with a loop since there are so many, but not important for now I suppose.

packages/dialtone-tokens/build-layered.js
```diff
@@ -0,0 +1,255 @@
...
+  }
+
+  console.log('\n' + '='.repeat(60));
+  console.log(`Total reduction: ${((1 - totalOptimizedSize / totalOriginalSize) * 100).toFixed(1)}%`);
+  console.log(`Original total: ${(totalOriginalSize / 1024).toFixed(2)}KB`);
+  console.log(`Optimized total: ${(totalOptimizedSize / 1024).toFixed(2)}KB`);
+  console.log(`Savings: ${((totalOriginalSize - totalOptimizedSize) / 1024).toFixed(2)}KB`);
+
+  // Step 3: Clean up redundant full theme files (keep only dp, base, core files)
+  console.log('\nStep 3: Cleaning up redundant full theme files...\n');
```
braddialpad, line 241: sorta confused why we are generating theme files and then deleting them, is this just for testing purposes?

packages/dialtone-tokens/build-layered.js
```diff
@@ -0,0 +1,255 @@
...
+
+  const css = readFileSync(file, 'utf8');
+  const result = await postCssProcessor.process(css, { from: file, to: file });
+  writeFileSync(file, result.css);
+}
+
+/**
+ * Main function to generate optimized theme files
+ */
+async function main() {
```
braddialpad: This function is much too long, split it into some logical functions to reduce the complexity.

some examples:

```
generateThemes(themeFiles)
clean()
processHighContrast()
outputSummary()
```

packages/dialtone-tokens/build-layered.js
```diff
@@ -0,0 +1,255 @@
...
+          .filter(line => line.trim() && !line.includes('color-scheme'))
+          .map(line => '  ' + line.trim())
+          .join('\n');
+      };
+
+      const lightVars = extractVars(lightContent);
+      const darkVars = extractVars(darkContent);
+
+      const combined = `/**
+ * Do not edit directly, this file was auto-generated.
```
braddialpad, line 306: Is this autogenerated? If so it shouldn't be committed to source control.

braddialpad (commented):
Did an initial review and will do another one tomorrow. There is a ton of complexity here and I'm not sure I have a good understanding of how it all works even after reading the code.

packages/dialtone-tokens/themes/config.js
```diff
@@ -61,3 +111,89 @@ function _setThemeAttributeOnRoot (theme, brand, rootNode) {
...
+  // Load core tokens (once)
+  if (coreTheme.core) {
+    _setStyleTag('dialtone-css-core', coreTheme.core, rootNode);
+    coreTokensLoaded = true;
+  }
+
+  // Load base colors (once)
+  if (coreTheme.baseColors) {
+    _setStyleTag('dialtone-css-base-colors', coreTheme.baseColors, rootNode);
+  }
```
ninarepetto, line 490: Core tokens and base colors should always be loaded for all brands, right? in that case and if I understood correctly, the first parameter should not be necessary.

packages/dialtone-tokens/themes/config.js
```diff
@@ -61,3 +111,89 @@ function _setThemeAttributeOnRoot (theme, brand, rootNode) {
...
+
+  // Set initial mode
+  setMode(mode, rootNode);
+
+  // Set initial brand
+  setBrand(brandTheme, rootNode);
+
+  // Set default contrast
+  rootNode?.setAttribute('data-dt-contrast', 'default');
+}
```
ninarepetto, line 561: I am a bit confused because these utils are used in storybook but not in the doc site, where the theme, brand and contrast are manually set (in `useThemeManager`), why is that?

francisrupert:
@braddialpad, @ninarepetto Thanks for the feedback! I'm hoping to jump back into this Friday.

packages/dialtone-tokens/themes/config.js
francisrupert: **For core tokens:** it's module-level variable since they're loaded once globally, regardless of which root node they're applied to (i.e. ModeIsland). They're the foundation and never change.

**For base colors:** querySelector since they're loaded once per root node, e.g. should there be a shadow dom that will need its own style tags while still avoid duplicate style elements in same root.

packages/dialtone-tokens/package.json
francisrupert: Yeah, this one drove me up the wall. I initially took it for granted the `./*` catch-all, but unlike node, Vite specifically needs these explicit exports for _nested_ directories. 

I simplified them in 163912402429d4214147539eec6f18eb41969b43, and retained the aliases but clarified them.

apps/dialtone-documentation/docs/.vuepress/theme/client.js
francisrupert: I was going to do that at first, but wanted to retain manual control for the moment. Once we pare it down to the ~dozen themes we actually ship, we can do so.

packages/dialtone-tokens/build-layered.js
francisrupert: Heh, yeah this is kind of the secret sauce to strip down to bare layered essentials. When I first started, I was trying to wrestle with modifying Style Dictionary to output only the diffs in order to calculate which tokens differ, and then generate optimized overwrite files. But that got really really complex and breaks Style Dictionary's architecture.

So what's going on here is...

1. Generate **full** theme files for all themes
2. Compare each theme against the DP base theme
  a. Calculate differences (overrides) for each theme
  b. Generate the optimized override files in `themes/` directory
  c. e.g. `themes/tokens-125.css` (~.5KB, just the differences!)
3. Subsequently delete the **full** theme files from step 1

So I am allowing the style dictionary build process to generate the complete token files, but those are the ones that we don't want to ship because they are each extremely huge, and redundant, and the layer system only needs the small overrides. 

Comments added to explain, 9cc269c2a1e5bbda071caa8f3b60ec072ceaa45c

packages/dialtone-tokens/build-layered.js
francisrupert: Thx. Done in cb3978e01876e5d1f418861a8b7d9124f9676adb

packages/dialtone-tokens/build-layered.js
francisrupert: Yeah, the resulting file this comment appears in is auto-generated to dist directory which is already ignored.

packages/dialtone-tokens/themes/config.js
francisrupert: Good catch. I was going with defensive conditionals out of habit early on; no longer necessary. 954061df1c5044f95aae7c853897ef2dc91137ad

packages/dialtone-tokens/themes/config.js
francisrupert: Doc site uses a Vue composable with additional features like media query'ing `system` mode support and UI helpers. Storybook uses the simpler framework-agnostic utilities from config.js. They do the same thing internally. They evolved separately. I did an explore for consolidating them, but that's quite a refactor that doesn't feel worth it right now.

packages/dialtone-tokens/build-layered.js
francisrupert: And comments improved in https://github.com/dialpad/dialtone/pull/945/commits/8ef15ebdc70fcb943e7c222c61a247213cbde356

apps/dialtone-documentation/docs/.vuepress/config.js
```diff
@@ -96,6 +96,9 @@ export default defineUserConfig({
     '@views': path.resolve(__dirname, './views'),
     '@mixins': path.resolve(__dirname, './common/mixins/'),
     '@utilities': path.resolve(__dirname, './common/utilities.js'),
+    '@composables': path.resolve(__dirname, './theme/composables'),
+    '@theme-constants': path.resolve(__dirname, './theme/constants'),
+    '@theme-utils': path.resolve(__dirname, './theme/utils'),
```
iropolo: I cannot find usage of these two

.husky/post-checkout
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+. "$(dirname "$0")/_/husky.sh"
```
iropolo: I guess this could be better integrated in our current start command instead of running it over git.
Some `clean` script.

iropolo: I can try some approach later

apps/dialtone-documentation/docs/.vuepress/config.js
francisrupert: Thanks, they're no longer used. Removed in 6ce52c195fbb30a95c61aad81c53339bc8b0b8bc

The files themselves remain, imported w/ relative in `useThemeManager.js`

.husky/post-checkout
francisrupert: OK, I'll give that a whirl.

packages/dialtone-tokens/build-layered.js
```diff
@@ -0,0 +1,400 @@
...
+
+  await generateFullTokens(outputDir);
+  const stats = await generateThemeOverrides(outputDir);
+  await cleanupFullThemeFiles(outputDir);
+  await processHighContrast(outputDir);
+  outputBuildSummary(stats);
+}
+
+// Export for use in build.js
+export { main as buildLayeredTokens };
```
iropolo: this is redundant, call the function `buildLayeredTokens`

.husky/post-checkout
francisrupert: This was bugging the crap out of me. I wound up with this as a fairly aggressive workaround nx cache when switching between branches with different build results. I'll stab at a `clean` script that can be opted in, and not necessarily part of `start`, since `start` and other build scripts already clear them out.

francisrupert: Updated in [9444aa3](https://github.com/dialpad/dialtone/pull/945/commits/9444aa316c53b4bdace2b93ba951d136529eaf93).

Removed git hook and added `clean` scripts.

francisrupert: Moved to PR https://github.com/dialpad/dialtone/pull/958

packages/dialtone-tokens/build-layered.js
francisrupert: Done [ccda6e4](https://github.com/dialpad/dialtone/pull/945/commits/ccda6e43f10735f56096069e514cbee8e2b278f4)

francisrupert:
Toggled visual test labels. Don't want it to continue testing since I already approved its previous builds. If something significant changes, I'll add it back in.

apps/dialtone-documentation/docs/scratch.md
```diff
@@ -0,0 +1,1354 @@
+---
+layout: Blank
+---
```
iropolo, line 3: I will not review scratch.md since seems like is a hidden route for playground. Not documented anywhere.

iropolo: If there is some part that we should take a look, please let us know.

apps/dialtone-documentation/docs/components/mode-island.md
```diff
@@ -0,0 +1,834 @@
...
+                :name="currentModeIconName"
+              />
+            </template>
+          </dt-button>
+        </template>
+        <template #list>
+          <dt-list-item-group
+            heading-class="d-py4 d-px8 d-c-default d-fc-tertiary d-label--sm"
+            heading="Mode"
+          >
```
iropolo, line 36: I can see this changes the entire site configuration.
It could leave a unintended configuration behind.

Since this is a component to `create independent sections with their own color modes.` I will expect it will only change the below examples?




https://github.com/user-attachments/assets/f4b8bb8c-6b34-4d1d-af41-ca67b934bc00

apps/dialtone-documentation/docs/components/presence.md
```diff
@@ -7,6 +7,54 @@ image: assets/images/components/presence.png
...
+        <div class="d-w24 d-h24" style="background: var(--dt-color-black-900);" title="black-900"></div>
+        <div class="d-w24 d-h24" style="background: var(--dt-color-black-950);" title="black-950"></div>
+        <div class="d-w24 d-h24" style="background: var(--dt-color-black-1000);" title="black-1000"></div>
+    </dt-stack>
+    <dt-stack data-dt-mode="dark" gap="400" class="d-bgc-primary d-ba d-p16">
+        <div>
+          <h3>Always Dark Section</h3>
+          <p>This section always renders in dark mode, even when the page is in light mode.</p>
+        </div>
+        <dt-stack direction="row">
```
iropolo: Hmm..

Not sure why this is here... should be in mode-island documentation?
I didn't understand what is showing me related to `presence`... if it is intendeed maybe some short description would be useful.


Apart from that, it seems like some text is completely hidden but the white one is still visible in both mode.
I guess it should be readable or maybe not, but should be consistent in both cases.



https://github.com/user-attachments/assets/cf5b8c7b-0a22-4658-8771-a99e4dd040e5

apps/dialtone-documentation/docs/.vuepress/theme/client.js
```diff
@@ -172,127 +181,131 @@ async function importDocumentation (app) {
...
-    for (const path in dialtoneThemeFiles) {
-      const themeName = path.split('/').pop().split('.').shift();
+    console.log(`Successfully loaded ${Object.keys(themes).length - 1} themes + high contrast`);
 
-      themes[themeName] = dialtoneThemeFiles[path].default;
-    }
-    app.provide('themes', themes)
-  } catch(error) {
+    app.provide('themes', themes);
+    app.provide('layeredTokensEnabled', true);
```
iropolo: Where this provided `layeredTokensEnabled` is used?

I can see there is something about it in our `preview.jsx` but they are not connected to this right?

I know `themes` is being used actively through the documentation with `useThemeManager.js`

apps/dialtone-documentation/docs/.vuepress/theme/components/Navbar.vue
```diff
@@ -117,14 +117,131 @@
...
+        <dt-dropdown-separator />
+        <dt-list-item-group
+          heading-class="d-py4 d-px8 d-c-default d-fc-tertiary d-label--sm"
+          heading="Partner Themes"
+        >
+          <dt-list-item
+            role="menuitem"
+            navigation-type="arrow-keys"
+            @click="setTheme('tmo')"
+          >
```
iropolo, line 167: Funny, this change shows a warning that is a missing declare event of `click` in `emits`.

For some reason there is a `click` event in vue2 but not in vue3.



https://github.com/user-attachments/assets/ca8265c8-7377-41f6-9820-79ea02f31ca0





please add this code (the same we have in vue2)
here: 
packages/dialtone-vue3/components/list_item/list_item.vue

```
/**
     * Native click event
     *
     * @event click
     * @type {PointerEvent | KeyboardEvent}
     */
    'click',
```

On `emits: [`

README.md
```diff
@@ -43,71 +43,210 @@ import "@dialpad/dialtone/css-default-theme";
...
-import { setTheme } from '@dialpad/dialtone/themes/config';
-import DpLight from '@dialpad/dialtone/themes/dp-light';
-setTheme(DpLight);
+import { initDialtoneTheme } from '@dialpad/dialtone/themes/config';
+import CoreTokens from '@dialpad/dialtone-tokens/layered/tokens-core.css';
+import BaseColors from '@dialpad/dialtone-tokens/layered/tokens-base-colors.css';
+import DpColors from '@dialpad/dialtone-tokens/layered/tokens-dp-colors.css';
+
+// Initialize base layers and DP theme
+initDialtoneTheme(
```
ninarepetto: I have a few questions about this function `initDialtoneTheme`:

1- Do we want to allow the consumer to override the coreTokens or baseColors? If that is not a possibility, this doesn't need to be a parameter because `initDialtoneTheme` is defined in `dialtone-tokens`, where the tokens are available so it can load them internally.

2- Regarding the `brand` parameter, I see it receives a `name` and a `css`, what is the relationship between the two? Are we allowing to pass a custom css here? What happens if someone passes `{ brand: { name: 'myBrand', css: DpColors } }`? Or `{ brand: { name: 'dp', css: otherCss } }`? Is there a list of valid brands? if they each correspond to a specific css then I suggest to only pass the name.

3- I see that this function internally calls `setMode` and `setBrand`, but those functions are also exported. I think that could be confusing for consumers, they might call `initDialtoneTheme`, and then `setMode` to set the mode. I suggest we either provide one function to initialize and set everything, or different functions that each set mode, brand, contrast, etc.

4- Since the parameters have a specific structure `{ core: CoreTokens, baseColors: BaseColors }` and `{ brand: { name: 'dp', css: DpColors } }`, they should be validated in the first line of `initDialtoneTheme` and throw an error in case they are not valid, because if not later in `setBrand`:

```
export function setBrand(brandTheme, rootNode = document.documentElement) {
  if (rootNode?.shadowRoot) {
    rootNode = rootNode.shadowRoot;
  }

  if (brandTheme.brand) {
    _setStyleTag('dialtone-css-brand-colors', brandTheme.brand.css, rootNode);
    rootNode?.setAttribute('data-dt-brand', brandTheme.brand.name);
  }
}
```
if `brandTheme` is not correct it will not do anything and fail silently and the consumer wouldn't know that they are using the function incorrectly.

packages/dialtone-tokens/themes/config.js
```diff
@@ -61,3 +111,85 @@ function _setThemeAttributeOnRoot (theme, brand, rootNode) {
...
+
+/**
+ * Initialize Dialtone theme system - call once on app startup
+ * @param coreTheme theme object with core and baseColors properties
+ * @param brandTheme initial brand theme
+ * @param mode initial mode ('light' or 'dark')
+ * @param rootNode optional root element
+ */
+export function initDialtoneTheme(coreTheme, brandTheme, mode = 'light', rootNode = document.documentElement) {
```
ninarepetto: Regarding the `rootNode` parameter, and given the bugs we've been having lately with the integrations not providing the node and setting their styles in the root, I think this is an opportunity to make this not optional. We should always demand a node, so that the integrations pass the shadow node here and Dialpad can simply pass the `documentElement`.

ninarepetto:
Another general question I have is how do we plan to roll this out. Can it be associated to an experiment in dp? So that we can do a progressive rollout? And rollback easily if we find problems in production?

apps/dialtone-documentation/docs/.vuepress/theme/components/Navbar.vue
```diff
@@ -233,8 +334,7 @@
 
 <script setup>
 import { useRoute } from 'vue-router';
-import { onMounted, onUnmounted, inject, computed } from 'vue';
-import { setTheme } from '@dialpad/dialtone-tokens/themes/config';
+import { useThemeManager } from '../composables/useThemeManager';
```
iropolo, line 337: Love the use of composable here.
This is the kind of refactor we need to implement on Dialtone within next steps.

apps/dialtone-documentation/docs/.vuepress/theme/constants/themes.js
```diff
@@ -0,0 +1,114 @@
...
+  'high-desert',
+  'melon',
+  'plum',
+  'sunflower',
+  'verdant-haze',
+  // Numbered themes (experimental - not yet named)
+  '101', '102', '103', '104', '105', '106', '107', '108', '109', '110',
+  '111', '112', '113', '114', '115', '116', '117', '118', '119', '120',
+  '121', '122', '123', '124', '125', '126', '127', '128', '129', '130',
+  '131', '132', '133', '134', '135', '136', '137',
```
iropolo: Could be this replaced by `...NUMBERED_THEMES` ?

It can be outdated if it is not updated correctly.

apps/dialtone-documentation/docs/.vuepress/theme/constants/themes.js
```diff
@@ -0,0 +1,114 @@
...
+  // Named themes (alphabetical)
+  'aegean',
+  'botany',
+  'buttercream',
+  // 'ceruleo',
+  'high-desert',
+  'melon',
+  'plum',
+  'sunflower',
+  'verdant-haze',
```
iropolo: Seems like this can be replaced also with `...NAMED_THEMES`, you got the idea.

packages/dialtone-vue2/components/mode_island/mode_island.mdx
```diff
@@ -0,0 +1,120 @@
+import { Canvas, Story, Controls, Meta } from '@storybook/blocks';
+
+import * as ModeIslandStories from './mode_island.stories';
+
+<Meta of={ModeIslandStories} />
```
iropolo: Docs in storybook are deprecated, we redirect it to the component doc site in https://dialtone.dialpad.com/.

packages/dialtone-vue2/components/mode_island/mode_island.stories.js
```diff
@@ -0,0 +1,295 @@
...
+import DtInput from '@/components/input/input.vue';
+import DtNotice from '@/components/notice/notice.vue';
+import DtStack from '@/components/stack/stack.vue';
+import { createRenderConfig } from '@/common/storybook_utils';
+import { DT_MODE_ISLAND_TYPES } from './mode_island_constants';
+
+export const argsData = {
+  mode: DT_MODE_ISLAND_TYPES.INVERTED,
+  as: 'div',
+};
```
iropolo, line 10: I expect the storybook controls will work in the examples.
I can see there is some change in the code but it doesn't reflect in the examples.


https://github.com/user-attachments/assets/b58cf5b4-b6a4-4635-aa74-de2599197d25

packages/dialtone-vue2/components/mode_island/mode_island.test.js
```diff
@@ -0,0 +1,190 @@
...
+const baseSlots = {
+  default: '<div data-qa="test-content">Test Content</div>',
+};
+
+describe('DtModeIsland Tests', () => {
+  let wrapper;
+  let defaultContent;
+
+  const updateWrapper = (props = {}, slots = {}, attrs = {}) => {
+    wrapper = mount(DtModeIsland, {
```
iropolo: This doesn't follow our writing test convention.

To update props or slots, we have the `mock*` variables.

Check some other test file and update it.

packages/dialtone-vue2/components/mode_island/mode_island.test.js
```diff
@@ -0,0 +1,190 @@
...
+        });
+      }).toThrow('[DtModeIsland] The data-dt-brand attribute is not allowed on mode islands');
+    });
+  });
+
+  describe('Attribute Passthrough Tests', () => {
+    it('should pass through class attribute', async () => {
+      updateWrapper({}, {}, { class: 'd-p16 d-bgc-primary' });
+      expect(wrapper.classes()).toContain('d-p16');
+      expect(wrapper.classes()).toContain('d-bgc-primary');
```
iropolo: This should be like:

```
mockAttrs: { class: 'd-p16 d-bgc-primary' };

updateWrapper();

expect(wrapper.classes()).toContain('d-p16');
expect(wrapper.classes()).toContain('d-bgc-primary');
```

packages/dialtone-vue2/components/mode_island/mode_island.vue
```diff
@@ -0,0 +1,207 @@
+<template>
+  <component
+    :is="as"
+    class="d-mode-island"
+    v-bind="$attrs"
+    :data-dt-mode="computedMode"
+    :data-mode-island-inverted="isInverted ? '' : null"
```
iropolo: Im more on no logic in template.

Lets move this to computed like:

```
...
:data-mode-island-inverted="invertedAttribute"
...

computed: {
   ...
    invertedAttribute () {
      return this.isInverted ? '' : null;
    },
```

packages/dialtone-vue2/components/mode_island/mode_island.vue
```diff
@@ -0,0 +1,207 @@
...
+      dtModeIslandMode: () => this.computedMode,
+    };
+  },
+
+  inject: {
+    parentModeIslandMode: {
+      from: 'dtModeIslandMode',
+      default: null,
+    },
+  },
```
iropolo, line 39: Im okay to have this pattern in vue2, but here in vue3 I would like to move it to composables.

Shared logic should be wrapped in composables, it can be a refactor we can move on in the future.
This is my perspective, no need to do it now. 

When we will move definitely to vue3, probably some of these refactor will be needed and I can share my thoughts.

apps/dialtone-documentation/docs/scratch.md
francisrupert: Yeah, a blank layout template so I could explore. https://dialtone.dialpad.com/deploy-previews/pr-945/scratch.html

I've emptied it out now. 
image

apps/dialtone-documentation/docs/components/mode-island.md
francisrupert: > I can see this changes the entire site configuration. It could leave a unintended configuration behind.

It syncs up with the global one in the to nav bar. It shares the global state, it uses the same composable, and is perfectly synced. Since there's only one source of truth (the `currentMode` and `currentContrast`) there shouldn't be any config collision or risk.

> I will expect it will only change the below examples?

Intentionally applying the theme and mode to the entire site to demo the complete context for the the mode and contrast.

I have thought about a future doc site feature where you can change the mode of each preview using the Mode Island.

apps/dialtone-documentation/docs/components/presence.md
francisrupert: Good catch. That was a leftover old explore. Removed.

apps/dialtone-documentation/docs/.vuepress/theme/client.js
francisrupert: Dead code. Removing, including from storybook.

apps/dialtone-documentation/docs/.vuepress/theme/components/Navbar.vue
francisrupert: Done

packages/dialtone-tokens/package.json
braddialpad: weird

apps/dialtone-documentation/docs/.vuepress/theme/client.js
braddialpad: yeah no prob

apps/dialtone-documentation/docs/.vuepress/theme/constants/themes.js
francisrupert: Yes, I like. Done.

packages/dialtone-vue2/components/mode_island/mode_island.test.js
francisrupert: Done

packages/dialtone-vue2/components/mode_island/mode_island.vue
francisrupert: Good catch. Done

packages/dialtone-vue2/components/mode_island/mode_island.mdx
francisrupert: Done

packages/dialtone-vue2/components/mode_island/mode_island.stories.js
francisrupert: Done

packages/dialtone-tokens/themes/config.js
francisrupert: Feels like overkill and just adds noise. 

rootNode as required won't fix the knowledge gap, and might just makes the API more annoying for 99% of users who won't need shadow DOM.

Those bugs stem from integrations not understanding shadow DOM implementation, and not necessarily from not from having a default parameter. 

I'll address with improved docs, including making shadow dom usage explicit.

francisrupert:
@ninarepetto 

> Another general question I have is how do we plan to roll this out. Can it be associated to an experiment in dp? So that we can do a progressive rollout? And rollback easily if we find problems in production?

This is targeting `next` prerelease. An internal flag experiment is indeed the presumptive approach. Not sure of timing in coordinating with a product engineer team (e.g. Core UX), but it would more likely be done paired with updating the Preferences>Appearance panel updates [seen in Beacon](https://dialpadhq.com/beacon) now.

packages/dialtone-tokens/themes/config.js
braddialpad: This has been such a huge and confusing problem for us and QA when it happens I think we should do something about it. We are moving towards embedded microfrontends more in the future and this problem will become bigger. I don't really think just making it a non optional parameter is enough though. I'm wondering if within that init we can somehow prevent writing the theme if it already exists on that element.

packages/dialtone-tokens/themes/config.js
```diff
@@ -61,3 +111,85 @@ function _setThemeAttributeOnRoot (theme, brand, rootNode) {
...
+ * @param mode 'light' or 'dark'
+ * @param rootNode optional root element
+ */
+export function setMode(mode, rootNode = document.documentElement) {
+  if (mode !== 'light' && mode !== 'dark') {
+    console.warn(`Invalid mode: ${mode}. Must be 'light' or 'dark'`);
+    return;
+  }
+  rootNode?.setAttribute('data-dt-mode', mode);
+}
```
braddialpad, line 239: we'll need the same shadowRoot condition here as in the other setX functions, or else we could have the attribute being applied to different elements in the case it gets applied to the shadowRoot.host

packages/dialtone-tokens/themes/config.js
ninarepetto: The other idea I got when talking to Brad about this is if someone calls the init function to set the theme with the same brand and mode, don't set it again. And that would prevent embedded apps from setting the root with outdated tokens, and would still work for them when testing the standalone version.

francisrupert: ok, i'll play with it. I'll add some tracking to initialization for each, and skip redundant initialization when brand/mode unchanged.

francisrupert: Implemented rootNode tracking for each rootNode in 72e14e296

- Tracks initialization state (brand/mode) per rootNode
- Warns on duplicate initialization (same params or different params)
- Detects embedded app scenario (multiple calls on document.documentElement) with specific warning
- Kept rootNode optional for backward compatibility
- Added resetTheme() utility for cleanup

packages/dialtone-tokens/themes/config.js
francisrupert: Addressed in c56a06f79. shadowRoot detection added to setMode()

Now warns if s`hadowRoot` passed directly, and auto-detects `rootNode?.shadowRoot`. Matches pattern in `setBrand()`, `setContrast()`, `initDialtoneTheme(`)

README.md
francisrupert: Thanks, I needed that. Addressed in aae26c67a and ae6df4349.

> 1- Do we want to allow the consumer to override the coreTokens or baseColors? ...

Made `coreTheme` internal, no reason for it to be exposed. Now e.g. `initDialtoneTheme(Dp, 'light')` instead of `initDialtoneTheme(Core, Dp, 'light')`. Core tokens auto-loaded internally no matter what anyway.

> 2- Regarding the brand parameter, I see it receives a name and a css...

Left object structure alone. Brand themes export `{brand: {name, css}}` format. This supports both official themes (e.g. Dp, Tmo, Melon, et al) and custom brand
overrides. Did add validation though, that was needed.

> 3- I see that this function internally calls `setMode` and `setBrand`, but those functions are also exported...

Intentional design of init and setters pattern. `initDialtoneTheme()` for setup, `setMode()`/`setBrand()`/`setContrast()` for dynamic
switching. That's the intent of the pattern.

> 4- Since the parameters have a specific structure...if brandTheme is not correct ... it fail silently...

Added validation and nixed silent fail. Invalid params throw errors (e.g. missing `brandTheme`, wrong structure, invlid `mode`).

Also added straight-forward theming starting points docss and examples to main `README.md`.

francisrupert:
Hoo boy, you all are keeping me on my toes. Responses added to your comments, and re-requested review.

francisrupert:
Also added a short "Theme and Mode" section to [Getting Started](https://dialtone.dialpad.com/deploy-previews/pr-945/guides/getting-started/) and a dedicated [Theme and Mode](https://dialtone.dialpad.com/deploy-previews/pr-945/guides/theme-and-mode/) page.

I think the overall organization/IA of those pages can be improved, but fine for now.

packages/dialtone-vue3/components/mode_island/mode_island.vue
```diff
@@ -0,0 +1,212 @@
...
+
+      this.contrastObserver = new MutationObserver(callback);
+      this.contrastObserver.observe(document.documentElement, config);
+    },
+
+    setupModeObserver () {
+      const config = {
+        attributes: true,
+        attributeFilter: ['data-dt-mode'],
+        subtree: false,
```
iropolo, line 181: pass the callback function direclty inline fo the mutationObserver

```
this.modeObserver = new MutationObserver((mutationsList) => {
        for (const mutation of mutationsList) {
          if (mutation.type === 'attributes' && mutation.attributeName === 'data-dt-mode') {
            // Recalculate and update the reactive data property
            this.calculatedMode = this.calculateInvertedMode();
          }
        }
      });
```

packages/dialtone-vue3/components/mode_island/mode_island.vue
```diff
@@ -0,0 +1,212 @@
...
+      const callback = (mutationsList) => {
+        for (const mutation of mutationsList) {
+          if (mutation.type === 'attributes' && mutation.attributeName === 'data-dt-contrast') {
+            this.currentContrast = getRootContrast();
+          }
+        }
+      };
+
+      this.contrastObserver = new MutationObserver(callback);
+      this.contrastObserver.observe(document.documentElement, config);
```
iropolo: Replace it with:

```
this.contrastObserver = new MutationObserver((mutationsList) => {
        for (const mutation of mutationsList) {
          if (mutation.type === 'attributes' && mutation.attributeName === 'data-dt-contrast') {
            this.currentContrast = getRootContrast();
          }
        }
      });

      this.contrastObserver.observe(document.documentElement, {
        attributes: true,
        attributeFilter: ['data-dt-contrast'],
      });
```

iropolo:
Apply these last changes to vue3 too, it is some simplification of what AI always does.
The while loop also is very AI, but should be okay, it has the safe guard of `?.` on `this.elementRef`

Im okay with this PR overall. It will go on `next` first so it will have some interaction after moving it to staging. 

I didn't go further on dialtone-tokens changes. I will let those to @braddialpad

iropolo (approved):

README.md
ninarepetto: @francisrupert thanks for answering, I am still learning about our token's system. Is looking really good, will continue to review today.

packages/dialtone-vue3/components/mode_island/mode_island.vue
francisrupert: Address in https://github.com/dialpad/dialtone/pull/945/commits/8a18c9de9df6873f54b94bf49f143fca285ecb61 and https://github.com/dialpad/dialtone/pull/945/commits/a668471a4f9c2e08a6dd320cf95f2f2375e7d32c

packages/dialtone-vue3/components/mode_island/mode_island.vue
francisrupert: Address in https://github.com/dialpad/dialtone/pull/945/commits/8a18c9de9df6873f54b94bf49f143fca285ecb61 and https://github.com/dialpad/dialtone/pull/945/commits/a668471a4f9c2e08a6dd320cf95f2f2375e7d32c

apps/dialtone-documentation/docs/guides/theme-and-mode/index.md
```diff
@@ -0,0 +1,395 @@
...
+- **Support Shadow DOM** for web components and micro-frontends
+
+The system uses a layered architecture that loads core tokens once and applies theme-specific overrides, resulting in better performance and a tighter bundle size compared to legacy approaches.
+
+## Quick Start
+
+Install the required packages:
+
+```shell
+npm install @dialpad/dialtone @dialpad/dialtone-tokens
```
ninarepetto: `npm install @dialpad/dialtone` should be enough

packages/dialtone-tokens/themes/config.js
```diff
@@ -61,3 +190,341 @@ function _setThemeAttributeOnRoot (theme, brand, rootNode) {
...
+ *
+ * @example
+ * // In Web Components
+ * class MyWidget extends HTMLElement {
+ *   switchBrand(brandTheme) {
+ *     setBrand(brandTheme, this);
+ *   }
+ * }
+ */
+export function setBrand(brandTheme, rootNode = document.documentElement) {
```
ninarepetto, line 264: Since `setBrand` can be called separately from `initDialtoneTheme` you need to add the same validations for the `brandTheme` parameter here.

packages/dialtone-tokens/themes/config.js
```diff
@@ -39,7 +164,11 @@ function _setStyleTag (id, content, rootNode) {
       rootNode?.appendChild(style);
     }
   } else {
-    rootNode.querySelector('#' + id).innerHTML = content;
+    const existingTag = rootNode.querySelector('#' + id);
+    // Only update if content changed (performance optimization)
+    if (existingTag.innerHTML !== content) {
+      existingTag.innerHTML = content;
+    }
```
braddialpad: I think this is actually hurting performance rather than helping it. Doing a content comparison is likely a much heavier operation than just directly setting the content.

packages/dialtone-tokens/themes/config.js
```diff
@@ -61,3 +190,341 @@ function _setThemeAttributeOnRoot (theme, brand, rootNode) {
...
+      );
+    } else {
+      console.warn(
+        `[Dialtone] Theme already initialized for this rootNode. ` +
+        `Previous: brand='${existing.brand}', mode='${existing.mode}'. ` +
+        `New: brand='${brandTheme.brand.name}', mode='${mode}'. ` +
+        'Re-initializing with different parameters. Consider using setBrand()/setMode() for dynamic switching.',
+      );
+    }
+  }
```
braddialpad: I don't think it's really that necessary to do this check at all, but if we do want to do it we should output the warning and then immediately `return;` to exit this function. There just isn't really any reason to init more than once.

packages/dialtone-tokens/themes/config.js
```diff
@@ -61,3 +190,341 @@ function _setThemeAttributeOnRoot (theme, brand, rootNode) {
...
+  if (rootNode === document.documentElement && existing) {
+    const hasThemeStyles = document.querySelector('#dialtone-css-core') !== null;
+    if (hasThemeStyles) {
+      console.warn(
+        '[Dialtone] Multiple calls to initDialtoneTheme() detected on document.documentElement. ' +
+        'If you are in an embedded app/micro-frontend, you may need to pass your own container element instead. ' +
+        'Example: initDialtoneTheme(core, brand, mode, myAppContainerElement)',
+      );
+    }
+  }
```
braddialpad: This is the important one. We need to remove the `&& existing` from the condition here. This will prevent an embedded app from checking on the first init (which is exactly what we want).

If hasThemeStyles is true, we want to trigger an error rather than a warning, and then return to exit the function. It's important that we do NOT continue and set a style when this case happens.

packages/dialtone-tokens/themes/config.js
```diff
@@ -61,3 +190,341 @@ function _setThemeAttributeOnRoot (theme, brand, rootNode) {
...
+
+  // Load core tokens (once per rootNode)
+  _setStyleTag('dialtone-css-core', Core.core, rootNode);
+  coreTokensLoadedByRoot.set(rootNode, true);
```
braddialpad: Couldn't this just be a basic boolean, true or false? We never manage more than one root node within a single javascript instance.

braddialpad (changes requested):

packages/dialtone-tokens/themes/config.js
francisrupert: Added validation to `setBrand()` matching `initDialtoneTheme()` pattern. Throws errors for invalid brandTheme. https://github.com/dialpad/dialtone/pull/945/commits/311e8e3589a60fa18b4d78f445658281bf599489

packages/dialtone-tokens/themes/config.js
francisrupert: Removed string comparison optimization. 
 https://github.com/dialpad/dialtone/pull/945/commits/311e8e3589a60fa18b4d78f445658281bf599489

packages/dialtone-tokens/themes/config.js
francisrupert: Added early returns after warnings. That should ensure developers use correct patternof `setBrand()`/`setMode()` ratherthan re-calling `initDialtoneTheme()`. https://github.com/dialpad/dialtone/pull/945/commits/311e8e3589a60fa18b4d78f445658281bf599489

packages/dialtone-tokens/themes/config.js
francisrupert: Updated. Now throws error on FIRST init attempt. Addd a clear error messageto pass container element. Don't want a micro-frontend overwriting parent theme.

packages/dialtone-tokens/themes/config.js
francisrupert: This week I learned about WeakMap. So...

It previously was a boolean.This update addressed https://github.com/dialpad/dialtone/pull/945#discussion_r2492770118, e.g. micro-fronted and shadow dom increasingly at play.

Before I made that change, we had a true/falsae flag for "are core tokens loaded?" However with multiple containers (Shadow DOM, embedded apps, etc), they'd interfere with each other... first one loads, flag flips to true, second one thinks it's done and skips loading.

WeakMap tracks each separately... "has **_this_** specific element been set up?" instead of "has **_any_** element been set up?" Each gets its own independent tracking.

Switching back to a boolean would bring that bug back. Hope I'm making sense.

braddialpad (approved):
Just the small boolean issue remaining. After that I'm happy with it. Pre-approving.

apps/dialtone-documentation/docs/guides/theme-and-mode/index.md
francisrupert: done

packages/dialtone-tokens/themes/config.js
braddialpad: > Before I made that change, we had a true/falsae flag for "are core tokens loaded?" However with multiple containers (Shadow DOM, embedded apps, etc), they'd interfere with each other... first one loads, flag flips to true, second one thinks it's done and skips loading.

What you'd have in the case of a shadow dom app, is a completely separate instance of this config.js, and the variables within it. coreTokensLoadedByRoot would exist once in the dialpad/ubervoice app, another instance of coreTokensLoadedByRoot would exist in the shadowDom app. The coreTokensLoadedByRoot state is not shared between the two.

francisrupert: Ok cool. I'll give it a go since WeakMap appears to be defensive overkill with no scenario it cover. And of course simplify.

francisrupert: Updated in https://github.com/dialpad/dialtone/pull/945/commits/49240312b20b087c6da5b4c341a61d8d6ab329f5

Simplified out from WeakMaps. per-rootNode tracking was solving a problem that won't occur. `coreTokensLoadedByRoot` is back to boolean, and `initializationState` as plain object. Simpler, same functionality.

ninarepetto (approved):

PR #948, opened 2025-10-21
fix(split-button): NO-JIRA fix infinite loop

braddialpad:
# fix(split-button): NO-JIRA fix infinite loop

Obligatory GIF

- [x] Fix

## :book: Description

Fix infinite loop issue by moving the object out of the template into computed property.

## :bulb: Context

This was only happening on vue 2, not vue 3. The way it handles objects in templates is different.

Because it was an object in the template, a new object was being created every time, and then the object way being compared by reference instead of value causing an infinite loop

## :pencil: Checklist

For all PRs:

### DIFF
--- a/packages/dialtone-vue2/components/split_button/split_button-alpha.vue
+++ b/packages/dialtone-vue2/components/split_button/split_button-alpha.vue
@@ -1,6 +1,6 @@
 <template>
   <dt-button
-    v-dt-tooltip="{ message: tooltipText, inverted: kind === 'inverted' }"
+    v-dt-tooltip="tooltipConfig"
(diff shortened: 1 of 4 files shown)


bianca-artola-dialpad:
Screenshot 2025-10-21 at 1 14 33 PM

are you able to open the split button now? still not loading for me but maybe i'm the problem

braddialpad:
> are you able to open the split button now? still not loading for me but maybe i'm the problem

Weird it works fine on local, but not when deployed

bianca-artola-dialpad (approved):
fixed now!

PR #956, opened 2025-10-29
chore: NO-JIRA fix documentation watch command

ninarepetto:
# chore: NO-JIRA fix documentation watch command

Fix to avoid infinite loop when running `nx run dialtone-documentation:start` and making changes in `dialtone-vue3`.

Previously, when making any change in `dialtone-vue3` the build for `dialtone-icons` would enter an infinite loop.
I removed `dialtone-icons` build from the watch command. The caveat of doing this is that when making a change in `dialtone-icons` the web server would have to be restarted to see the changes.

### DIFF
--- a/apps/dialtone-documentation/docs/.vuepress/config.js
+++ b/apps/dialtone-documentation/docs/.vuepress/config.js
@@ -51,6 +51,17 @@ export default defineUserConfig({
       css: {
         devSourcemap: true,
       },
+      server: {
+        watch: {
+          ignored: [
+            '**/packages/dialtone-icons/dist/**',
+            '**/packages/dialtone-vue3/dist/**',
+            '**/packages/dialtone-css/dist/**',
+            '**/packages/dialtone-tokens/dist/**',
+            '**/node_modules/**',
+          ],
+        },
+      },
     },
     vuePluginOptions: {
       template: {
--- a/apps/dialtone-documentation/docs/.vuepress/config.js
+++ b/apps/dialtone-documentation/docs/.vuepress/config.js
@@ -51,17 +51,6 @@ export default defineUserConfig({
       css: {
         devSourcemap: true,
       },
-      server: {
-        watch: {
-          ignored: [
-            '**/packages/dialtone-icons/dist/**',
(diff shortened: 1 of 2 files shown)


apps/dialtone-documentation/docs/.vuepress/config.js
```diff
@@ -51,6 +51,17 @@ export default defineUserConfig({
...
       },
+      server: {
+        watch: {
+          ignored: [
+            '**/packages/dialtone-icons/dist/**',
+            '**/packages/dialtone-vue3/dist/**',
+            '**/packages/dialtone-css/dist/**',
+            '**/packages/dialtone-tokens/dist/**',
+            '**/node_modules/**',
+          ],
```
francisrupert: Boy have I experienced that loop. Thanks.

> ...making a change in dialtone-icons the web server would have to be restarted to see the changes.

Makes sense for icons and even tokens, but would still want to watch for css right?

ninarepetto: Yes, it is rebuilding when making css changes. Here I am only ignoring changes in `dist` folders.

francisrupert: cool then!

apps/dialtone-documentation/project.json
```diff
@@ -40,7 +40,7 @@
         "cwd": "{projectRoot}",
         "commands": [
           {
-            "command": "nx watch --projects=dialtone-css,dialtone-icons,dialtone-tokens,dialtone-vue3 --includeDependentProjects -- nx run \\$NX_PROJECT_NAME:build"
+            "command": "nx watch --projects=dialtone-css,dialtone-tokens,dialtone-vue3 -- nx run \\$NX_PROJECT_NAME:build"
```
iropolo, line 43: seems like this is the key fix, no need to do the ignore part

ninarepetto: I restored the config. It brings some errors on the console, but it works and the page reloads correctly. I think for now this is fine to unblock the development, then we could try and fix those errors.

iropolo (approved):

PR #960, opened 2025-11-02
fix(popover): DP-162957 tippy safe check

iropolo:
# fix(popover): DP-162957 tippy safe check

Jira: https://dialpad.atlassian.net/browse/DP-162957

this.tip can be null in some weird cases due to the component's lifecycle or asynchronous operations.

We already safe check this.tip in some places, I just added some more to prevent crashes.


braddialpad (approved):
Not sure this will actually fix our problems, but it doesn't hurt

iropolo:
> Not sure this will actually fix our problems, but it doesn't hurt

yeah I have doubts too but... we will see if the error continues ... is somewhere else !

PR #964, opened 2025-11-05
fix(popover): DP-162957 reinitialize the tippy when the anchor element changes

noellelevydialpad:
# PR Title
reinitialize the tippy when the anchor element changes

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DP-162957

## :book: Description

Use a mutationObserver to notice when the DOM element rendered into the content slot changes. This element is used to anchor the tippy, and if it has become detached from the DOM, tippy will throw an error when trying to find the element's parent.

As a bonus, this also makes sure to call destroy on this.tip before replacing it, this is possibly fixing a memory leak.

## :bulb: Context

In the native app, we wrap DtRecipeSettingsMenuButton in a DtPopover. When an app update becomes available, the root DOM element rendered by DtRecipeSettingsMenuButton changes, this breaks the tippy binding and clicking the menu just throws an error.

This does not seem to happen in vue 2, I suspect vue 2 is more sloppy about when it reuses DOM elments.

## :pencil: Checklist

For all PRs:

For all Vue changes:

## :crystal_ball: Next Steps

Similar changes should be made in other places we anchor a tippy to slot content, in particular in DtTooltip.

### DIFF
--- a/packages/dialtone-vue2/components/popover/popover.test.js
+++ b/packages/dialtone-vue2/components/popover/popover.test.js
@@ -1,7 +1,6 @@
 import { createLocalVue, mount } from '@vue/test-utils';
 import { DtPopover } from '@/components/popover';
 import SrOnlyCloseButtonComponent from '@/common/sr_only_close_button.vue';
-import { nextTick } from 'vue';
 
 const MOCK_TRANSITION_STUB = () => ({
   render: function () {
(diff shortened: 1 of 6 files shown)


iropolo:
I wouldn't touch the vue2 version if it is working as expected.
If we already verified it works the same Im okay with it.

iropolo (approved):
lgtm.

percy/dialtone failing is expected, approve the change before merge.

iropolo:
It is maybe related to https://dialpad.atlassian.net/browse/DP-164108

noellelevydialpad:
> I wouldn't touch the vue2 version if it is working as expected. If we already verified it works the same Im okay with it.

I actually managed to reproduce the bug in vue2, it just doesn't happen to repro with DtRecipeSettingsMenuButton in vue2 because it sloppily reused the dom node between two different instances of DtButton. If I force it to swap out with a `key` attr, I get the same result as in vue3.

iropolo:
> > I wouldn't touch the vue2 version if it is working as expected. If we already verified it works the same Im okay with it.
> 
> I actually managed to reproduce the bug in vue2, it just doesn't happen to repro with DtRecipeSettingsMenuButton in vue2 because it sloppily reused the dom node between two different instances of DtButton. If I force it to swap out with a `key` attr, I get the same result as in vue3.

ohhh... make sense, yeah a clean up would be okay then.

packages/dialtone-vue2/components/popover/popover.test.js
```diff
@@ -333,4 +334,55 @@ describe('DtPopover Tests', () => {
...
+        propsData: { showAlternateAnchor: false, open: false },
+        stubs: baseStubs,
+        localVue: testContext.localVue,
+        attachTo: document.body,
+      })
+
+      let popoverWindow = wrapper.findComponent({ ref: 'popover' }).findComponent({ ref: 'content' });
+      expect(popoverWindow.isVisible()).toBe(false);
+      wrapper.setProps({ open: true});
+      await nextTick();
```
braddialpad: nit: you can just await setProps instead of doing nextTick

braddialpad (approved):
Thanks for this!

PR #965, opened 2025-11-06
fix(i18n): NO-JIRA initialization tweaks

superfreddialpad:
- refactor DialtoneLocalization to allow i18n usage without needing to interact directly with app
- don't call useI18n outside of component init/setup, use a cached object for `.$t` et al. rather than calling `useI18N` for every access (useI18N should not be called in the render function)

corresponding changes on WCF side: https://github.com/dialpad/goblin-client-tools/pull/54

### DIFF
--- a/packages/dialtone-vue2/localization/index.js
+++ b/packages/dialtone-vue2/localization/index.js
@@ -70,7 +70,6 @@ export class DialtoneLocalization {
       namespaces: [dialtoneNamespace],
     });
 
-    //localeManager.install(dialtoneNamespace);
     this.i18n = localeManager.useI18N(dialtoneNamespace);
 
     DialtoneLocalization.instance = this;
@@ -123,7 +122,6 @@ export class DialtoneLocalization {
    * https://github.com/dialpad/goblin-client-tools/tree/main/packages/i18n#t
    */
   $t (...args) {
-    //return useI18N(dialtoneNamespace).$t(...args);
     return this.i18n.$t(...args);
   }
 
@@ -134,7 +132,6 @@ export class DialtoneLocalization {
    * https://github.com/dialpad/goblin-client-tools/tree/main/packages/i18n#ta
    */
   $ta (...args) {
-    //return useI18N(dialtoneNamespace).$ta(...args);
     return this.i18n.$ta(...args);
   }
 
--- a/packages/dialtone-vue2/localization/index.js
+++ b/packages/dialtone-vue2/localization/index.js
@@ -1,4 +1,4 @@
-import { LocaleManager, RawBundleSource, useI18N } from '@dialpad/i18n-vue2';
+import { LocaleManager, RawBundleSource } from '@dialpad/i18n-vue2';
 
 import enUS from './en-US.ftl?raw';
 import zhCN from './zh-CN.ftl?raw';
@@ -36,6 +36,10 @@ const localeManagerStorageKey = 'user-locale';
  * https://github.com/dialpad/goblin-client-tools/tree/main/packages/i18n#i18n-vue-3-compatible
  */
 export class DialtoneLocalization {
+  /**
+   * @prop {import('@dialpad/i18n').UseI18N} i18n
+   * @private
+   */
   constructor (locale = null) {
     if (typeof DialtoneLocalization.instance === 'object') {
       return DialtoneLocalization.instance;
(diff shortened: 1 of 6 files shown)


dlovero-dialpad:
FYI in case you want to review it @braddialpad: this PR is dependant on this version: https://github.com/dialpad/goblin-client-tools/pull/54/files and it should solve the problem with the js-core repo, where we did the hotfix for the last release

packages/dialtone-vue3/localization/index.js
```diff
@@ -37,12 +37,17 @@ const localeManagerStorageKey = 'user-locale';
...
+   * @private
+   */
   constructor (locale = null) {
     if (typeof DialtoneLocalization.instance === 'object') {
       return DialtoneLocalization.instance;
     }
 
-    const app = getCurrentInstance().appContext.app;
+    // no good
+    //const app = getCurrentInstance().appContext.app;
```
braddialpad: Yeah this is vue3 internal. Happy to get rid of it. You can also remove the import `import { getCurrentInstance } from 'vue';`

packages/dialtone-vue3/localization/index.js
```diff
@@ -60,6 +65,10 @@ export class DialtoneLocalization {
       ]),
     });
 
+    // ehhh im pretty sure we want there to be one of these already?? so rather
+    // than instatiate it directly, there should be a createLocaleManager(),
+    // which uses the global version (while also adding necessary options!) or
+    // creates a new one like this:
```
braddialpad: so are you saying we need to create a new singleton for LocaleManager, like we already have for DialtoneLocalization?

dlovero-dialpad: I think https://github.com/dialpad/dialtone/pull/965/files/1c49ab9cec72339bd8cec79e1e04c3bcdd0f5dad#diff-8923ea3657ff38996b6760514a56c8afe2287e80c58bfa7449c318a256d5eae0R74

It's already taking care of it, so we don't need to do much, iirc the comment is mostly whose responsibility is to set up the locale manager properly

dlovero-dialpad:
@braddialpad I think we have everything merged now. Do we need anything else to have a well-built PR for Dialtone?

packages/dialtone-vue3/localization/index.js
```diff
@@ -120,7 +125,8 @@ export class DialtoneLocalization {
    * https://github.com/dialpad/goblin-client-tools/tree/main/packages/i18n#t
    */
   $t (...args) {
-    return useI18N(dialtoneNamespace).$t(...args);
+    //return useI18N(dialtoneNamespace).$t(...args);
```
braddialpad: We can just remove these entirely instead of commenting

packages/dialtone-vue3/localization/localization.mdx
```diff
@@ -133,6 +133,10 @@ You can pass in a locale to the constructor. The default is your browser locale
 
 `dialtonei18n.currentLocale = 'en-US';`
 
+another todo - interacting with the current locale this way is only a workaround
+for dialtone users that aren't already using @dialpad/i18n; those who are should
+use set it there
```
braddialpad, line 138: Is there a function to set it? other than the constructor?

packages/dialtone-vue2/localization/index.js
```diff
@@ -128,7 +134,8 @@ export class DialtoneLocalization {
    * https://github.com/dialpad/goblin-client-tools/tree/main/packages/i18n#ta
    */
   $ta (...args) {
-    return useI18N(dialtoneNamespace).$ta(...args);
+    //return useI18N(dialtoneNamespace).$ta(...args);
+    return this.i18n.$ta(...args);
```
braddialpad: remove commented code

braddialpad (approved):
Looks good, just some minor cleanup and a question.

francisrupert:
@dlovero-dialpad Is this PR still in the works? Last activity was three weeks ago. One nit for the title:

`fix: NO-JIRA i18n initialization tweaks` → `fix(i18n): NO-JIRA initialization tweaks`

dlovero-dialpad:
@francisrupert It is about to cross the finish line, but we had to switch priorities to fully commit to the Vue 3 migration. After that's complete, I'll get back to it 🙏

packages/dialtone-vue3/localization/localization.mdx
dlovero-dialpad: not a function 'traditionally speaking', but I believe the setter `dialtonei18n.currentLocale = 'es-LA';` internally calls `setI18N()` which does the magic: https://github.com/dialpad/dialtone/blob/b90844e0b1e55c308f887bf81b54872c0318e465/packages/dialtone-vue3/localization/index.js#L154

is this what you expected?

braddialpad: right, okay that's good, just didn't know exactly how it was supposed to be set in dialtonei18n

PR #970, opened 2025-11-14
fix(tab): DLT-2685 tabs update

iropolo:
# fix(tab): DLT-2685 tabs update

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DP-148500

## :book: Description

The main issue is the tabs are not reactive to DOM changes.
There is the `beforeUpdate` hook but it is not in timing to catch the DOM changes.

Here is the hotfix PR. -> https://github.com/dialpad/dialtone/pull/849

The first solution of it was change `beforeUpdate` to `update`, this could generate some infinite loop issue because mutating state.

We just added a safecheck to it and it is okay.

I spend a little bit more time on this ticket and I came with some alternatives.
Maybe it is not necessary to do any change at all and we can continue with this `hotfix` as a `fix`.
I also tried to replicate the issue in beta with both vue2 and vue3 but seems like it is working fine now.

---

First solution is go with the initial idea, change `beforeUpdate` to `updated`.
`tabs` is an array and we are updating the elements inside of it. Since Vue doesn't go deeply we should be okay with it.

Second solution is to go with a nextTick on beforeUpdate. It works... It is not a better solution as using `updated` because its the same.
It is more explicit that we are intentionally waiting for DOM changes. 

Third solution is trigger an update on any tab interaction. These are on:
`tabLeft` - `tabRight` - `selectTab` - `onHomeButton` - `onEndButton`.
With this we will always work with updated `tabs`.

You can switch to the branch and play a little bit, the examples and solution are already in place.

## :bulb: Context

This is a second round of research of the main issue and find (if there is) some alternative solution.

### DIFF
--- a/packages/dialtone-vue2/components/tab/tab_group.vue
+++ b/packages/dialtone-vue2/components/tab/tab_group.vue
@@ -180,7 +180,7 @@ export default {
     this.updateSelected();
   },
 
-  beforeUpdate () {
+  updated () {
     this.updateSelected();
   },
 
@@ -216,6 +216,7 @@ export default {
     },
 
     tabLeft () {
+      // this.tabs = this.getTabChildren();
       const index = this.getFocusedTabIndex();
       if (index === -1) return;
 
@@ -224,6 +225,7 @@ export default {
     },
 
     tabRight () {
+      // this.tabs = this.getTabChildren();
       const index = this.getFocusedTabIndex();
       if (index === -1) return;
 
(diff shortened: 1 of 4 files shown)


braddialpad (approved):
Hey, I think using "updated" is the best solution. It is possible it may run more times than we "need it to" but it is better for future proofing any new changes to this component.

PR #990, opened 2025-12-04
fix(combobox-multi-select): DP-162879 fix event listeners

braddialpad:
# fix(combobox-multi-select): DP-162879 fix event listeners

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DP-162879

## :book: Description

Updated Vue 3 multiselect component to have all "listeners" pass down to the input as it did in Vue 2.

## :bulb: Context

Justin investigated in https://github.com/dialpad/firespotter/pull/67203 and found that listeners were not going to the same element that they were in vue 2. Seems `$listeners` was just removed from the component instead of being migrated to the vue 3 equivalent via `extractVueAttrs()`. Fixed and added the focus event to the storybook so we can confirm it works as expected.

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

update dialtone in vue 3 branch and verify

### DIFF
--- a/packages/dialtone-vue3/recipes/comboboxes/combobox_multi_select/combobox_multi_select.stories.js
+++ b/packages/dialtone-vue3/recipes/comboboxes/combobox_multi_select/combobox_multi_select.stories.js
@@ -24,6 +24,7 @@ export const argsData = {
   onRemove: action('remove'),
   onMaxSelected: action('maxSelected'),
   onComboboxHighlight: action('comboboxHighlight'),
+  onFocus: action('focus'),
 };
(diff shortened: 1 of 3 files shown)


jawrey (approved):

jawrey:
wait, @braddialpad we need `inheritAttrs: false` here right?  and then manually exclude those listeners and bind the resulting object back to the root.  Otherwise I think this is binding event listeners twice, once to the root and once to the input.  which will be bad for events that bubble

PR #1002, opened 2025-12-17
fix(modal): DP-169147 nested modal style

iropolo:
# fix(modal): DP-169147 nested modal style

https://github.com/user-attachments/assets/a06b077f-cb92-49d1-a5fb-0cf42d732406

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DP-169147

## :book: Description

Seems like the removal of `background-clip: padding-box;` bring some unwanted changes.

## :bulb: Context

The issue can be visualized following this instructions:

1. Go to the Office setting
2. Go to Local Presence
3. Click on the purchase bundle
4. Click on the edit

I couldn't reproduce it on that way but you can also see it with the remove custom emoji flow.

1. Go to Preferences
2. Custom Emojis
3. Delete Emoji
(Attached video)

---

This is linked to the recent change. #992 

I see it was also removed on popover but Im not sure if it is needed to be restored too.
@francisrupert please verify this and if restoring on modal and/or popover is the way to go.
Thanks,

### DIFF
--- a/packages/dialtone-css/lib/build/less/components/modal.less
+++ b/packages/dialtone-css/lib/build/less/components/modal.less
@@ -83,7 +83,6 @@
   font-size: var(--dt-font-size-200);
   line-height: var(--dt-font-line-height-400);
   background-color: var(--modal-dialog-color-background);
-  background-clip: padding-box;
   border: var(--dt-size-100) solid var(--modal-dialog-color-border);
   border-radius: var(--dt-size-500);
   box-shadow: var(--modal-dialog-shadow);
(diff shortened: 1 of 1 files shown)


francisrupert:
geez, this is so weird. It only occurs when there's a nested Modal within a Modal (in the rendered html at least). Modals are each supposed to sit equally at the root. For that matter, from a design perspective a Modal sitting on top of a Modal is also bad – but that's a different solve.

I have a CSS update in mind that I'll do on this branch rather than reverting the bg clip property.

francisrupert:
https://github.com/user-attachments/assets/f424d171-e963-4952-9313-8fff6c4b3e9f

francisrupert (approved):

PR #1014, opened 2026-01-08
feat(rich-text-editor): DP-171112 emit event for channel click

braddialpad:
# feat(rich-text-editor): DP-171112 emit event for channel click

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DP-171107

## :book: Description

Very similar to commit https://github.com/dialpad/dialtone/commit/6e2334f440e1be542d60ec7018ad6b346993cc87

Realized we will also have to do the exact same thing for channel

## :bulb: Context

allows the consumer to handle the channel click event within the rich text editor


iropolo (approved):
Looks good.

PR #1017, opened 2026-01-12
fix(rich-text-editor): DLT-2770 spacing issue on copy

iropolo:
# fix(rich-text-editor): DLT-2770 spacing issue on copy

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2770

## :book: Description

This code line in `shouldPreserveLineBreaks` function:

```
if (!this.pasteRichText) {
        return !!textData;
      }
```

Is what forces our message to go trough `insertPlainTextWithHardBreaks` in the `without extensions` story.
Because `pasteRichText` is  `false`.

As I can see this is why produces extra blank lines, we are forcing “plain text” and then translating paragraph separators (\n\n) into two hardBreaks.
Im just combine into one if there is two, if there is one, no problemo!

Some tecnical explanation. I just simplified it (I believe) , seems like previously we had two issues, you can see on the video,
first is trailing newlines: when splitting `"test\n"` by newlines, we get `["test", ""]` and the code was inserting a hard break before the empty string, creating an extra blank line.

secondly is some kind of position desync? not sure but the manual tracking pos have some issue. So I just replaced it with an `content` array to push the lines there.

Before:

https://github.com/user-attachments/assets/2ee23460-a9e9-476a-99fc-484af3c361e8

After:

https://github.com/user-attachments/assets/914f3b3e-cf11-450b-9ec0-b3b991ff36d4

### DIFF
--- a/packages/dialtone-vue3/components/rich_text_editor/rich_text_editor.vue
+++ b/packages/dialtone-vue3/components/rich_text_editor/rich_text_editor.vue
@@ -1047,25 +1047,23 @@ export default {
     },
 
     insertPlainTextWithHardBreaks (view, textData) {
-     // Remove trailing newlines to avoid inserting empty lines at the end
-      const trimmedData = textData.replace(/[\r\n]+$/, '');
+      // If we convert both newlines into hardBreak, we create a blank line.
+      // Collapsing avoids the extra hardBreak while preserving single line breaks.
+      const normalizedData = this.pasteRichText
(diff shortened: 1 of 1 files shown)


braddialpad (approved):
This has been an annoying problem for a while, thanks!

PR #1033, opened 2026-01-22
fix(tokens): NO-JIRA multi-layer box shadows processing

francisrupert:
# Multi-layer box shadows processing

- [x] Fix

## :book: Description

**Basic test to confirm:** tab through page, blue focus state should appear.

Fixes multi-layer shadow composite token generation which broke, presumably after the Node 24 update (commit e8255214f) bumped `@tokens-studio/sd-transforms` and `style-dictionary`. 

The dependency update appears to change the output order of shadow component variables, exposing a bug in the postcss plugin where the shadow layer count was being overwritten rather than tracking the maximum layer index.

This caused multi-layer shadows to generate composite variables referencing non-existent non-numbered component variables. The fix uses `Math.max()` to correctly track the highest layer index regardless of processing order. 

## :pencil: Checklist

For all PRs:

Before
image

image

After
image

image

### DIFF
--- a/packages/dialtone-tokens/postcss/dialtone-tokens.cjs
+++ b/packages/dialtone-tokens/postcss/dialtone-tokens.cjs
@@ -52,8 +52,9 @@ function boxShadows (shadowDeclarations, Declaration) {
     .reduce((shadows, shadow) => {
       const [name, index] = shadow
         .split(shadowSegmentsRegex).slice(1, -1);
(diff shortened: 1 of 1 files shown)


ninarepetto (approved):

braddialpad (approved):
Thanks, a smart solution. Shadows have always been difficult with styledictionary

PR #1039, opened 2026-01-29
fix(tokens): DP-173715 downgrade sd-transforms/colorjs.io

braddialpad:
# fix(tokens): DP-173715 downgrade sd-transforms/colorjs.io

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DP-173715

## :book: Description

Downgraded sd-transforms to 1.2.3 and colorjs.io to 1.4.5

## :bulb: Context

When we did the node 24 update we regenerated the lock file which caused some dependencies to update. This caused sd-transforms and colorjs.io to upgrade to later versions that follow the CSS Color Level 4 spec.

See the above jira ticket. Windows 7 only supports an older version of chrome that does not support the CSS Color Level 4 spec causing the issues shown in the ticket, main issue being using "none" instead of "0" in hsl.

My personal thoughts are that we should not support an operating system that Microsoft and Google Chrome both no longer support, but apparently we have a significant number of users still using Windows 7.

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

CP to beta

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,7 +141,7 @@
     "axe-core": "^4.7.2",
     "axe-playwright": "^2.0.3",
     "c8": "^8.0.0",
-    "colorjs.io": "0.5.2",
+    "colorjs.io": "0.4.5",
     "concurrently": "^8.2.2",
     "del": "^6.1.1",
     "eslint": "^9.33.0",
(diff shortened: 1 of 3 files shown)


iropolo:
> we have a significant number of users still using Windows 7

image

iropolo (approved):

PR #1040, opened 2026-01-30
fix(tokens): NO-JIRA output line-height as unitless instead of percentages

francisrupert:
> [!IMPORTANT]
> Targets `next` 

# Output line-height as unitless instead of percentages

![friday](https://github.com/user-attachments/assets/2c93bc09-d5be-4bf9-bbfc-99c77b01c581)

- [x] Fix

## :book: Description

- Update transform filters to use path-based detection for lineHeight tokens with DTCG-compliant `type: dimension`                        
- Exclude lineHeight dimension tokens from generic size transforms (`pxToDp`, `pxToCGFloat`) so they use the correct lineHeight transforms                                                                                                                                
- Add `dt/lineHeight/percentToDecimal` to CSS transform group                                                                             
- Remove duplicate transform registration  

**Verified outputs after fix:**                                                                                                           
                                                                                                                                          
CSS (unitless decimals):                                                                                                                  
```css                                                                                                                                    
--dt-font-line-height-600: 2; // not 200%
--dt-font-line-height-500: 1.8; // not 180%
--dt-font-line-height-100: 1; // not 100%
```                                                                                                                                       
                                                                                                                                          
Android Compose (.em suffix):                                                                                                             
```kotlin                                                                                                                                 
val dtFontLineHeight100 = 1.em                                                                                                            
val dtFontLineHeight200 = 1.2.em                                                                                                          
val dtFontLineHeight300 = 1.4.em                                                                                                          
```                                                                                                                                       
                                                                                                                                          
iOS Swift (CGFloat with percentage values):                                                                                               
```swift                                                                                                                                  
public static let dtFontLineHeight100 = CGFloat(100.00)                                                                                   
public static let dtFontLineHeight200 = CGFloat(120.00)                                                                                   
public static let dtFontLineHeight300 = CGFloat(140.00)                                                                                   
```  

## :bulb: Context

Line-height values were being output as percentages (`140%`) instead of unitless values (`1.4`), which causes CSS inheritance issues, as realized in Beacon https://github.com/dialpad/beacon-app/pull/399#issuecomment-3825994577
                                                                                                                                          
The root cause was commit `5160cae34` ("update types to match DTCG specification") which changed lineHeight token types from              
`"lineHeights"` to `"dimension"`, breaking the transform filters that relied on type-based detection.
                                                                                                                                          
This fix updates the transforms to use path-based detection in addition to type-based detection, ensuring lineHeight tokens are correctly identified regardless of their DTCG type.  

## :pencil: Checklist

For all PRs:

For all CSS changes:

<img width="1918" height="768" alt="image" src="https://github.com/user-attachments/assets/c2ed7741-acf0-4d90-914f-9fd84be9e4ba" />

### DIFF
--- a/packages/dialtone-tokens/build-sd-transforms-layered.js
+++ b/packages/dialtone-tokens/build-sd-transforms-layered.js
@@ -120,7 +120,7 @@ StyleDictionary.registerAction({
 
 StyleDictionary.registerTransformGroup({
   name: 'custom/css/tokens-studio',
-  transforms: [...getTransforms({ platform: 'css' }), 'name/kebab', 'dt/size/pxToRem', 'dt/space/pxToRem']
+  transforms: [...getTransforms({ platform: 'css' }), 'name/kebab', 'dt/size/pxToRem', 'dt/space/pxToRem', 'dt/lineHeight/percentToDecimal']
     .filter(transform => !['name/camel', 'ts/size/px', 'ts/typography/css/fontFamily'].includes(transform)),
 });
 
--- a/packages/dialtone-tokens/build-sd-transforms.js
+++ b/packages/dialtone-tokens/build-sd-transforms.js
@@ -28,7 +28,7 @@ StyleDictionary.registerAction({
 
 StyleDictionary.registerTransformGroup({
   name: 'custom/css/tokens-studio',
-  transforms: [...getTransforms({ platform: 'css' }), 'name/kebab', 'dt/size/pxToRem', 'dt/space/pxToRem'].filter(transform => !['name/camel', 'ts/size/px', 'ts/typography/css/fontFamily'].includes(transform)),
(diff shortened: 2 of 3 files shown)


jeverhart-dialpad:
👍 Docs site preview has the correct values...

Screenshot 2026-02-02 at 10 00 31 AM
<br/><br/>
👍 Prototype also has the correct values when installing locally...

Screenshot 2026-02-02 at 10 06 19 AM
<br/><br/>

One note: the `1.4` value was correct immediately for me when installing in a new, blank prototype. However, I had to add a local workaround for some reason in an older prototype because it was pulling the github packages token version for the layered theming. I used `pnpm pack` for both. I would imagine that will be resolved once this version is published.

braddialpad (approved):

PR #1042, opened 2026-02-02
chore(lint): NO-JIRA update recommend-typography-style to only flag multiple 

francisrupert:
# Update recommend-typography-style to only flag multiple 

lint

- [x] Feature

## :book: Description

Updates the `recommend-typography-style` ESLint rule to only report violations when **two or more different typography categories** are combined in the same class attribute. Previously, the rule flagged **any** single typography utility (like `d-fw-bold` alone).

Changes:                                                                                                                 

- Rule now only reports when 2+ different typography categories are present in the same class (e.g `d-fw*` + d-fs-*), but won't flag single typography utilities (e.g `d-fw-bold`)
- Rule documentation with clearer examples      
- Updated tests

## :bulb: Context

https://dialpad.com/app/messages/agxzfnViZXItdm9pY2VyGAsSC1RleHRNZXNzYWdlGIDAsOq8wN0JDA

Single typography utilities are often used intentionally for overrides (e.g., making text bold without changing other typography properties). The previous behavior flagged these valid use cases, making the rule too **noisy** to be useful.

This change makes the rule smarter by only flagging combinations instead of using the composed `d-headline--*`, `d-body--*`, or `d-label--*` utilities which already combine typography style properties.

## :pencil: Checklist

For all PRs:

### DIFF
--- a/packages/eslint-plugin-dialtone/docs/rules/recommend-typography-style.md
+++ b/packages/eslint-plugin-dialtone/docs/rules/recommend-typography-style.md
@@ -4,9 +4,13 @@ Before combining multiple typography utilities from different categories, first 
 
 ## Rule Details
 
-This rule flags when **two or more different categories** of typography utilities are combined in the same class attribute. Single typography utilities are acceptable for overrides.
+This rule flags two issues:
+
+1. **Multiple categories**: When two or more different categories of typography utilities are combined in the same class attribute. Single typography utilities are acceptable for overrides.
+2. **Conflicting utilities**: When multiple utilities from the same category are used together (e.g., `d-fw-bold d-fw-medium`), which likely indicates a mistake since only one will be applied.
 
 The typography categories are:
+
 - **font-weight**: `d-fw-normal`, `d-fw-medium`, `d-fw-semibold`, `d-fw-bold`
 - **font-size**: `d-fs-*` (e.g., `d-fs-200`, `d-fs-300`)
 - **line-height**: `d-lh-*` (e.g., `d-lh-300`, `d-lh-400`)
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-dialtone/docs/rules/recommend-typography-style.md
```diff
@@ -1,21 +1,44 @@
...
 
 ```html
+<!-- Composed typography utility -->
 <span class="d-headline--md">
+
+<!-- Single typography utility (acceptable for overrides) -->
+<span class="d-fw-bold">
+
+<!-- Single category (same category multiple times is fine) -->
+<span class="d-fw-bold d-fw-medium">
```
ninarepetto: I find this example a bit confusing, I don't know if we want to include it in the docs, even if we allow it

ninarepetto (approved):
Looks really good!

packages/eslint-plugin-dialtone/docs/rules/recommend-typography-style.md
braddialpad: This doesn't make any sense to me

francisrupert: That was just saying that it wouldn't flag for multiple of same type (e.g. font-weight), but that is confusing. I updated to also flag duplicates of the same type (e.g. bold and medium). That was done in [f64f197](https://github.com/dialpad/dialtone/pull/1042/commits/f64f197bde0f8e23e51d5e8f0824c83053cc8218)

braddialpad (approved):
Thanks!

PR #1047, opened 2026-02-09
feat(avatar): DLT-2942 updated avatar component

hynes-dialpad:

https://github.com/user-attachments/assets/db64675a-89c2-4ab2-8805-36c892ba2926

- [x] Feature
- [x] Refactor

## :book: Jira Ticket
[DLT-2942](https://dialpad.atlassian.net/browse/DLT-2942)

## :book: Description

Overhaul of the avatar component: new 11-size numeric scale, dynamic theme-aware OKLCH colors, squircle presence masks, and several new props.

### Sizing

Replaces t-shirt sizes (`xs`--`xl`) with an 11-step numeric scale: `100`, `150`, `200`, `250`, `300` (default), `400`, `500`, `600`, `700`, `800`, `900`. T-shirt sizes remain as deprecated aliases. Sizes 100--500 support presence indicators; 600--900 do not.

Group avatars cap at 2 digits (9+) for sizes 100--250 and 3 digits (99+) for sizes 300--500. The group count badge now uses `role="img"` with an `aria-label` for screen reader context.

### Color system

Replaces 120 hardcoded `d-avatar--color-XXX` classes with a runtime OKLCH color system driven by `data-avatar-family` (1--12) and `data-avatar-variant` (0--9) attributes. Colors are computed via `oklch()` + `calc()` in CSS, rotating around a per-theme anchor hue extracted from `shell.base.color.accent`. Hex fallback is computed in JS only for browsers without `oklch()` support.

**Props API:** `seed` (recommended -- pass a user ID for deterministic colors), `family`/`variant` (advanced -- explicit control), `color` (legacy -- backward compatible).

### Icon sizing

Removes the JS-based `iconSize` slot prop. Icon sizing is now handled entirely in CSS via `--avatar-icon-size: calc(var(--avatar-size-shape) * 0.75)`.

### New props

- **`iconOnly`** -- Transparent background for non-user icons (channels, navigation).
- **`deactivated`** -- Desaturated/washed-out state for inactive users. Uses `filter: saturate(0)` with a `mix-blend-mode: lighten` overlay at 40% opacity.

### CSS

- Squircle `mask-image` cutouts for presence and group badges at each size
- OKLCH color computation via CSS custom properties and `data-*` attribute selectors
- Deactivated state via `::after` pseudo-element overlay
- Removed all hardcoded color classes (~120 selectors)

### Fixes

- Removed nested `role="button"` on icon div when avatar is clickable (parent `<button>` suffices)
- Presence no longer renders for sizes that don't support it (600--900)
- `colorToFamilyVariant()` rejects fractional variant values
- Empty string `seed` now produces deterministic output instead of falling through to `Math.random()`

## :bulb: Context

The previous color system used 120 hardcoded hex classes that couldn't adapt to themes. The new OKLCH system generates 120 color combinations (12 families x 10 variants) from a single theme anchor hue, so avatars automatically shift palette when the theme changes.

The 11-size system replaces t-shirt sizes to allow finer granularity and support larger display sizes (up to 256px).

## :pencil: Checklist

For all PRs:

For all Vue changes:

For all CSS changes:

## :crystal_ball: Next Steps

- Migrate existing `color` prop usages to `seed` across consuming apps
- Remove deprecated t-shirt size aliases in next major version
- Verify hex fallback in browsers without `oklch()` support

## :link: Sources

- [Figma: Avatar component](https://www.figma.com/design/W58r5BkO8qTw3vem9YieJd/DT9-Component-Library--Rebrand-2025-?node-id=5413-13176)

[DLT-2942]: https://dialpad.atlassian.net/browse/DLT-2942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dialtone-css/lib/build/less/components/avatar.less
```diff
@@ -20,40 +20,183 @@
   //  Component CSS Vars
   //  ------------------------------------------------------------------------
   --avatar-color-background: var(--dt-avatar-color-background-000);
-  --avatar-color-text: var(--dt-avatar-color-foreground);
-  --avatar-size-shape: var(--dt-size-600);
+  --avatar-color-text: var(--dt-color-neutral-black);
+  --avatar-size-shape: var(--dt-spacing-400);
```
francisrupert, line 24: Now that you've merged the `layout`/`spacing` system, can update this file. Let me know if you'd like me to take that on for this branch.

packages/dialtone-tokens/tokens/components/avatar/default.json
```diff
@@ -1,88 +1,72 @@
 {
   "avatar": {
+    "anchor": {
```
francisrupert, line 3: At first Tokens Studio was struggled to work with this, but appears fine now. 

Figma in practice won't be able to use it unless we hook up something custom to generate pre-resolved colors. 

Easy path would be manually created prefab Design Tokens strictly for Figma, and excluded from generating CSS Variables. Overengineered path would be using all this to create Figma Styles via Figma API. Chalk that up among other things to reconcile when/if we want to represent these in Figma. 

Image

apps/dialtone-documentation/docs/components/avatar.md
francisrupert, line 1: `avatar.svg` broken because it no longer has access to `var(--dt-avatar-color-background-1700)`

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 264 107" fill="none">
...
</svg>
```

These thumbnails probably all should be redone as part of that ultimate doc site redesign we'll get to.

Image

packages/dialtone-vue/components/avatar/avatar.vue
```diff
@@ -300,6 +346,48 @@ export default {
       return 'dt-avatar-icon';
     },
 
+    /**
+     * Compute family from props, color prop, or random
+     */
+    computedFamily () {
+      // Icon-type avatars don't use color
+      if (this.isIconType()) return undefined;
```
francisrupert: **BUG** 

`isIconType()` is a method, not a computed. Calling a method inside a computed property means Vue cannot track its reactive dependencies. `this.$slots.icon` changes won't trigger recomputation. This works on initial render but will fail if the icon slot is dynamically toggled. 

**RECOMMENDATION**

Convert `isIconType()` to a computed property, or inline `hasSlotContent(this.$slots.icon)` directly.

packages/dialtone-vue/components/avatar/avatar.vue
```diff
@@ -308,15 +396,31 @@ export default {
...
+          'd-avatar--deactivated': this.deactivated,
         },
       ];
     },
 
+    /**
+     * Compute inline styles for fallback color in browsers that don't support oklch()
+     */
+    avatarStyles () {
+      const styles = { ...this.$attrs.style };
```
francisrupert: Add a guard
```suggestion
      const styles = typeof this.$attrs.style === 'object' ? {{ ...this.$attrs.style } : {};
```

packages/dialtone-vue/components/avatar/avatar.test.js
francisrupert, line 1: Missing test coverage for new props, according to Claude

- `family` / `variant` props: No unit tests verify that passing explicit `family/variant` produces the correct `data-avatar-family/data-avatar-variant` attributes.
- `iconOnly` prop: No tests verify the class is applied or that `data-avatar-family/data-avatar-variant` are omitted.
- Group digit capping by size: No tests verify that small sizes cap at `9+` instead of `99+`.
- `color` prop migration: No tests verify `colorToFamilyVariant` produces correct mappings.

packages/dialtone-vue/components/avatar/avatar.vue
```diff
@@ -3,7 +3,9 @@
     :is="clickable ? 'button' : 'div'"
     :id="id"
     :class="avatarClasses"
-    :style="$attrs.style"
+    :style="avatarStyles"
+    :data-avatar-family="!iconOnly ? computedFamily : undefined"
+    :data-avatar-variant="!iconOnly ? computedVariant : undefined"
     data-qa="dt-avatar"
```
francisrupert, line 9:
```suggestion
    data-qa="dt-avatar"
    :type="clickable ? 'button' : undefined"
```

packages/dialtone-css/lib/build/less/components/avatar.less
```diff
@@ -20,40 +20,183 @@
...
+  //  ------------------------------------------------------------------------
+  &--deactivated {
+    .d-avatar__canvas {
+      position: relative;
+      filter: saturate(0);
+
+      &::after {
+        position: absolute;
+        background-color: var(--dt-color-neutral-white);
+        border-radius: inherit;
```
francisrupert: Image

Only noticeable in dark mode.

Can't inherit radius
```suggestion
        border-radius: var(--avatar-border-radius);
```

francisrupert (approved):
I'll approve as it is from design perspective, though with comments. We'll need to be careful is everything in `dialtone-vue/components/avatar/`. This is as breaking as it gets, and definitely need review from Dialtone engineer, only @braddialpad is available Monday/Tuesday (Feb 16/17).

packages/dialtone-vue/components/avatar/avatar.vue
francisrupert: I committed that change

packages/dialtone-tokens/tokens/components/avatar/default.json
hynes-dialpad: I guess the question becomes how much do we need Figma to reflect. We don't need Figma avatars to auto-generate (like in Vue component). We can pre-select 2 dozen backgrounds in there. 

In the end, I'm not worried about Figma components.

packages/dialtone-vue/components/avatar/avatar.vue
francisrupert: I committed that change

packages/dialtone-css/lib/build/less/components/avatar.less
francisrupert: I committed that change

apps/dialtone-documentation/docs/components/avatar.md
francisrupert: I embedded the colors and committed. Still a circle, but punting.

packages/dialtone-tokens/tokens/components/avatar/default.json
francisrupert: yep, that's the easy path.

francisrupert:
Changed label to `no-visual-test` so it doesn't run Percy every single commit. That eats up monthly allotment.

hynes-dialpad:
Thanks @francisrupert. I had Claude address your comments. Also I ran the PR through Codex and addressed a few issues it highlighted as well. Let's merge for now and have the engineers check the `next` branch. I appreciate the help.

packages/dialtone-vue/components/avatar/avatar.vue
```diff
@@ -113,31 +121,61 @@ export default {
...
       default: undefined,
     },
 
     /**
-     * The size of the avatar
-     * @values xs, sm, md, lg, xl
+     * The size of the avatar.
+     * T-shirt sizes (xs, sm, md, lg, xl) are deprecated and will be removed in the next major version.
+     * Please use the numeric scale instead.
+     * @values 100, 150, 200, 250, 300, 400, 500, 600, 700, 800, 900, xs, sm, md, lg, xl
```
braddialpad, line 173: Do we have a guide to translate the old t-shirt sizes to the equivalent (or closest) numerical value?

packages/dialtone-vue/components/avatar/avatar.vue
```diff
@@ -78,20 +81,25 @@
...
+  AVATAR_VARIANT_COUNT,
+  colorToFamilyVariant,
+  getRandomFamilyVariant,
+  computeAvatarHex,
 } from './avatar_constants';
-import { ICON_SIZE_MODIFIERS } from '@/components/icon/icon_constants.js';
 import { extractInitialsFromName } from './utils';
 
+// Check if browser supports oklch() - only compute hex fallback if not
+const supportsOklch = typeof CSS !== 'undefined' && CSS.supports?.('background', 'oklch(0.5 0.1 0)');
```
braddialpad, line 101: If I'm not mistaken we are going to use oklch within all our tokens, therefore browsers would be required to support them if they want to use dialtone correct @francisrupert ?

If the case we can remove this check.

packages/dialtone-vue/components/avatar/avatar.vue
```diff
@@ -63,9 +64,11 @@
       v-if="showGroup"
       class="d-avatar__count"
       data-qa="dt-avatar-count"
+      role="img"
+      :aria-label="`${group} participants`"
```
braddialpad, line 68: Needs to be i18n translated.

packages/dialtone-vue/components/avatar/avatar.vue
francisrupert: Agree, can remove that. I have a back-pocket option to backwards-support. Generally...

Each base color set with HEX and then overridden with `@supports`, e.g...
```css
:root {
  ...
  --dt-color-red-100: #e53935;
  ...
}

@supports (color: oklch(0 0 0)) {
  :root {
    ...
    --dt-color-red-600: oklch(0.5892 0.2142 28.35);
    ...
  }
}
```
...and each reference resolves to whichever one. But that doubles the base colors, adds PostCSS complexity.

Officially, our Chrome support is officially "latest" and official support for Microsoft Edge is well beyond oklch support.

PR #1061, opened 2026-02-16
refactor(style): DLT-3014 wrap all styles in css cascade layers

francisrupert:
# refactor(style): wrap all Dialtone styles in CSS Cascade Layers

> [!NOTE]
> Targets `next` 

image

- [x] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-3014

## :book: Description

All Dialtone CSS is now organized into four cascade layers:

```css
@layer dialtone.reset, dialtone.base, dialtone.components, dialtone.utilities;
```

image

This gives utilities guaranteed priority over components via layer ordering instead of relying solely on source order and specificity. No breaking changes for consumers — existing usage works identically.

### What was done

- Wrapped all ~80 Less source files (components, recipes, utilities, reset, base, themes) in the appropriate `@layer` block
- Updated `dialtone-generators.cjs` to output generated utilities inside `@layer dialtone.utilities`, with a fix to reset state between invocations (watch mode)
- Updated `postcss-responsive-variations` plugin to accept a `layer` option so responsive `@media` utilities also land inside the correct layer
- Added a new `validate-layers.cjs` PostCSS plugin that runs during build and fails if any `.d-` or `.dt-` selector is found outside a valid layer
- Added unit tests for the generators plugin
- Made the `DtText` component globally available via the Vue plugin export
- Fixed the `DialtoneUsage` component to conditionally render do/dont sections when slots are empty

### Documentation

- New guide page: **CSS Cascade Layers** (`/guides/css-layers/`) covering both consumer and contributor usage
- New blog post: **Dialtone now uses CSS Cascade Layers** (`/whats-new/`)
- Updated the Contributing guide with a CSS Architecture section linking to the layers guide
- Added "CSS Cascade Layers" to the site nav under Guides > Getting Started
- Fixed prop warnings on the Text component documentation page (`maxLines` type, missing `kind`, missing `as`)

## :bulb: Context

Dialtone previously relied on source order and `!important` to ensure utilities override components. This worked but was fragile — adding a new component file in the wrong order or having third-party CSS load after utilities could break overrides. CSS Cascade Layers make the override hierarchy explicit and deterministic regardless of source order or specificity.

## :pencil: Checklist

For all PRs:

For all CSS changes:

## :link: Sources

- [MDN: CSS Cascade Layers](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@layer)
- [CSS Cascade Layers Explainer](https://css.oddbird.net/layers/explainer/)


braddialpad (approved):
This is great! love it

PR #1083, opened 2026-02-26
fix(scroller): NO-JIRA fix dynamic text

iropolo:
Fix the dynamic text on scroller component to avoid Percy false positive.

Before:

https://github.com/user-attachments/assets/12fabdf3-3be7-4afe-9862-c0a499ecfce5

After: 

https://github.com/user-attachments/assets/599b3e76-4234-4f47-a78a-667e21bc5577


iropolo:
The Percy error on the dynamic scroll text is expected and it should be the definitive.
Should be always the same from now on.

braddialpad (approved):

PR #1100, opened 2026-03-04
fix(tooltip): NO-JIRA add Shadow DOM support for DtTooltipDirective

patosabogal:
## Summary
- `DtTooltipDirective` and `DtTooltip`'s `externalAnchor` computed property relied on `document.body.querySelector()` which cannot pierce Shadow DOM boundaries. This caused tippy.js to receive `null` as its target when used inside custom elements (`defineCustomElement`), resulting in the error: `tippy() was passed null as its targets (first) argument`.
- Uses `getRootNode()` instead of `document.body` for `externalAnchor` queries in `tooltip.vue`, so lookups resolve within the shadow root when inside a custom element, and fall back to `document` in regular DOM.
- Mounts the directive's tooltip app inside the element's root node instead of `document.body`, creating per-root-node instances so each Shadow DOM gets its own tooltip container.

## Context
https://dialpad.com/app/messages/agxzfnViZXItdm9pY2VyGAsSC1RleHRNZXNzYWdlGIDAqKO36MALDA

## Test plan
- [ ] Verify tooltips still work in regular Vue apps (no Shadow DOM)
- [ ] Verify `v-dt-tooltip` directive works inside `defineCustomElement` / Shadow DOM
- [ ] Verify `DtTooltip` with `externalAnchor` prop works inside Shadow DOM
- [ ] Run existing tooltip tests

### DIFF
--- a/packages/dialtone-vue/components/tooltip/tooltip.vue
+++ b/packages/dialtone-vue/components/tooltip/tooltip.vue
@@ -315,7 +315,9 @@ export default {
     },
 
     anchor () {
-      return this.externalAnchor ? document.body.querySelector(this.externalAnchor) : getAnchor(this.$refs.anchor);
+      return this.externalAnchor
+        ? (this.$el.getRootNode() || document.body).querySelector(this.externalAnchor)
+        : getAnchor(this.$refs.anchor);
     },
   },
 
--- a/packages/dialtone-vue/components/tooltip/tooltip.vue
+++ b/packages/dialtone-vue/components/tooltip/tooltip.vue
@@ -4,7 +4,7 @@
          elements within the span rather than on the span itself -->
     <!-- eslint-disable-next-line vuejs-accessibility/no-static-element-interactions -->
     <span
-      v-if="!externalAnchor"
+      v-if="!externalAnchor && !externalAnchorElement"
       ref="anchor"
       data-qa="dt-tooltip-anchor"
       @focusin="onEnterAnchor"
@@ -247,9 +247,19 @@ export default {
     /**
      * External anchor id to use in those cases the anchor can't be provided via the slot.
      * For instance, using the combobox's input as the anchor for the popover.
+     * @deprecated Use externalAnchorElement instead for Shadow DOM compatibility.
      */
     externalAnchor: {
       type: String,
+      default: null,
+    },
+
+    /**
+     * External anchor element reference. Use this instead of externalAnchor when
+     * the anchor may be inside a Shadow DOM, as querySelector cannot pierce shadow boundaries.
+     */
+    externalAnchorElement: {
+      type: HTMLElement,
       default: null,
     },
   },
@@ -315,9 +325,8 @@ export default {
     },
 
     anchor () {
-      return this.externalAnchor
-        ? (this.$el.getRootNode() || document.body).querySelector(this.externalAnchor)
-        : getAnchor(this.$refs.anchor);
+      if (this.externalAnchorElement) return this.externalAnchorElement;
+      return this.externalAnchor ? document.body.querySelector(this.externalAnchor) : getAnchor(this.$refs.anchor);
     },
   },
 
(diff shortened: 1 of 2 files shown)


maysamkf (approved):
LGTM

braddialpad:
Adding @noellelevydialpad since this was originally your change

noellelevydialpad:
Why do we need to render the tooltips in the same shadow dom as the anchor? why not keep it a singleton?

braddialpad:
> Why do we need to render the tooltips in the same shadow dom as the anchor? why not keep it a singleton?

I think this is probably the way to go. This change feels overly complex

packages/dialtone-vue/components/tooltip/tooltip.vue
```diff
@@ -315,7 +315,9 @@ export default {
     },
 
     anchor () {
-      return this.externalAnchor ? document.body.querySelector(this.externalAnchor) : getAnchor(this.$refs.anchor);
+      return this.externalAnchor
+        ? (this.$el.getRootNode() || document.body).querySelector(this.externalAnchor)
+        : getAnchor(this.$refs.anchor);
```
braddialpad: I think what we need to to do here is stop the use of querySelector as it doesn't penetrate shadowDom. Instead let's create a new prop called `externalAnchorElement` and instead of being a selector this would be a reference to an actual element. This way we don't have to use querySelector. We'll have to keep the externalAnchor prop for now so we don't cause a breaking change, but we can get rid of it in the next major release of dialtone.

Lets see if we can make this work with no changes to the prior version of tooltip.js, other than making it use this new prop. I think it is still okay if the tooltip renders at the document.body rather than the shadowRoot body.

braddialpad (changes requested):

packages/dialtone-vue/directives/tooltip_directive/tooltip.js
```diff
@@ -125,6 +125,7 @@ export const DtTooltipDirective = {
         }
       });
 
+      tooltipConfig.anchorElement = anchor;
       anchor.setAttribute('data-dt-tooltip-id', tooltipId);
```
noellelevydialpad, line 129: I believe this is now unused.

braddialpad (approved):
This looks good, thanks!

braddialpad:
Have we tested to see if this still solves the original issue?

packages/dialtone-vue/directives/tooltip_directive/tooltip.js
```diff
@@ -17,8 +17,8 @@ export const DtTooltipDirective = {
           };
         },
 
-        created () {
-          globalThis.__DtTooltipDirectiveAppInstance = getCurrentInstance();
+        mounted () {
+          globalThis.__DtTooltipDirectiveApp = this;
```
patosabogal, line 21: Previously this used `getCurrentInstance()` which returns Vue's **internal** instance object. Calling methods through `instance.ctx` (e.g. `instance.ctx.addOrUpdateTooltip(...)`) does modify the reactive `tooltips` array, but it does **not** trigger the singleton app's render function to re-run — meaning the DtTooltip components were never actually created.

Storing `this` (the component proxy) instead ensures that method calls go through Vue's public reactive proxy. When the proxy's methods mutate `this.tooltips`, Vue's reactivity system properly tracks the dependency and queues a re-render of the singleton app, which then creates the DtTooltip instances for each registered tooltip.

This was verified by testing in digital-channels inside a Shadow DOM custom element — the directive tooltips only started rendering after this change.

packages/dialtone-vue/directives/tooltip_directive/tooltip.js
patosabogal: It's no longer used as a CSS selector for `querySelector`, but the attribute is still used for internal bookkeeping:

- **Line 79** (`unmounted`): reads the ID to call `removeTooltip` when the element is destroyed
- **Lines 85/91** (`setupTooltip`): reads the ID to check if this anchor already has a tooltip (update vs create) and to reuse the same ID

Removing it would break cleanup on unmount and tooltip updates.

packages/dialtone-vue/directives/tooltip_directive/tooltip.js
braddialpad: sounds good 👍

PR #1102, opened 2026-03-04
feat(loader,progress-circle): NO-JIRA polish DtProgressCircle visual and align DtLoader visual with DtProgressCircle

francisrupert:
image

## :hammer_and_wrench: Type Of Change

- [x] Feature

## :book: Jira Ticket

NO-JIRA

## :book: Description

- Added stroke-linecap="round" to DtProgressCircle fill shape
- Polished DtProgressCircle stroke rendering with rounded linecaps and minimum visual thresholds
- Replaced DtIconLoading icon import in DtLoader with an inline SVG spinner arc that visually aligns with DtProgressCircle
- Added a comet-fade conic-gradient mask to the loader icon so the leading edge fades to transparent

## :bulb: Context

After landing the DtProgressCircle component, we aligned DtLoader (the indeterminate spinner) to use a matching SVG arc so the two components share a consistent visual language. The loader also gains a comet-tail fade effect via a CSS mask for a more polished spin animation.

## :package: Cross-Package Impact

| Package | Changes | Downstream Impact |
|---|---|---|
| dialtone-css | Updated loader.less, progress_circle.less | Vue components depend on these styles |
| dialtone-vue | Updated DtLoader SVG, polished DtProgressCircle | --- |

## :pencil: Checklist

For all PRs:

- [x] I have ensured no private Dialpad links or info are in the code or pull request description (Dialtone is a public repo!).
- [x] I have reviewed my changes.
- [x] I have added all relevant documentation.
- [x] I have considered the performance impact of my change.

For all Vue changes:

- [x] I have added / updated unit tests.
- [x] I have validated components with a screen reader.
- [x] I have validated components keyboard navigation.

For all CSS changes:

- [x] I have used design tokens whenever possible.
- [x] I have considered how this change will behave on different screen sizes.
- [x] I have visually validated my change in light and dark mode.
- [x] I have used gap or flexbox properties for layout instead of margin whenever possible.

## :crystal_ball: Next Steps

- Validate comet-fade visually in light and dark mode across all loader sizes

https://github.com/user-attachments/assets/8d517828-1e10-484c-a5fc-715ed65b2e07

https://github.com/user-attachments/assets/0bd50a1b-156a-4b47-a312-83d5c47a525b

PR #1098, opened 2026-03-03
feat(progress-circle): DLT-2983 add progress-circle component

braddialpad:
## Obligatory GIF (super important!)

Obligatory GIF

## :hammer_and_wrench: Type Of Change

These types will increment the version number on release:

- [x] Feature

These types will not increment the version number, but will still deploy to documentation site on release:

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Style (code style changes, not css changes)
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket

[DLT-2983](https://dialpad.atlassian.net/browse/DLT-2983)

## :book: Description

Adds a new `DtProgressCircle` component — a circular SVG progress indicator for determinate upload or processing progress. Named "Progress Circle" (following Adobe Spectrum's convention) in anticipation of a future linear/bar variant.

Key changes:
- New `DtProgressCircle` component with `ariaLabel`, `progress` (0–100), `size` (100–800, matching icon scale), and `kind` (color variant) props
- Seven `kind` variants: `default`, `brand`, `critical`, `positive`, `warning`, `info`, `ai`
- The `ai` kind renders an SVG `<linearGradient>` in `<defs>` since SVG strokes don't support CSS gradients — uses the same color stops as `--dt-color-gradient-gold-red-magenta-purple` converted to oklch
- New CSS component `d-progress-circle` with BEM modifiers for size and kind
- Old `DtProgress` component removed; `image_carousel.vue` updated to import `DtProgressCircle` from `@/components/progress_circle`
- Exported from the `dialtone-vue` package barrel and registered in `common/components_list.js`
- Full documentation page with interactive demo, variant examples, class table, and Vue API docs

## :package: Cross-Package Impact

| Package | Changes | Downstream Impact |
|---|---|---|
| `dialtone-css` | New `components/progress_circle.less`; old `progress.less` removed | `dialtone-vue` depends on these styles |
| `dialtone-vue` | New `DtProgressCircle` component + constants; `image_carousel.vue` consumer updated | Documentation site needs the built dist |
| `dialtone-documentation` | New `/components/progress-circle.html` page; sidebar nav entry; class table JSON | — |

## :page_facing_up: Documentation Artifacts

| Artifact | Status |
|---|---|
| Vue source (`components/progress_circle/progress_circle.vue`) | ✅ Done |
| Tests (`progress_circle.test.js`, 16 passing) | ✅ Done |
[...]

PR #1111, opened 2026-03-08
feat(button): DLT-3086 update loading state to use dt-loader component

francisrupert:
Obligatory GIF

- [x] Feature

## :book: Jira Ticket

[DLT-3086](https://dialpad.atlassian.net/browse/DLT-3086)

## :book: Description

Follow up to https://github.com/dialpad/dialtone/pull/1098 and https://github.com/dialpad/dialtone/pull/1102

- DtButton's loading state now renders a `<DtLoader>` component instead of relying solely on a CSS `::before` pseudo-element spinner.
- The CSS `::before` fallback is preserved for backwards compatibility with raw HTML consumers. A `:has(> .d-btn__loader)` rule suppresses it when the Vue loader is present.
- Loader size scales proportionally to button size via existing `BUTTON_ICON_SIZES` mapping.

## :pencil: Checklist

For all PRs:

For all Vue changes:

For all CSS changes:

https://github.com/user-attachments/assets/b6f127db-0af1-4c4b-b3f0-46024c9aceb9

[DLT-3086]: https://dialpad.atlassian.net/browse/DLT-3086?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dialtone-css/lib/build/less/components/button.less
+++ b/packages/dialtone-css/lib/build/less/components/button.less
@@ -561,7 +561,38 @@
 //  $$  LOADING STATE
 //  ----------------------------------------------------------------------------
 .d-btn--loading {
-  .d-btn-loading();
+  //  Disable clicking while the loader is in place
+  pointer-events: none;
+
+  //  Hide anything that was there before
+  .d-btn__icon,
+  .d-btn__label {
+    opacity: 0;
+    transition: opacity 50ms var(--ttf-in-out);
+  }
+
+  //  Pseudo-element spinner (fallback for raw HTML consumers)
+  &::before {
+    position: absolute;
+    width: var(--dt-size-500);
+    height: var(--dt-size-500);
+    border: var(--dt-size-200) solid currentColor;
+    border-left-color: transparent !important;
+    border-radius: var(--dt-size-radius-circle);
+    animation: d-loading-circle 900ms infinite linear;
+    content: "";
+  }
+
+  // When a real DtLoader element is present, suppress the pseudo-element.
+  // Back-compat with old class name.
+  &:has(> .d-btn__loader)::before {
+    content: none;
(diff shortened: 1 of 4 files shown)


packages/dialtone-css/lib/build/less/components/button.less
```diff
@@ -561,7 +561,38 @@
 //  $$  LOADING STATE
 //  ----------------------------------------------------------------------------
 .d-btn--loading {
-  .d-btn-loading();
```
francisrupert, line 564: Zero reason to be a LESS mixin. Internally it's only used for this one purpose, and I see zero use at least in `firespotter`. 

I suspect it might have been seen as a reusable less mixin pre-Vue by Josh, but even then I doubt it was actually used that way.

packages/dialtone-css/lib/build/less/components/button.less
```diff
@@ -561,7 +561,38 @@
...
+  }
+
+  // When a real DtLoader element is present, suppress the pseudo-element.
+  // Back-compat with old class name.
+  &:has(> .d-btn__loader)::before {
+    content: none;
+  }
+
+  //  Position the DtLoader element within the button
+  > :where(.d-btn__loader) {
```
francisrupert, line 593: `0,0,0` specificity. Getting some good use out of this lately

Image

francisrupert (commented):
Basic notes

francisrupert:
One reason this bugged me enough to do this was because the "Loading With Label" _does_ use DtLoader in icon slot and of course that should be aligned. 
image

iropolo (approved):
Looks great!

PR #1112, opened 2026-03-09
chore(combobox-multi-select): DLT-2930 moved from recipe to component

iropolo:
# Moving combobox-multi-select from recipe to component

image

- [x] Refactor

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2930

## :book: Description

We are moving the combobox-multi-select component from recipes to component section.
Component name is now `DtComboboxMultiSelect`
We also update the style file moving it and updating d-recipe to d-.
We update the classes from `d-recipe` to `d-`.
Added a `combobox-multi-select.md` file for doc site.
We are exporting the component as `DtRecipeComboboxMultiSelect` and `DtComboboxMultiSelect`

This should not require any migration beyond Dialtone since we still export the component under `DtRecipe..`

## :bulb: Context

ComboboxMultiSelectWith is now a core dialtone component as a results of our efforts on ui-kits project.

## :crystal_ball: Next Steps

Do the same with ComboboxMultiSelectWithPopover


iropolo:
Approving expected Percy changes.

packages/dialtone-vue/components/combobox_multi_select/index.js
```diff
@@ -1,2 +1,3 @@
+export { default as DtComboboxMultiSelect } from './combobox_multi_select.vue';
 export { default as DtRecipeComboboxMultiSelect } from './combobox_multi_select.vue';
```
braddialpad, line 2: 👍

packages/dialtone-css/lib/build/less/components/combobox-multi-select.less
braddialpad, line 1: hmm lets also alias these just in case. We will get rid of them in the new major version.

```
.d-combobox-multi-select__chip,
.d-recipe-combobox-multi-select__chip {
  ...
}
```

braddialpad (changes requested):

packages/dialtone-css/lib/build/less/components/combobox-multi-select.less
iropolo: hm I did a quick search previously and there is nothing also if we update the component classes it is just useless, right?

braddialpad: In dialtone it is unused yes, but they could be used in consuming apps

iropolo: yeah.. it will be better... there could be some around, so I will  restore the classes on the component.
I will add a `TODO` to replace it later.

iropolo: https://github.com/dialpad/dialtone/pull/1112/commits/974834dee7a462d67ef111bb9506f45624bd61c1

braddialpad: That wasn't quite what I. meant, you should keep both the old class name AND the new class name like this example.

```
.d-combobox-multi-select__chip,
.d-recipe-combobox-multi-select__chip {
  ...
}
```

iropolo: We will keep `d-recipe` for now.

PR #1118, opened 2026-03-11
chore(motion-text): DLT-2932 moved from recipe to component

iropolo:
Moving motion-text from recipe to component


iropolo:
Im updating the classes here because this component is not being used at all. Is quite new.

braddialpad:
> Im updating the classes here because this component is not being used at all. Is quite new.

yep should be fine

braddialpad (approved):
Looks really good!

PR #1121, opened 2026-03-12
fix(editor): NO-JIRA empty div issue with nested block elements

maysamkf:
# Empty div issue with nested block elements

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Description

- Added logic to not convert divs with block children into paragraphs

## 💡 Context
Nesting a block element in a div would cause it to render an extra newline (see SS below).
This is likely due to how Tiptap tries to flatten the html when parsing it.

## :pencil: Checklist

For all PRs:

#### Observed Issue: 
Screenshot 2026-03-11 at 6 17 09 PM

#### Fix:
Screenshot 2026-03-11 at 6 17 05 PM

### DIFF
--- a/packages/dialtone-vue/components/rich_text_editor/extensions/div/div.js
+++ b/packages/dialtone-vue/components/rich_text_editor/extensions/div/div.js
@@ -1,13 +1,35 @@
 import { mergeAttributes } from '@tiptap/core';
 import Paragraph from '@tiptap/extension-paragraph';
 
+const BLOCK_TAGS = new Set([
+  'address', 'article', 'aside', 'blockquote', 'dd', 'details',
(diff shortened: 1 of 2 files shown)


braddialpad (commented):
This feels like a bit of a hack, is this not something that can be configured in TipTap/prosemirror itself? 

If there's truly no other option we can go with it.

maysamkf:
> This feels like a bit of a hack, is this not something that can be configured in TipTap/prosemirror itself?
> 
> If there's truly no other option we can go with it.

tried looking for other ways, but this seems to be the only way AFAIK.

braddialpad (approved):

PR #281, opened 2024-04-22
feat(dt-notice): DLT-1671 add support for truncating text

ninamarina:
# feat(dt-notice): DLT-1671 add support for truncating text

## :hammer_and_wrench: Type Of Change

These types will increment the version number on release:

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket
[DLT-1671]

## :book: Description
Adds support for truncating text for the DtNotice component

## :bulb: Context
This is a common use case in the product

## :pencil: Checklist

For all PRs:

- [x] I have ensured no private Dialpad links or info are in the code or pull request description (Dialtone is a public repo!).
- [x] I have reviewed my changes.
- [x] I have added all relevant documentation.
- [x] I have considered the performance impact of my change.

For all Vue changes:

- [ ] I have added / updated unit tests.
- [x] I have made my changes in Vue 2 and Vue 3. Note: you may sync your changes from Vue 2 to Vue 3 (or vice versa) using the `./scripts/dialtone-vue-sync.sh` script.
- [x] I have validated components with a screen reader.
- [x] I have validated components keyboard navigation.

For all CSS changes:

- [x] I have used design tokens whenever possible.
- [x] I have considered how this change will behave on different screen sizes.
- [x] I have visually validated my change in light and dark mode.
- [x] I have used gap or flexbox properties for layout instead of margin whenever possible.

## :camera: Screenshots / GIFs
notice-truncate

[DLT-1671]: https://dialpad.atlassian.net/browse/DLT-1671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #29, opened 2023-11-29
ci: improve deploy preview speed

juliodialpad:
## Description

Merged works together to avoid installing and setting up dependencies multiple times to improve timing and reuse pnpm build cache.

## Pull Request Checklist

 - [ ] Ask the contributors if a similar effort is already in process or has been solved.
 - [x] Review the [contribution guidelines](https://github.com/dialpad/dialtone/blob/staging/.github/CONTRIBUTING.md).
 - [x] Use `staging` as your pull request's base branch. (All PRs using `production` as its base branch will be declined).
 - [ ] Ensure all `gulp` scripts successfully compile.
 - [ ] Update, remove, or extend all affected documentation.
 - [ ] Ensure no private Dialpad links or info are in the code or pull request description (Dialtone is a public repo!).

PR #1142, opened 2026-03-19
feat(dialtone-cli,dialtone-query-core): DLT-3163 extract query core and add CLI

hynes-dialpad:

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

[DLT-3163](https://dialpad.atlassian.net/browse/DLT-3163)

## :book: Description

Extracts the search engine from `dialtone-mcp-server` into a shared `dialtone-query-core` package, then builds a new `dialtone-cli` on top of it.

**New packages:**
- `@dialpad/dialtone-query-core` — shared search engine (types, search functions, filters, data loading). Built with tsc, consumed by both MCP server and CLI.
- `@dialpad/dialtone-cli` — terminal interface for searching components, tokens, utility classes, and icons. Built with rollup + citty.

**Refactored:**
- `@dialpad/dialtone-mcp-server` — now imports from `dialtone-query-core` instead of owning the search logic. All existing behavior preserved (99% test pass rate unchanged).

**CLI commands:**

| Command | Purpose |
|---------|---------|
| `dialtone search <query>` | Cross-domain search, supports `\|` for OR |
| `dialtone component <name>` | Component docs with `--props`, `--prop <name>`, `--events`, `--slots`, `--describe` |
| `dialtone token <query>` | Token lookup with `--values`, `--all` for HSL |
| `dialtone utility <query>` | CSS utility class lookup |
| `dialtone prompt <name>` | LLM-optimized compact context block |

All commands support `--format minimal` (default), `--format markdown`, `--format json`, and `--limit`.

## :bulb: Context

The CLI exists as a token-efficient complement to the MCP server for AI agent workflows. MCP returns JSON over a protocol layer that costs context window tokens. The CLI returns plain text — same data, fewer tokens. Install both: CLI for fast lookups, MCP for deep exploration via resources.

**Search improvements (in core, affects both CLI and MCP):**
- HSL decomposition tokens filtered by default (559 → 91 results for "foreground color"). Pass `--all` / `{ includeHsl: true }` to include them.
- Component search now handles camelCase queries (`DtButton` resolves correctly).
- OR search via pipe separator (`input|select|menu`).

## :package: Cross-Package Impact

| Package | Changes | Downstream Impact |
|---|---|---|
| `dialtone-query-core` | **New** — extracted search engine | Foundation for MCP server and CLI |
| `dialtone-mcp-server` | Refactored to import from core | No public API changes, identical behavior |
| `dialtone-cli` | **New** — terminal interface | Available via `pnpm dialtone` or global install |
| Root `package.json` | Added `@dialpad/dialtone-cli` dependency | `dialtone` bin available in monorepo |

Dependency flow: tokens → CSS → Vue → **query-core** → MCP server / CLI

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

- Publish `@dialpad/dialtone-query-core` and `@dialpad/dialtone-cli` to npm
- Add Vitest test suite for `dialtone-query-core`
- Consider adding `dialtone` and `dialtone-mcp-server` bins to root `@dialpad/dialtone` package
- Document recommended Claude Code setup in other Dialpad repos (firespotter, web-clients)

[DLT-3163]: https://dialpad.atlassian.net/browse/DLT-3163?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dialtone-query-core/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "files": [
+    "build"
+  ],
+  "scripts": {
+    "build": "tsc"
+  },
+  "devDependencies": {
+    "@dialpad/dialtone-css": "workspace:*",
+    "@dialpad/dialtone-icons": "workspace:*",
+    "@dialpad/dialtone-vue": "workspace:^3",
```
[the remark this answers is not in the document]
hynes-dialpad: This is intentional. `dialtone-query-core` is never installed standalone by external consumers — both `dialtone-mcp-server` and `dialtone-cli` use rollup to bundle the core's compiled output, which inlines the JSON data from these packages into their own self-contained `build/index.js`. The published MCP server and CLI have zero runtime dependencies on the core or its data sources.

`devDependencies` is correct here because the data packages (`dialtone-css`, `dialtone-vue`, `dialtone-icons`) are only needed at build time to resolve the JSON imports that rollup bundles. This matches the existing pattern in `dialtone-mcp-server`'s `package.json`, which also lists these as `devDependencies` for the same reason.

If we later publish `dialtone-query-core` as a standalone library for direct consumption (not bundled), we'd need to move these to `dependencies` or switch to a rollup build for the core itself.

packages/dialtone-query-core/src/tools/components.ts
belumontoya, line 20: This gives 'D t B utton'. So a name-bucket search for "button" against DtButton won't match.

packages/dialtone-cli/src/silence-debug.ts
```diff
@@ -0,0 +1,18 @@
...
+// The core was designed for the MCP server where stderr is invisible to
+// the client. In a CLI, stderr goes to the terminal, so we silence it.
+
+const originalError = console.error;
+
+export function silenceDebug() {
+  console.error = (...args: unknown[]) => {
+    const first = String(args[0] ?? '');
+    // Swallow lines from core search/filter debug logging
+    if (first.startsWith('[') || first.startsWith('\n[')) return;
```
belumontoya: This suppresses any console.error starting with [, not just core debug lines. Node deprecation warnings ([DEP0XXX]), dependency errors, etc. all silently eaten.

packages/dialtone-query-core/src/tools/tokens.ts
```diff
@@ -58,6 +58,13 @@ export function searchTokens(query: string, data: TokensData): { results: Search
       continue;
     }
 
+    // Skip HSL decomposition tokens (-h, -s, -l, -a, -hsl, -hsla suffixes).
+    // These are machine-generated for programmatic color manipulation and
+    // will be removed in a future release.
+    if (!options?.includeHsl && /-(h|s|l|a|hsl|hsla)$/.test(tokenName)) {
```
belumontoya: The regex /-(h|s|l|a|hsl|hsla)$/ matches any token ending in -a, -h, -s, or -l — not just HSL decomposition tokens. The actual HSL tokens are always under --dt-color-*.

packages/dialtone-cli/src/formatters.ts
```diff
@@ -0,0 +1,303 @@
...
+
+// ── Prompt formatter (LLM-optimized) ────────────────────────────────────────
+
+export function formatPrompt(component: Component): string {
+  const lines: string[] = [];
+  lines.push(`<${component.displayName}>`);
+
+  if (component.description) {
+    // Truncate description to one sentence
+    const firstSentence = component.description.split(/\.\s/)[0];
```
belumontoya: e.g. , i.e. , etc.  all trigger the split. Since this feeds into LLM context (the whole point of prompt), worth fixing but not blocking.

belumontoya (changes requested):

packages/dialtone-query-core/src/tools/components.ts
hynes-dialpad: Verified — `splitCamelCase("DtButton")` actually returns `["dt", "button"]`, not `["d", "t", "b", "utton"]`. The regex `([A-Z])` inserts a space before each uppercase letter, but consecutive lowercase letters stay grouped: `" Dt Button"` → trim → split → `["Dt", "Button"]`.

It would break on consecutive uppercase like `DtRTE` → `["dt", "r", "t", "e"]`, but we don't have component names like that in the library.

packages/dialtone-cli/src/silence-debug.ts
hynes-dialpad: Good catch. Fixed — now uses an explicit allowlist of core debug prefixes (`CLASS SEARCH`, `TOKEN SEARCH`, `COMPONENT SEARCH`, `ICON SEARCH`, `FILTER`, `INIT`) instead of a blanket `[` check.

packages/dialtone-query-core/src/tools/tokens.ts
hynes-dialpad: Valid. Tightened the regex to `/^--dt-.*color.*-(h|s|l|a|hsl|hsla)$/` so it only matches color tokens. Zero false positives in current data either way, but this is more future-proof.

packages/dialtone-cli/src/formatters.ts
hynes-dialpad: Fixed. Split now uses `/\.(?=\s+[A-Z])/` — only breaks after a period followed by whitespace and an uppercase letter (actual sentence boundary). Abbreviations like `e.g.`, `i.e.` no longer trigger the split.

packages/dialtone-query-core/package.json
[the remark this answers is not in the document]
braddialpad: Agree, this is unnecessary for a package that is never meant to be published

packages/dialtone-cli/src/formatters.ts
```diff
@@ -0,0 +1,305 @@
...
+  if (filter === 'props') {
+    if (format === 'markdown') {
+      const lines = ['| Prop | Type | Description | Values |', '|------|------|-------------|--------|'];
+      (result.details.props || []).forEach((p: ComponentProp) => {
+        const type = p.type?.name || 'unknown';
+        const desc = (p.description || '').replace(/\|/g, '\\|').replace(/\n/g, ' ');
+        const values = (p.values || []).join(', ').replace(/\|/g, '\\|');
+        lines.push(`| ${p.name} | ${type} | ${desc} | ${values} |`);
+      });
+      return lines.join('\n');
```
braddialpad: very similar to the block on 177-185, does it make sense to extract it to a reusable function?

packages/dialtone-cli/src/formatters.ts
```diff
@@ -0,0 +1,305 @@
...
+  // Calculate column widths
+  const widths = headers.map((h, i) => {
+    const contentMax = Math.max(h.length, ...rows.map(r => (r[i] || '').length));
+    return maxWidths?.[i] ? Math.min(contentMax, maxWidths[i]) : contentMax;
+  });
+
+  // Box-drawing characters
+  const topBorder    = '┌' + widths.map(w => '─'.repeat(w + 2)).join('┬') + '┐';
+  const headerSep    = '├' + widths.map(w => '─'.repeat(w + 2)).join('┼') + '┤';
+  const rowSep       = '├' + widths.map(w => '─'.repeat(w + 2)).join('┼') + '┤';
```
braddialpad: These are identical?

packages/dialtone-cli/src/formatters.ts
```diff
@@ -0,0 +1,305 @@
...
+  if (str.length <= max) return str;
+  return str.slice(0, max - 1) + '…';
+}
+
+function borderedTable(headers: string[], rows: string[][], maxWidths?: number[]): string {
+  // Calculate column widths
+  const widths = headers.map((h, i) => {
+    const contentMax = Math.max(h.length, ...rows.map(r => (r[i] || '').length));
+    return maxWidths?.[i] ? Math.min(contentMax, maxWidths[i]) : contentMax;
+  });
```
braddialpad: This is currently iterating through all the rows for every column. Can we change this to be the other way around and instead iterate through each column for every row? The reason this is safer is because rows are infinitely growable while columns are relatively fixed. Best practice and smaller chance of a stack overflow.

packages/dialtone-query-core/package.json
braddialpad, line 1: I guess the main question I have here, is does `dialtone-query-core` need to be a "package"? I think it would depend on what we are planning on doing with it in the future, but right now it's basically just a library of functions that are not intended for external use and are only used internally. It has it's own package.json / project.json / ci config etc which are all really only necessary for externally consumed packages. The way it is being used now, we could just as easily add a new directory called dialtone_query_core in `dialtone/common` and it would be a whole lot simpler to just import from that folder in both projects. Wouldn't have to worry about package dependencies or the release pipeline or anything.

packages/dialtone-cli/src/formatters.ts
braddialpad, line 1: A lot of code to maintain here to render a CLI table. Should we standardize on a library like cli-table3? https://github.com/cli-table/cli-table3

Probably more flexible and easier to keep standard if we do more of these in the future.

braddialpad (changes requested):

packages/dialtone-query-core/package.json
hynes-dialpad: Fair point. The core is internal-only right now and a `common/` folder would eliminate the package.json, project.json, release config, and the workspace dependency ceremony.

The trade-off is build orchestration. As a package, NX handles the build ordering automatically (`dependsOn: ["dialtone-query-core:build"]`), and both consumers resolve it cleanly via `workspace:*`. Moving to `common/` means:
- Each consumer's tsconfig needs to include `common/` in its compilation scope
- Rollup configs need relative path aliases instead of package resolution
- NX build ordering becomes implicit (no `dependsOn`, just hope the files are there)

None of that is hard, but it's swapping one kind of config for another. If the team prefers `common/`, happy to move it — it's a straightforward refactor. The search logic and API surface don't change either way.

packages/dialtone-cli/src/formatters.ts
hynes-dialpad: Agreed. Swapping to cli-table3 — it also resolves the duplicate block (#281), identical variables (#34), and iteration order (#29) comments since all of that was in the hand-rolled borderedTable function.

packages/dialtone-query-core/package.json
braddialpad: Yeah it's a good point about the nx build ordering. I think we should keep it the way it is, maybe we can remove the `release-ci.config.cjs` and the `publish` and `release` targets from project.json so it's clear it's not releaseable.

braddialpad (approved):
Looks good, let's potentially remove the "release" config from dialtone-query-core before merging

PR #1146, opened 2026-03-20
fix(input): DP-180970 ime composition bug

braddialpad:
# fix(input): DP-180970 IME composition bug

<!--- Feel free to remove any unused sections -->

<!--- go to giphy.com, pick a gif, click share, copy gif link, paste within the () brackets below. -->

![Obligatory GIF](https://media1.giphy.com/media/v1.Y2lkPTc5MGI3NjExZWp0YXd4bzdxM2tqYmVvYm9scjdvanV2bno2bmJtcDU0anhjbXZzayZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/nLhdSinRtaL2E/giphy.gif)

<!--- Tick or place an `x` for the type of change. Should match the type in the commit message / PR title
in https://github.com/dialpad/dialtone/blob/staging/.github/COMMIT_CONVENTION.md -->

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DP-180970

## :book: Description

Fixes prod bug with japanese composition where they were unable to type into the search input

## :bulb: Context

This report from claude explains the issue

```
This is the key finding. In Chrome/Chromium (and Electron), the input event fires BEFORE compositionend, not after. The event order is:                                               
                                                                                                                                                                                        
  beforeinput (isComposing: true)                                                                                                                                                       
  input (isComposing: true)        ← DtInput blocks this                                                                                                                                
  compositionend                   ← DtInput sets isComposing = false, but no emit                                                                                                      
                                   ← NO input event follows!                                                                                                                            
                                                                                                                                                                                        
  So the Dialtone test was wrong — it triggers compositionend then input, which is Firefox's event order, not Chrome's. In Chrome/Electron:                                             
                                                                                                                                                                                        
  1. The last input during composition has isComposing: true → blocked                                                                                                                  
  2. compositionend fires → isComposing = false but no update:modelValue is emitted
  3. No further input event fires → the committed text is never sent to the parent                                                                                                      
                                                                                                                                                                                        
  This means the bug affects Chrome too, not just Electron/Native. The Dialtone test passed only because it used Firefox's event ordering.                                              
                                                                                                                                                                                        
  Confirmed fix needed in Dialtone                                                                                                                                                      
                  
  The compositionend handler must emit the final value:                                                                                                                                 
  
  compositionend: () => {                                                                                                                                                               
    this.isComposing = false;
    const val = this.$refs.input.value;
    this.$emit('input', val);                                                                                                                                                           
    this.$emit('update:modelValue', val);
  }, 
```

## :pencil: Checklist

<!--- Tick or place an `x` in all of the checkboxes that apply -->
<!--- Remove checkboxes that do not apply -->

For all PRs:

For all Vue changes:

## :crystal_ball: Next Steps

CP to prod


iropolo (approved):
yeah... we can give it a try.....

braddialpad:
Not 100% sure this is the best solution here, will look into more. Just reverted the prior composition change to fix the production bug.

braddialpad:
After looking into this some more, I think it is the best way to fix it.

PR #1155, opened 2026-03-26
fix(rich-text-editor): DP-179946 fix suggestion list positioning

braddialpad:
Attempting this fix for a suggestion list issue that I can't personally repro, but QA is able to. Should in theory fix it but I can't fully confirm.


francisrupert:
I can't say that I have much to say on an approval level, but I guess... LGTM?

PR #1147, opened 2026-03-22
docs(components): DLT-3178 DLT-3182 unify code examples into single code-example component

francisrupert:
hhhack

> [!NOTE]
> Targets `next`. 

## :hammer_and_wrench: Type Of Change

- [x] Documentation

## :book: Jira Tickets

- [DLT-3178](https://dialpad.atlassian.net/browse/DLT-3178) — Convert static htmlCode to ref-based pattern
- [DLT-3182](https://dialpad.atlassian.net/browse/DLT-3182) — Refactor code-well-header and code-example-tabs into single component

## :book: Description

Bottomline: I hate duplicating the same preview's code and code snippet.

Replaces the two-component `<code-well-header>` + `<code-example-tabs>` pattern with a single `<code-example>` component across all 58 component doc pages. Slot content is the single source of truth — a build-time markdown-it plugin auto-extracts it as the Vue code tab.

**Key changes:**

- **-6,446 net lines** — over 10K lines of manually-maintained static HTML eliminated
- **`<code-example>` component** — wraps CodeWellHeader + CodeExampleTabs internally. Handles ref wiring, htmlCode extraction, and showHtmlWarning automatically
- **Build-time plugin** (`markdown-it-code-example-source.js`) — auto-extracts slot content, entity-encodes it for Vue compiler safety, strips `data-demo-wrapper` layout wrappers
- **`data-demo-wrapper`** attribute — marks layout-only wrappers (e.g. `<dt-stack direction="row">`) for stripping from the code tab while keeping them in the live demo
- **`vueCode` override** — explicit prop for cases where the code tab genuinely differs (custom example wrappers, interactive state, placeholder syntax). ~150 justified overrides remain
- **Lint script** (`scripts/lint-doc-examples.mjs`) — 6 checks enforced via pre-commit: no static htmlCode, no self-closing `<code-example />`, no empty lines in blocks, no vueCode with empty slot, plus legacy pattern checks
- **CodeWellHeader** migrated from deprecated CSS utilities (`d-fl-center`, `d-fd-column`, `d-stack8`) to `<dt-stack>` component props
- **CSS utility cleanup** — `d-stack*`/`d-flow*` wrappers → `<dt-stack>`, `d-jc-*`/`d-ai-*` on dt-stack → `justify`/`align` props, `<legend class="d-label">` → `<dt-checkbox-group>`/`<dt-radio-group>` with `legend` prop, removed redundant `d-input-group__fieldset` class props

[...]

PR #1162, opened 2026-03-31
feat(resizable): DLT-2097 add DtResizable panel layout component

hynes-dialpad:

Obligatory GIF

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket

[DLT-2097](https://dialpad.atlassian.net/browse/DLT-2097)

**Stories:**
- [DLT-3234](https://dialpad.atlassian.net/browse/DLT-3234) — Core resize engine + constraints
- [DLT-3235](https://dialpad.atlassian.net/browse/DLT-3235) — Persistence + storage adapter
- [DLT-3236](https://dialpad.atlassian.net/browse/DLT-3236) — Keyboard accessibility (W3C separator pattern)
- [DLT-3237](https://dialpad.atlassian.net/browse/DLT-3237) — Offset, docs, CSS extraction

## :book: Description

New `DtResizable` component system — a resizable panel layout ported from Beacon and genericized for Dialtone. Three components, 10 composables, 160 tests across 8 test files, 7 Storybook stories, and a full VuePress docs page.

### Components
- **DtResizable** — Layout orchestrator (direction, collapse rules, storage, offset, i18n messages)
- **DtResizablePanel** — Individual panel (constraints, collapse, initial size)
- **DtResizableHandle** — Drag handle (W3C keyboard resize, ARIA separator, offset positioning)

### Key Features
1. **Core resize** — Drag handles between panels with proportional sizing, row/column direction
2. **Dual constraint hierarchy** — userMin/Max (drag limits) + systemMin/Max (viewport resize limits)
3. **Collapse** — Manual + auto-collapse with priority-based collapseRules, space allocation strategies
4. **Persistence** — `storageKey` for localStorage, `:storage` prop for custom adapters (Pinia/Vuex/API)
5. **Keyboard accessibility** — W3C separator pattern: always-focusable handles, arrow keys resize on focus, Enter collapses, Home/End for min/max, R resets, Escape blurs. ARIA live announcements.
6. **Offset positioning** — Parent-level `offsetElement`/`offsetAmount` props. Handles and panel content offset automatically from fixed headers or toolbars.
7. **i18n** — All ARIA strings configurable via `messages` prop with `{placeholder}` templates

### Architecture
- Pure layout engine (`computeLayout.ts`) — single source of truth for all panel positions. Zero Vue dependencies.
- **One-way data flow**: `User Action → savedState → computeLayout reruns → Vue renders`
- Shadow DOM drag (`useResizableDrag.ts`) — inline styles during pointer move for 60fps, commits to savedState on pointer up
- Keyboard follows same shadow → commit discipline as drag
- Constraint resolver (`constraintResolver.ts`) — 3-tier system (user → system → collapse) applied within computeLayout
- Styles in `packages/dialtone-css/lib/build/less/components/resizable.less`

## :package: Cross-Package Impact

| Package | Changes | Downstream Impact |
|---|---|---|
| dialtone-vue | New component (3 SFCs + 10 composables) | Documentation site needs component page |
| dialtone-css | New `resizable.less` component styles | Included in CSS build |
| dialtone-documentation | New VuePress page + sidebar entry + 5 example components | None |

## :page_facing_up: Documentation Artifacts

| Artifact | Status | Notes |
|---|---|---|
| Vue source | :white_check_mark: | 3 components in `packages/dialtone-vue/components/resizable/` |
| Tests | :white_check_mark: | 160 tests across 8 test files |
| Storybook stories | :white_check_mark: | 7 stories + MDX docs page |
| Component docs JSON | :white_check_mark: | All 3 components generate props/events/slots via vue-docgen-api |
| VuePress docs | :white_check_mark: | `docs/components/resizable.md` with full API tables |
| MCP server data | :warning: | Will update after next MCP server data rebuild |
| components_list.js | :white_check_mark: | All 3 components registered |
| Package exports | :white_check_mark: | Exported from `packages/dialtone-vue/index.js` |

## :bulb: Context

This is an FY27 Q1 Key Result under the "Dialtone Next and Beacon Alignment" initiative (DLT-2979). The resizable panel system was originally built for Beacon's sidebar/main/peek layout. This PR ports it to Dialtone as a first-class component, genericized for any resizable panel layout use case.

### Reviewer feedback addressed
- **Brad**: computeLayout as single source of truth (eliminated 15 direct mutations across 3 files), deleted ~510 lines of dead code (useDOMCache, useResizablePanelSizing, handleViewportResize, dead exports), keyboard mutation discipline
- **Francis**: W3C separator keyboard pattern (deleted edit mode), CSS extraction to dialtone-css, d-* prefix, design tokens, ARIA (controls, valuetext, disabled, i18n), Enter/Home/End/R keys, test quality (|| true removal, drag tests, event emission tests)
- **Codex bot**: P1 edit mode listener (moot — deleted), P1 keyboard cursor (confirmed false positive), P2 resetBehavior (fixed — all modes differentiated)

## :pencil: Checklist

For all PRs:

[DLT-2097]: https://dialpad.atlassian.net/browse/DLT-2097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DLT-3234]: https://dialpad.atlassian.net/browse/DLT-3234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DLT-3235]: https://dialpad.atlassian.net/browse/DLT-3235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DLT-3236]: https://dialpad.atlassian.net/browse/DLT-3236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DLT-3237]: https://dialpad.atlassian.net/browse/DLT-3237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/dialtone-vue/components/resizable/dt_resizable_handle.vue
```diff
@@ -0,0 +1,420 @@
...
+function handleBlurEvent () {
+  keyboard.handleBlur();
+}
+</script>
+
+<style lang="less">
+// ─── Resize Handle ───
+// Positioned absolutely between panels. The layout engine sets inline `left`
+// (or `top` for column). The hit area extends beyond the visible bar via ::before.
+.dt-resizable-handle {
```
francisrupert: * Dialtone styles for core components are `d-*`, not `dt-*`. 
* Dialtone styles for core components are not within vue files. Rip them all out into a `src/dialtone/packages/dialtone-css/lib/build/less/components/resizable.less`

packages/dialtone-vue/components/resizable/dt_resizable_handle.vue
```diff
@@ -0,0 +1,420 @@
...
+  }
+
+  // ─── State Modifiers ───
+  &--active {
+    // Active state handled via &__indicator selector above
+  }
+
+  &--disabled {
+    cursor: default;
+    opacity: 0.3;
```
francisrupert:
```suggestion
    opacity: var(--dt-opacity-500);
```

packages/dialtone-vue/components/resizable/dt_resizable_handle.vue
```diff
@@ -0,0 +1,420 @@
...
+    pointer-events: none;
+    background-color: transparent !important;
+  }
+
+  // ─── Indicator (visible drag line) ───
+  &__indicator {
+    position: absolute;
+    border-radius: var(--dt-size-radius-pill);
+    background-color: transparent;
+    transition: background-color 150ms ease;
```
francisrupert:
```suggestion
    transition: background-color var(--td150) ease;
```

packages/dialtone-vue/components/resizable/dt_resizable_handle.vue
```diff
@@ -0,0 +1,420 @@
...
+    top: 0;
+    bottom: 0;
+    left: 0;
+    right: 0;
+  }
+
+  // ─── Hover ───
+  &:hover:not(&--disabled) &__indicator,
+  &--active &__indicator {
+    background-color: var(--dt-color-focus-ring);
```
francisrupert:
```suggestion
    background-color: var(--dt-color-border-focus);
```

packages/dialtone-vue/components/resizable/dt_resizable_handle.vue
```diff
@@ -0,0 +1,420 @@
...
+  }
+
+  &:focus-visible {
+    outline: var(--dt-size-200) solid var(--dt-color-focus-ring);
+    outline-offset: var(--dt-size-200);
+    background-color: var(--dt-color-focus-ring);
+    box-shadow:
+      0 0 0 var(--dt-size-200) var(--dt-color-surface-primary),
+      0 0 0 var(--dt-size-300) var(--dt-color-focus-ring);
+  }
```
francisrupert:
```suggestion
  &:focus-visible {
    background-color: var(--dt-color-border-focus);
    box-shadow: none;
  }
```
No need for the extra outline and shadow. Just the background color is enough.

francisrupert (dismissed):
Just an initial quick response for style.

packages/dialtone-vue/components/resizable/dt_resizable_handle.vue
hynes-dialpad: @francisrupert I kept them with the component because, like recipes/UI kits, you really can't use the component on it's own. Many of the components in `dialtone-css` could also still be used to style the component but there's not much you can do without JS. I can still move them, but the classes in `dialtone-css` would be pointless without the Vue component.

hynes-dialpad: Fixed in `03fa836` — renamed all CSS classes from `dt-resizable` to `d-resizable`. Also extracted all styles from Vue SFC `<style>` blocks into `packages/dialtone-css/lib/build/less/components/resizable.less` in `2f81287`.

packages/dialtone-vue/components/resizable/dt_resizable_handle.vue
hynes-dialpad: Applied in `6481ee0` — using `var(--dt-opacity-500)`.

packages/dialtone-vue/components/resizable/dt_resizable_handle.vue
hynes-dialpad: Applied in `6481ee0` — using `var(--td150)`.

packages/dialtone-vue/components/resizable/dt_resizable_handle.vue
hynes-dialpad: Applied in `6481ee0` — using `var(--dt-color-border-focus)` via the component variable `--d-resizable-handle-color-surface`.

packages/dialtone-vue/components/resizable/dt_resizable_handle.vue
hynes-dialpad: Applied in `6481ee0` — simplified to `background-color` + `box-shadow: none`.

packages/dialtone-vue/components/resizable/composables/useResizableEditMode.ts
```diff
@@ -0,0 +1,445 @@
...
+
+  onUnmounted(() => {
+    if (announcementTimeout) {
+      clearTimeout(announcementTimeout);
+    }
+
+    if (globalEditMode.keydownListener === handleGlobalKeydown) {
+      document.removeEventListener('keydown', globalEditMode.keydownListener);
+      document.removeEventListener('click', handleDocumentClick, true);
+      globalEditMode.keydownListener = null;
```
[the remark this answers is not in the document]
francisrupert: @hynes-dialpad Claude says:

The bug is real: line 419 checks function identity (=== handleGlobalKeydown), so if instance A registered the listener and unmounts first, B loses it. However, this only matters if you have multiple DtResizable groups on the same page and they unmount in a different order than they mounted. That's a real but uncommon scenario. The fix is also trivial — use a ref-count (like they already do for announcementRefCount on line 404/425) instead of identity checking. Worth fixing, easy fix, but not a showstopper for most usage.

packages/dialtone-vue/components/resizable/composables/useResizableKeyboard.ts
```diff
@@ -0,0 +1,298 @@
...
+
+    const resizeResult = resizeHandler.processResizeMove(
+      newCursorPosition,
+      beforePanel,
+      afterPanel,
+      containerSize.value,
+      `${beforePanelId.value}:${afterPanelId.value}`,
+      panels.value,
+      0,
+    );
```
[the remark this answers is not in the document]
francisrupert: @hynes-dialpad Claude says:

**I'd call this a false positive** — the keyboard path is using a different coordinate frame intentionally (or accidentally-correctly). The tests would confirm this

The drag path correctly reads beforePanelLeft from the DOM (useResizableDrag.ts:251), but the keyboard path hardcodes 0 (useResizableKeyboard.ts:209). Looking at processResizeMove (line 372-373), it does proposedBeforeSize = rawCursorPosition - beforeLeft, so when beforeLeft is 0, the math works out only for the first handle. For a second or third handle, the cursor position would be wrong. However — looking at line 200, newCursorPosition = beforePanel.pixelSize + delta, which is relative to the panel, not the container. So passing 0 as beforePanelLeft actually makes the math consistent: proposedBeforeSize = (pixelSize + delta) - 0 = pixelSize + delta. The Codex comment assumes processResizeMove always needs container-relative coordinates, but when beforePanelLeft is 0, the cursor is effectively panel-relative and the subtraction is a no-op.

packages/dialtone-vue/components/resizable/composables/useResizablePanelControls.ts
```diff
@@ -0,0 +1,617 @@
...
+    beforePanelId?: string,
+    afterPanelId?: string,
+    behavior: 'both' | 'before' | 'after' | 'all' = 'all'
+  ) {
+    try {
+      if (behavior === 'all') {
+        resetAllPanelPairs();
+      } else {
+        resetSpecificPanelPair(beforePanelId, afterPanelId);
+      }
```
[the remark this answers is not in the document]
francisrupert: @hynes-dialpad, Claude says:

The code treats 'both', 'before', and 'after' identically. The prop is documented and exposed on DtResizableHandle. **Whether this is a bug or an unfinished feature depends on the author's intent** — if resetBehavior="before" is supposed to only reset the before panel to its initial size while leaving the after panel alone, that logic doesn't exist. 

It's also possible the author shipped 'both' as the default and MVP, **planning to differentiate later**. Not broken in practice since the default is 'both'.

packages/dialtone-vue/components/resizable/composables/useResizableEditMode.ts
```diff
@@ -0,0 +1,445 @@
...
+        }),
+      );
+    });
+
+    announce(msg.allPanelsReset);
+  }
+
+  // ─── Global keyboard handler ──────────────────────────────────────────
+
+  function handleEditModeToggle(event: KeyboardEvent): boolean {
```
francisrupert: I've not seen this kind of mechanism for entering resize mode. It conflicts with the w3 spec and haven't seen it in other major design systems that have a "splitter/resizable" feature. 

I recommend Tab to the separator, arrow keys resize immediately. That's the interaction pattern keyboard power users and  screen reader user will expect.

The `aria-description` hints at this ("Press Control+E to enter panel edit mode"), **__BUT__** you have to already be focused on the handle to hear that description — and they can't focus it because it's tabindex="-1". 

Standard solution is simpler: separator is always focusable, arrow keys only resize when the separator has focus. That's how every other implementation works and what the spec prescribes.

packages/dialtone-vue/components/resizable/composables/useResizableEditMode.ts
```diff
@@ -0,0 +1,445 @@
...
+    if (
```
francisrupert: Additionally, Ctrl+E activates edit mode for **everything**, and Tab cycles through handles from all groups indiscriminately. Confusing with multiple resizable regions on a single view.

packages/dialtone-vue/components/resizable/resizable_handle.vue
```diff
@@ -0,0 +1,341 @@
...
+        'd-resizable-handle--active': isActive,
+        'd-resizable-handle--disabled': isDisabled,
+        'd-resizable-handle--edit-mode': isEditMode,
+      },
+    ]"
+    :style="handleStyles"
+    :data-handle-id="handleId"
+    :tabindex="isEditMode ? '0' : '-1'"
+    role="separator"
+    :aria-orientation="direction === 'row' ? 'vertical' : 'horizontal'"
```
francisrupert: Need `aria-controls`. The separator shouldo reference the primary pane via aria-controls. This lets assistive tech navigate from the handle to the pane it affects.

packages/dialtone-vue/components/resizable/composables/useResizableKeyboard.ts
```diff
@@ -0,0 +1,272 @@
...
+        ),
+      );
+    }
+
+    return true;
+  }
+
+  // ─── Event handlers ─────────────────────────────────────────────────
+
+  function handleKeyDown(event: KeyboardEvent): void {
```
francisrupert: Add Enter key to collapse/restore while focused. The spec defines Enter as the way to toggle collapse on a fixed splitter while focused. The component has collapsible panels but doesn't wire Enter to that behavior.

packages/dialtone-vue/components/resizable/resizable_handle.vue
```diff
@@ -0,0 +1,341 @@
...
+    :aria-label="computedAriaLabel"
+    :aria-valuenow="ariaValueNow"
```
francisrupert, line 20: Add aria-valuetext — e.g. "Pane width 256 pixels". Raw aria-valuenow as a percentage is less useful without context.

packages/dialtone-vue/components/resizable/composables/useResizableEditMode.ts
```diff
@@ -0,0 +1,445 @@
...
+        enterEditMode();
+      } else {
+        exitEditMode();
+      }
+      return true;
+    }
+    return false;
+  }
+
+  function handleResetKey(event: KeyboardEvent): boolean {
```
francisrupert: "CMD R" to reset. Yikes, too easy to accidentally reload window. Can just be `R` since separator is focused.

packages/dialtone-vue/components/resizable/composables/useResizableKeyboard.ts
```diff
@@ -0,0 +1,272 @@
...
+        generateSizeAnnouncement(
+          beforePanel, afterPanel, resizeDirection, incrementPixels,
+        ),
+      );
+    }
+
+    return true;
+  }
+
+  // ─── Event handlers ─────────────────────────────────────────────────
```
francisrupert: `Home` and `End` keys for minimum and maximum size.

francisrupert (commented):
A few comments mainly on keyboarding/accessibility.

packages/dialtone-vue/components/resizable/resizable_handle.vue
```diff
@@ -0,0 +1,341 @@
...
+    :aria-valuemin="ariaValueMin"
+    :aria-valuemax="ariaValueMax"
+    :aria-keyshortcuts="ariaKeyShortcuts"
+    :aria-description="ariaDescription"
```
francisrupert: Need `:aria-disabled="isDisabled"`

packages/dialtone-vue/components/resizable/resizable_keyboard.test.js
```diff
@@ -0,0 +1,167 @@
...
+      );
+    });
+  });
+
+  describe('Keyboard events on handle', () => {
+    beforeEach(() => {
+      _setWrapper();
+    });
+
+    it('should listen for keydown events', () => {
```
francisrupert: three tests `use || true`, making them always pass. Zero value

packages/dialtone-vue/components/resizable/composables/useResizableDrag.ts
```diff
@@ -0,0 +1,384 @@
+/**
```
francisrupert, line 1: `useResizableDrag.ts` has zero tests. No mousedown/mousemove/mouseup sequences, no Escape-to-cancel, etc.

packages/dialtone-vue/components/resizable/resizable.test.js
```diff
@@ -0,0 +1,181 @@
+/**
```
francisrupert, line 1: No event emission tests — panel-resize, panel-collapse, resize-start, resize-end, panel-peek-start, panel-peek-end — none verified with wrapper.emitted().

packages/dialtone-vue/components/resizable/resizable_handle.vue
```diff
@@ -0,0 +1,341 @@
...
+  return {
+    insetInlineStart: `${Math.max(0, pos.left)}px`,
+    visibility: '',
+    ...offset.handleStyles.value,
+  };
+});
+
+// ── ARIA values ──────────────────────────────────────────────────────────────
+
+const computedAriaLabel = computed(() => {
```
francisrupert, line 142: aria-label has no messages override — this builds "Resize handle between {before} and {after} panels" with hardcoded English The ariaLabel prop overrides per-handle, but there's no template key in the messages prop for batch translation.

packages/dialtone-vue/components/resizable/composables/useResizableKeyboard.ts
```diff
@@ -0,0 +1,272 @@
...
+
+  function generateSizeAnnouncement(
+    beforePanel: ResizablePanelState,
+    afterPanel: ResizablePanelState,
+    resizeDirection: 'increase' | 'decrease',
+    increment: number,
+  ): string {
+    const beforePx = Math.round(beforePanel.pixelSize);
+    const afterPx = Math.round(afterPanel.pixelSize);
+    const action =
```
francisrupert: `{action}` and `{incrementType}` are untranslatable, i.e. "increased", "decreased", "fine", "large", "normal".

packages/dialtone-vue/components/resizable/composables/useResizablePanelControls.ts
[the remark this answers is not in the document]
hynes-dialpad: Fixed in `29b13cb` — `resetPanels` now branches on all four behaviors. Added `resetSinglePanel()` for before/after modes that resets one panel and absorbs the delta into the nearest unlocked neighbor.

packages/dialtone-vue/components/resizable/composables/useResizableEditMode.ts
hynes-dialpad: Agreed — removed the edit mode system entirely in `eb5e2ea`. Handles are now always `tabindex="0"`, arrow keys resize on focus, Tab moves between handles via native browser focus. No more global Ctrl+E or cross-group Tab cycling. The 445-line `useResizableEditMode.ts` is replaced by a 70-line `useResizableAnnouncements.ts` that only manages the aria-live region.

packages/dialtone-vue/components/resizable/composables/useResizableEditMode.ts
hynes-dialpad: Addressed in `eb5e2ea` — edit mode removed entirely. Each handle is independently focusable, scoped to its own group via DOM structure.

packages/dialtone-vue/components/resizable/resizable_handle.vue
hynes-dialpad: Added in `eb5e2ea` — handle gets `:aria-controls="'dt-resizable-panel-' + resolvedBeforePanelId"`. Panel root element now has `:id="'dt-resizable-panel-' + props.id"`.

packages/dialtone-vue/components/resizable/composables/useResizableKeyboard.ts
hynes-dialpad: Added in `eb5e2ea` — Enter toggles collapse on the before panel (if collapsible). Announces state change via aria-live.

packages/dialtone-vue/components/resizable/resizable_handle.vue
hynes-dialpad: Added in `eb5e2ea` — `:aria-valuetext` renders e.g. "sidebar: 256px". Template is i18n-configurable via `messages.ariaValueText`.

packages/dialtone-vue/components/resizable/composables/useResizableEditMode.ts
hynes-dialpad: Fixed in `eb5e2ea` — R key (no modifier) resets adjacent panels when handle is focused. Cmd+R removed entirely.

packages/dialtone-vue/components/resizable/composables/useResizableKeyboard.ts
hynes-dialpad: Added in `eb5e2ea` — Home sets panel to minimum size, End sets to maximum.

packages/dialtone-vue/components/resizable/resizable_handle.vue
hynes-dialpad: Added in `eb5e2ea` — `:aria-disabled="isDisabled || undefined"`. Disabled handles also get `tabindex="-1"`.

packages/dialtone-vue/components/resizable/resizable_keyboard.test.js
hynes-dialpad: Fixed in `22e9c96` — replaced with real interaction tests. Added 25 new tests total covering Enter, Home, End, R, aria-controls, aria-valuetext, aria-disabled, event emissions, and drag.

packages/dialtone-vue/components/resizable/composables/useResizableDrag.ts
hynes-dialpad: Added `resizable_drag.test.js` in `22e9c96` — 6 tests covering handle rendering, mousedown, disabled blocking, touch events, double-click reset, and indicator element.

packages/dialtone-vue/components/resizable/resizable.test.js
hynes-dialpad: Added in `22e9c96` — tests for `panel-collapse` and `panel-resize` via `wrapper.emitted()`, plus verification that all 4 event types are defined.

packages/dialtone-vue/components/resizable/resizable_handle.vue
hynes-dialpad: Fixed in `eb5e2ea` — `computedAriaLabel` now reads from `messages.handleAriaLabel` with `{before}` and `{after}` placeholders, falling back to the English default.

packages/dialtone-vue/components/resizable/composables/useResizableKeyboard.ts
hynes-dialpad: Fixed in `eb5e2ea` — simplified resize announcement to `'{beforeId}: {beforePx}px, {afterId}: {afterPx}px'`. No English verbs. Added separate `collapseAnnouncement`, `expandAnnouncement`, `resetAnnouncement` message keys.

packages/dialtone-vue/components/resizable/composables/useResizablePanelControls.ts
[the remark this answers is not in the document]
hynes-dialpad: Fixed in `29b13cb` — `resetPanels` now branches on all four behaviors. Added `resetSinglePanel()` for before/after modes that resets one panel and absorbs the delta into the nearest unlocked neighbor.

francisrupert:
Wonkiness on three-panel key'ing. Can't resize second splitter to the right. Hitting _either_ left or right continues to resize to the left, and seemingly too much. See attached video. 

https://github.com/user-attachments/assets/e40de647-b5b3-4e51-9e35-1b7f7d939230

francisrupert:
Peek. I have no idea how to work with the `peek` functionality. [story/components-resizable--peek-hover](https://dialtone.dialpad.com/vue/deploy-previews/pr-1162/?path=/story/components-resizable--peek-hover) or [story/components-resizable--peek-button](https://dialtone.dialpad.com/vue/deploy-previews/pr-1162/?path=/story/components-resizable--peek-button) aren't helpful, let alone on http://localhost:4000/components/resizable.html

I hooked it up locally as I can understand it is meant to be, but the use of `pointer-events: none` and the trigger being in the collapsed container with a width of zero and clipped out of view suggests to me this doesn't actually work? The trigger and its overlay can't be surfaced.

apps/dialtone-documentation/docs/.vuepress/exampleComponents/ExampleResizable.vue
```diff
@@ -0,0 +1,21 @@
...
+  <div class="d-w100p" style="height: 300px;">
+    <dt-resizable>
+      <dt-resizable-panel id="ex-sidebar" initial-size="25p" user-min-size="825">
+        <div class="d-d-flex d-ai-center d-jc-center d-w100p d-h100p d-bgc-purple-100">
+          <span class="d-fs-200 d-fw-bold d-fc-purple-400">Sidebar</span>
+        </div>
+      </dt-resizable-panel>
+      <dt-resizable-handle />
+      <dt-resizable-panel id="ex-content">
+        <div class="d-d-flex d-ai-center d-jc-center d-w100p d-h100p d-bgc-gold-100">
```
francisrupert: All of these example vue files are using `d-d-flex` instead of DtStack.

apps/dialtone-documentation/docs/components/resizable.md
```diff
@@ -0,0 +1,432 @@
+---
+title: Resizable
+description: A layout component that lets users resize adjacent panels by dragging a handle between them.
+status: ready
```
francisrupert:
```suggestion
status: beta
```

apps/dialtone-documentation/docs/components/resizable.md
```diff
@@ -0,0 +1,432 @@
...
+
+### Examples
+
+#### Two panels
+
+<code-well-header custom class="d-p24 d-bgc-secondary d-bar8">
+  <example-resizable />
+</code-well-header>
+
+```vue
```
francisrupert, line 25: Each of these fenced vue blocks should be using `<code-example-tabs>`. e.g. 
```html
<code-example-tabs
vueCode='
<dt-resizable-panel>
  ...
</dt-resizable-panel>
'
/>
```

No need to use `htmlCode`.

When this is folded into `next`, those would be migrated to use `<code-example>` which now covers _both_ the preview and code snippet. View on `next` branch to see what that looks like now, or the PR that did that https://github.com/dialpad/dialtone/pull/1147

Not using fenced blocks will make those easy to migrate with Claude skill.

packages/dialtone-vue/components/resizable/resizable_offset.test.js
```diff
@@ -0,0 +1,118 @@
...
+
+// Mock ResizeObserver
+global.ResizeObserver = vi.fn().mockImplementation(() => ({
+  observe: vi.fn(),
+  unobserve: vi.fn(),
+  disconnect: vi.fn(),
+}));
+
+// Mock MutationObserver (used by useDOMCache)
+global.MutationObserver = vi.fn().mockImplementation(() => ({
```
francisrupert, line 22: CI failure is from here. 

`MutationObserver` is mocked as a regular function but `useDOMCache.ts` line `264` calls it with `new`.

Arrow functions aren't constructors, so CI's jsdom throws `TypeError: () => ({...}) is not a constructor`.

Passes locally because your Node/jsdom version has a real `MutationObserver` that shadows the mock.

**Presumed Fix:**

```javascript
// resizable_offset.test.js:22
global.MutationObserver = vi.fn().mockImplementation(function (callback) {                                                                 
  this.observe = vi.fn();                                                                                                                  
  this.disconnect = vi.fn();                                                                                                               
  this.callback = callback;                                                                                                                
}); 
```

packages/dialtone-vue/common/composables/useDOMCache.ts
braddialpad, line 1: I don't think we really need this whole DOM caching module, it is a TON of complicated code and it seems like the only time we use it is a single querySelector on mount in useResizeable offset? Just regular `document.querySelector` should be plenty sufficient. Did AI suggest this? I don't think this sort of thing is often done in web development.

packages/dialtone-vue/components/resizable/composables/useResizableKeyboard.ts
braddialpad, line 1: `useResizableDrag` correctly defers reactive mutations to `onDragEnd` — during drag it only writes inline styles. `useResizableKeyboard` does the opposite: it simultaneously writes inline DOM styles and mutates `panel.pixelSize` reactively, which can trigger a recompute that overwrites the styles it just set. The two systems need to follow the same discipline.

packages/dialtone-vue/components/resizable/composables/useResizablePanelControls.ts
```diff
@@ -0,0 +1,646 @@
...
+    const overflow = calculateOverflow(targetSizes, newContainerSize);
+    redistributeIfNeeded(activePanels, targetSizes, overflow);
+
+    applyFinalSizes(activePanels, targetSizes);
+    updateCollapsedPanelConstraints(newContainerSize);
+
+    processAutoCollapseExpand();
+
+    if (!isInitializing.value) saveToStorage(panels.value);
+  }
```
braddialpad: From what I understand this should be the responsibility of computeLayout right?

fold this + helpers into computeLayout or trigger a rerun of computeLayout upon viewport resize and handle it there.

packages/dialtone-vue/components/resizable/composables/useResizableCalculations.ts
```diff
@@ -0,0 +1,499 @@
...
+
+  return {
+    panelSizeInfo,
+    initializePanelSize,
+    updatePanelsForViewportChange,
+    ensureAtLeastOneUnlocked,
+    getTotalLockedPixels,
+    getRemainingSpace,
+    convertInitialSizeToPixels,
+  };
```
braddialpad: This appears to be dead code / never used.

packages/dialtone-vue/components/resizable/composables/useResizableCalculations.ts
```diff
@@ -0,0 +1,499 @@
...
+  function getEventPosition(event: MouseEvent | TouchEvent) {
+    if ('touches' in event) {
+      return { clientX: event.touches[0].clientX, clientY: event.touches[0].clientY };
+    }
+    return { clientX: event.clientX, clientY: event.clientY };
+  }
+
+  function getCoordinate(position: { clientX: number; clientY: number }) {
+    return direction === 'row' ? position.clientX : position.clientY;
+  }
```
braddialpad: never used

packages/dialtone-vue/components/resizable/composables/useResizableCalculations.ts
```diff
@@ -0,0 +1,499 @@
...
+  function calculateNewSizes(delta: number, startBeforeSize: number, startAfterSize: number) {
+    const totalSize = startBeforeSize + startAfterSize;
+    const currentContainerSize = containerSize ? containerSize() : 1000;
+    const deltaPercentage = (delta / currentContainerSize) * 100;
+
+    const newBeforeSize = startBeforeSize + deltaPercentage;
+    const newAfterSize = totalSize - newBeforeSize;
+
+    return { newBeforeSize, newAfterSize };
+  }
```
braddialpad: never used

packages/dialtone-vue/components/resizable/composables/useResizableCalculations.ts
```diff
@@ -0,0 +1,499 @@
...
+    const newBeforePixels = (beforePanel.pixelSize || 0) + deltaPixels;
+    const newAfterPixels = (afterPanel.pixelSize || 0) - deltaPixels;
+
+    if (skipConstraints) {
+      return applyBasicBoundsCheck(beforePanel, afterPanel, newBeforePixels, newAfterPixels);
+    }
+
+    applyFullConstraints(beforePanel, afterPanel, newBeforePixels, newAfterPixels, containerSizeValue);
+    return true;
+  }
```
braddialpad: never used

packages/dialtone-vue/components/resizable/composables/useResizableCalculations.ts
```diff
@@ -0,0 +1,499 @@
...
+  containerElement: ComputedRef<HTMLElement | null>;
+}
+
+export interface PanelSizeInfo {
+  isLocked: boolean;
+  pixelSize: number;
+  absoluteStyles: Record<string, string>;
+}
+
+export function useResizablePanelSizing(options: ResizablePanelSizingOptions) {
```
braddialpad: the work of this function should be done in computeLayout?

packages/dialtone-vue/components/resizable/composables/computeLayout.ts
braddialpad, line 1: It appears to me that compute layout is supposed to be the single source of truth for all layout data? It doesn't appear to be working that way right now. It triggers when certain params change (panels, containerSize, savedState) but there are many internal mutations that are NOT run through here when they are changed. We basically have multiple sources of truth. Here is a list of mutations that are not run through here from claude:

### `useResizablePanelControls.ts`

| Line | Mutation | Operation |
|------|----------|-----------|
| 154 | `panel.collapsed = true` | `executeCollapse` |
| 155 | `panel.autoCollapsed = source === 'system'` | `executeCollapse` |
| 166 | `panel.collapsed = false` | `executeExpand` |
| 167 | `panel.autoCollapsed = undefined` | `executeExpand` |
| 180 | `panel.pixelSize = savedPanelSize` | `executeExpand` — restoring pre-collapse size |
| 200 | `panel.pixelSize = constrainedSize` | `executeExpand` — no saved state fallback |
| 212 | `panel.pixelSize = constrainedSize` | `resizePanel` |
| 235 | `panel.locked = true` | `lockPanel` |
| 245 | `panel.locked = false` | `unlockPanel` |
| 312 | `panel.pixelSize = initialSize` | `resetSinglePanel` |
| 314 | `panel.locked = false` | `resetSinglePanel` |
| 622 | `panel.pixelSize = finalSize` | `handleViewportResize` |

### `useResizableCalculations.ts`

| Line | Mutation | Operation |
|------|----------|-----------|
| 236 | `panel.pixelSize = Math.round(panel.pixelSize * scaleFactor)` | proportional scale during drag |

### `useResizableStorage.ts`

| Line | Mutation | Operation |
|------|----------|-----------|
| 172 | `panel.pixelSize = savedPanel.pixelSize` | storage restore |
| 175 | `panel.locked = savedPanel.locked` | storage restore |
| 178 | `panel.collapsed = savedPanel.collapsed` | storage restore |
| 181 | `panel.autoCollapsed = savedPanel.autoCollapsed` | storage restore |

That's 15 direct mutations across 3 files, none of which flow back into computeLayout's inputs. The storage restore ones are arguably the most dangerous — restoring from storage is exactly the scenario where you'd want the layout engine to validate and reconcile sizes against the current container dimensions.

braddialpad (changes requested):
Alright so I'm going to start with a high level architecture review rather than smaller code improvements. This will be by far the most complex component we have in Dialtone once merged. I have spent the entire morning just trying to understand what is going on here.

Main architectural concern here is with how panel state is handled.

packages/dialtone-vue/components/resizable/composables/useResizableEditMode.ts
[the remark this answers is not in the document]
hynes-dialpad: Moot — the entire edit mode system was removed in `bdd2aff`. Keyboard now uses the W3C separator pattern: handles are always focusable (tabindex="0"), arrow keys resize immediately on focus. No global listeners.

packages/dialtone-vue/components/resizable/composables/useResizableKeyboard.ts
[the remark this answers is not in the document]
hynes-dialpad: Confirmed — Francis's assessment is correct. Keyboard passes beforePanelLeft=0 intentionally since newCursorPosition is already panel-relative. processResizeMove's subtraction is a no-op when beforePanelLeft is 0. Tests cover this in `resizable_keyboard.test.js`.

packages/dialtone-vue/components/resizable/composables/useResizablePanelControls.ts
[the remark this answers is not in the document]
hynes-dialpad: Fixed — `resetBehavior` values (`both`, `before`, `after`, `all`) are now fully differentiated in `useResizablePanelControls.ts`. `before` resets only the before panel to initial size, `after` only the after panel, `both` resets the pair, `all` resets every panel.

packages/dialtone-vue/components/resizable/composables/useResizableEditMode.ts
hynes-dialpad: Agreed and fixed — edit mode removed entirely in `bdd2aff`. Now follows the W3C Window Splitter pattern: separator is always tabindex="0", arrow keys resize immediately on focus, no global Ctrl+E listener. The file `useResizableEditMode.ts` was deleted.

packages/dialtone-vue/components/resizable/composables/useResizableEditMode.ts
hynes-dialpad: Fixed — the entire `useResizableEditMode.ts` composable was deleted in `bdd2aff`. No more global Ctrl+E, no more cross-group Tab cycling. Each handle is independently focusable via normal Tab order.

packages/dialtone-vue/components/resizable/resizable_handle.vue
hynes-dialpad: Added in `8448e91` — handle now has `:aria-controls="\`dt-resizable-panel-${resolvedBeforePanelId}\`"`. Panel element gets a matching `id` attribute. Test coverage in `resizable_keyboard.test.js`.

packages/dialtone-vue/components/resizable/composables/useResizableKeyboard.ts
hynes-dialpad: Added in `025912819` — Enter toggles collapse on the before panel when focused. Uses `collapsePanel` callback from context. Covered in keyboard test suite.

packages/dialtone-vue/components/resizable/resizable_handle.vue
hynes-dialpad: Added in `8448e91` — `:aria-valuetext` now shows "{panelId}: {pixels}px" (e.g. "sidebar: 256px"). Template is overridable via the `messages.ariaValueText` prop for i18n.

packages/dialtone-vue/components/resizable/composables/useResizableEditMode.ts
hynes-dialpad: Fixed — reset is now plain `R` key (no modifier). Only fires when the separator has focus, so no conflict with browser reload. Part of the W3C keyboard rewrite in `025912819`.

packages/dialtone-vue/components/resizable/composables/useResizableKeyboard.ts
hynes-dialpad: Added in `025912819` — Home sets panel to minimum size, End sets to maximum. Both respect userMinSize/userMaxSize constraints. Tested in keyboard increment test suite.

packages/dialtone-vue/components/resizable/resizable_handle.vue
hynes-dialpad: Added in `8448e91` — `:aria-disabled="isDisabled || undefined"`. Renders as `aria-disabled="true"` when disabled, omitted when not. Test coverage in `resizable_keyboard.test.js` under 'When handle is disabled'.

packages/dialtone-vue/components/resizable/resizable_keyboard.test.js
hynes-dialpad: Fixed — the `|| true` in keyboard tests was removed in earlier refactoring (the tests were rewritten for the W3C pattern). The last remaining `|| true` was in `resizable_drag.test.js` and was fixed in `0cd35e2` by mocking JSDOM `clientWidth` so the drag actually activates.

packages/dialtone-vue/components/resizable/composables/useResizableDrag.ts
hynes-dialpad: Added `resizable_drag.test.js` — covers pointerdown → resizing class, disabled handle rejection, touch events, double-click reset, and handle indicator. 6 tests, all using the correct PointerEvent API.

packages/dialtone-vue/components/resizable/resizable.test.js
hynes-dialpad: Added in `resizable.test.js` — 'Event emissions' describe block covers:
- `panel-collapse` via `collapsePanel()` with payload verification
- `panel-resize` via `resizePanel()`
- `resize-start` and `resize-end` existence in emits definition
- All four event types registered

Peek events (`panel-peek-start`, `panel-peek-end`) were removed — peek feature was cut from this PR as Beacon-specific.

packages/dialtone-vue/components/resizable/resizable_handle.vue
hynes-dialpad: Fixed in `8448e91` — aria-label now uses `messages.handleAriaLabel` template with `{before}` and `{after}` placeholders. Pass `:messages="{ handleAriaLabel: 'Poignée entre {before} et {after}' }"` for i18n.

packages/dialtone-vue/components/resizable/composables/useResizableKeyboard.ts
hynes-dialpad: Fixed — all keyboard announcements are now i18n-overridable via the `messages` prop. Keys include `resizeAnnouncement`, `collapseAnnouncement`, `expandAnnouncement`, `resetAnnouncement`, `minSizeAnnouncement`, `maxSizeAnnouncement`, and `ariaValueText`. Each uses named placeholders (`{panelId}`, `{pixels}`, etc.) so translations can reorder freely.

hynes-dialpad:
Fixed — the keyboard system was completely rewritten to use the W3C separator pattern (`025912819`). The old edit-mode keyboard had a bug where it used container-relative coordinates incorrectly for handles after the first panel. The new implementation:

1. Each handle is independently focusable (tabindex="0")
2. Arrow keys resize the adjacent panel pair relative to the focused handle
3. Each handle resolves its own beforePanelId/afterPanelId from DOM order
4. No global state shared between handles

The three-panel keyboard issue should be resolved. Please re-test on the deploy preview.

hynes-dialpad:
Removed — the peek feature was cut from this PR in `bdd2aff`. It was a Beacon-specific feature with usability issues (as you found — the trigger was clipped inside a zero-width collapsed panel). 

The core component now focuses on: resize, collapse/expand, keyboard a11y, persistence, and constraints. Peek overlay can be built as a Beacon extension later if needed.

packages/dialtone-vue/common/composables/useDOMCache.ts
hynes-dialpad: Agreed — deleted the entire `useDOMCache` module (~280 lines) in `c949be04e`. `useResizableOffset` now uses a plain `document.querySelector()` call. The `common/composables/` directory and its barrel export were also removed.

packages/dialtone-vue/components/resizable/composables/useResizableKeyboard.ts
hynes-dialpad: Fixed — keyboard now follows the same discipline as drag. During key repeat, it writes inline styles only. On key release / blur, it commits the final sizes to `savedState` via the `onResize` callback, which triggers `computeLayout` to reconcile. No more direct `panel.pixelSize` mutations during keyboard interaction.

packages/dialtone-vue/components/resizable/composables/useResizablePanelControls.ts
hynes-dialpad: Done — `handleViewportResize` and all its helpers (~120 lines) were deleted. When `containerSize` changes, `computeLayout` reruns automatically via the reactive dependency chain: ResizeObserver → `containerSize.value` → `layout` computed → `syncedPanels`. Panels with `manualTargetRatio` scale proportionally (computeLayout line 233: `rawWidth = savedRatio * containerSize`). Auto-collapse is triggered from a watcher on `syncedPanels` changes.

packages/dialtone-vue/components/resizable/composables/useResizableCalculations.ts
hynes-dialpad: Deleted — `useResizablePanelSizing` and all its internal helpers removed in `b8863a8`. The function was ~230 lines of duplicated constraint logic that `computeLayout` already handles.

packages/dialtone-vue/components/resizable/composables/useResizableCalculations.ts
hynes-dialpad: Deleted — `getCoordinate` removed from the return object and the function itself in `b8863a8`.

packages/dialtone-vue/components/resizable/composables/useResizableCalculations.ts
hynes-dialpad: Deleted in `b8863a8`.

packages/dialtone-vue/components/resizable/composables/useResizableCalculations.ts
hynes-dialpad: Deleted in `b8863a8`. This one also had the anti-pattern of writing to `panel.pixelSize` directly — exactly the mutation discipline issue you flagged.

packages/dialtone-vue/components/resizable/composables/useResizableCalculations.ts
hynes-dialpad: Agreed — `useResizablePanelSizing` was deleted entirely in `b8863a8`. All constraint enforcement now lives in `computeLayout` via the `constraintResolver` module.

packages/dialtone-vue/components/resizable/composables/computeLayout.ts
hynes-dialpad: This was the most important feedback on the PR — thank you. All 15 direct mutations across 3 files have been eliminated. The architecture is now:

```
User Action → write to savedState → computeLayout reruns → positions update → Vue renders
```

**What changed:**
- `savedState.value` is the ONLY writable state for panel sizes. A new `updateSavedPanel()` helper handles targeted updates.
- `computeLayout` is the ONLY place that calculates positions — no more duplicate constraint logic in `handleViewportResize` or `useResizablePanelSizing` (both deleted).
- Drag uses shadow DOM during mousemove (inline styles for performance), commits to `savedState` on mouseup via `commitPanelSize()`.
- Keyboard follows the same shadow → commit pattern.
- Storage restore writes to `savedState`, which triggers `computeLayout` to validate and reconcile against the current container size.
- `handleViewportResize` (~120 lines) deleted — `computeLayout` already handles viewport scaling via `manualTargetRatio` (line 233: `rawWidth = savedRatio * containerSize`).

Key commits: `b8863a8` (Phase 1-4 refactor), `c949be04e` (simplify + consolidate), `6b63907` (final simplify pass).

apps/dialtone-documentation/docs/.vuepress/exampleComponents/ExampleResizable.vue
hynes-dialpad: Fixed in `70c2573` — example components updated to use `<dt-stack>` where appropriate.

apps/dialtone-documentation/docs/components/resizable.md
hynes-dialpad: Applied in `70c2573` — changed to `status: beta`.

apps/dialtone-documentation/docs/components/resizable.md
hynes-dialpad: Attempted in `70c2573` but reverted to fenced code blocks in `ffa258d` — `<code-example-tabs>` was parsing Vue template tags (`<dt-resizable-panel>`) as real DOM instead of code strings. The fenced blocks work correctly. Happy to revisit when this lands on `next` with the new `<code-example>` component from PR #1147.

packages/dialtone-vue/components/resizable/resizable_offset.test.js
hynes-dialpad: Moot — the `useDOMCache` composable (which needed MutationObserver) was deleted entirely in `c949be04e`. `useResizableOffset` now uses a plain `document.querySelector()` call, so no MutationObserver mock is needed in the offset tests.

apps/dialtone-documentation/docs/components/resizable.md
francisrupert: Just as well, I'd recently merged https://github.com/dialpad/dialtone/pull/1166 to `next` to use fenced blocks with some syntactic sugar instead of `<code-example>`, e.g...

````
```vue demo
<dt-button kind="muted" :size="200">...</dt-button>
```
````

All `next` component pages use that now, and currently have a PR up for CSS Utilities, https://github.com/dialpad/dialtone/pull/1167

**Comment by francisrupert** [comment]
@hynes-dialpad Great updates. I've gone through and see no further issues. I've stress-tested the interaction nuances and it feels bulletproof and snappy. 

Deferring approval to @braddialpad.

File: packages/dialtone-vue/components/resizable/composables/useResizablePanelControls.ts [hunk 32]
```diff
@@ -0,0 +1,556 @@
...
+    manualPanels.forEach(p => newSizes.set(p.id, p.pixelSize));
+    includeUnchangedPanels(allPanels, newSizes);
+  } else {
+    // Default: proportional
+    applyProportionalAllocation(availablePanels, newPanelSize, newSizes);
+    includeUnchangedPanels(allPanels, newSizes);
+  }
+
+  return newSizes;
+}
```
braddialpad: After the recent changes, I believe this is no longer used.

packages/dialtone-vue/components/resizable/resizable.vue
```diff
@@ -0,0 +1,284 @@
...
+const registerPanel = (config) => group.registerPanel(config);
+const unregisterPanel = (id) => group.unregisterPanel(id);
+const saveToStorage = (panels) => group.saveCurrentLayout(panels);
+
+function commitPanelSize (panelId, pixels) {
+  const rounded = Math.round(pixels);
+  const cSize = group.containerSize.value;
+  const ratio = cSize > 0 ? rounded / cSize : undefined;
+  group.updateSavedPanel(panelId, { pixelSize: rounded, manualTargetRatio: ratio });
+}
```
braddialpad: function duplicated in `useResizablePanelControls.ts`. Pretty minor, but could cause one to get out of sync if changes are ever made to it.

packages/dialtone-vue/components/resizable/composables/useResizablePanelControls.ts
```diff
@@ -0,0 +1,556 @@
...
+  }
+
+  // ---- Panel Operations ----
+
+  function commitPanelSize(panelId: string, pixels: number): void {
+    const rounded = Math.round(pixels);
+    const cSize = containerSize.value;
+    const ratio = cSize > 0 ? rounded / cSize : undefined;
+    updateSavedPanel(panelId, { pixelSize: rounded, manualTargetRatio: ratio });
+  }
```
braddialpad, line 206: duplicate I earlier mentioned.

packages/dialtone-vue/components/resizable/composables/useResizableKeyboard.ts
```diff
@@ -0,0 +1,380 @@
...
+
+    afterEl.style.insetInlineStart = `${cursorPos}px`;
+    afterEl.style.inlineSize = '';
+
+    if (handleEl) {
+      handleEl.style.insetInlineStart = `${Math.max(0, cursorPos - 2)}px`;
+    }
+  }
+
+  function applyResize(
```
braddialpad, line 180: A fairly minor issue here, I'll leave it up to you if you think it's worth fixing. I'll paste Claude's summary of it because I think it explains it clearly.

> Drag path (useResizableDrag.ts):                                                                                                                                                                                                                                
>   - During move (line 311–315): writes inline styles to DOM only. Reactive state untouched.                                                                                                                                                                       
>   - On mouseup, commitDrag (line 330–332): clears inline styles first, then calls onDragEnd (line 339) which commits to reactive state.                                                                                                                           
>                                                                                                                                                                                                                                                                   
>   Two clearly separated moments. Vue never sees competing style sources.                                                                                                                                                                                          
>                                                                                                                                                                                                                                                                   
>   Keyboard path (useResizableKeyboard.ts), inside applyResize (line 167–203):                                                                                                                                                                                     
>   - Lines 180–195: writes inline styles to DOM for immediate visual feedback.                                                                                                                                                                                     
>   - Lines 197–202: immediately calls onResize(...) in the same function, which commits to reactive state.                                                                                                                                                         
>                                                                                                                                                                                                                                                                   
>   Both happen synchronously in the same call. Vue then schedules a re-render which will overwrite the inline styles with the computed layout values in the next frame.                                                                                            
>                                                                                                                                                                                                                                                                   
>   Why it usually works fine: Vue's async re-render is fast enough that the inline styles get replaced cleanly before the next paint in most cases. The visual result is correct.                                                                                  
>                                                                                                                                                                                                                                                                   
>   The actual risk: if anything reads DOM positions (e.g. element.style.insetInlineStart) immediately after a keyboard resize — during the window between applyResize returning and Vue's re-render completing — it would get the keyboard-written inline value,   
>   not the layout-computed value. Any code that introspects element positions for further calculations could act on stale data in that window.                                                                                                                     
>                                                                                                                                                                                                                                                                   
>   The fix would mirror the drag approach: in applyResize, after calling onResize, clear the inline styles immediately rather than leaving Vue to do it a frame later. That way the DOM is always exclusively owned by one system at a time.                       
>                                                                                                                                                                                                                                                                   
>   It's low severity — it doesn't cause visual glitches in normal use — but the inconsistency is a latent bug surface if the component grows.

braddialpad (approved):
Looks much cleaner, thanks for the changes.

Approving, with a couple cleanup items

packages/dialtone-vue/components/resizable/composables/useResizablePanelControls.ts
hynes-dialpad: Removed in `9e6f7fe` — deleted `allocateSpaceOnPanelOpen`, `applyProportionalAllocation`, `includeUnchangedPanels` (~80 lines), the `SpaceAllocationStrategy` type, barrel export, and associated tests.

packages/dialtone-vue/components/resizable/resizable.vue
hynes-dialpad: Fixed in `9e6f7fe` — removed the duplicate from `resizable.vue`. The single definition lives in `useResizablePanelControls`, now exported from its return object. `resizable.vue` destructures it: `const { commitPanelSize, ... } = useResizablePanelControls({...})`.

packages/dialtone-vue/components/resizable/composables/useResizablePanelControls.ts
hynes-dialpad: Fixed — single definition, no duplicate. See above.

packages/dialtone-vue/components/resizable/composables/useResizableKeyboard.ts
hynes-dialpad: Fixed in `9e6f7fe` — `applyResize` now clears inline styles immediately after calling `onResize()`, matching the drag path's discipline. Added a `clearInlineStyles()` helper that zeroes out `insetInlineStart`, `insetInlineEnd`, and `inlineSize` on both panels and the handle. Vue exclusively owns the DOM from the moment `onResize` returns.

PR #1165, opened 2026-04-02
docs(playground): DLT-3242 segmented control for compact enum props

francisrupert:
small-lakers

- [x] Feature

## :book: Jira Ticket

[DLT-3242](https://dialpad.atlassian.net/browse/DLT-3242)

## :book: Description

Enum props with few, short options (e.g. `size: [100, 200, 300, 400, 500]`) now render as `DtSegmentedControl` instead of `DtDropdown` in the combinator option bar.

Heuristic in `getControlByMemberType()`: ≤6 values and all labels ≤4 chars → segmented; otherwise dropdown.

**New file:** `control_segmented.vue` — wraps `DtSegmentedControl`, same prop contract as `control_selection.vue`.

**Modified:** `control.js` — `segmented` entry in `controlMap`, `shouldUseSegmented()` heuristic.

## :bulb: Context

Dropdown controls require two interactions (open + select) for props that have only 3–5 short options. Segmented controls show all options inline — single click to change.

## :pencil: Checklist

## :crystal_ball: Next Steps

- Validate threshold tuning against all Dialtone component enum props
- Consider variant config override (`controlType`) if heuristic needs manual correction (deferred — current escape hatch is the control-type switcher)

[DLT-3242]: https://dialpad.atlassian.net/browse/DLT-3242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

https://github.com/user-attachments/assets/77bd8fd4-2efa-4e19-a606-a8994af08143

### DIFF
--- a/packages/combinator/src/components/controls/control_segmented.vue
+++ b/packages/combinator/src/components/controls/control_segmented.vue
@@ -1,0 +1,81 @@
+<template>
+  <dt-text
+    kind="label"
+    :size="100"
+    tone="secondary"
+    class="d-input__label-text d-c-default"
+  >
+    <slot />
+  </dt-text>
+  <dt-segmented-control
+    :model-value="String(value)"
+    :size="100"
+    :disabled="disabled"
+    @change="onInput"
+  >
+    <dt-segmented-control-item
+      v-for="option in options"
+      :key="option.value"
+      :value="String(option.value)"
+    >
+      {{ option.label }}
+    </dt-segmented-control-item>
+  </dt-segmented-control>
+</template>
+
+<script setup>
+import { DtSegmentedControl, DtSegmentedControlItem, DtText } from '@dialpad/dialtone-vue';
+
+import { VALUE_UPDATE_EVENT } from '@/src/lib/constants';
+import { computed } from 'vue';
+
+const props = defineProps({
+  value: {
+    type: undefined,
+    required: true,
+  },
+  validValues: {
+    type: Array,
+    default: undefined,
+  },
+  disabled: {
+    type: Boolean,
+    default: false,
+  },
+  generateLabel: {
+    type: Function,
+    default: (value) => value.toString(),
+  },
+});
+
+const emit = defineEmits([VALUE_UPDATE_EVENT]);
+
+const options = computed(() => {
+  return props.validValues?.map(v => ({
+    value: v,
+    label: props.generateLabel(v),
+  })) ?? [];
+});
+
+// DtSegmentedControl coerces values to strings — map back to original types on selection
+const valueMap = computed(() => {
+  const map = {};
+  props.validValues?.forEach(v => {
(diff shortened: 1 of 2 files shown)


packages/combinator/src/components/controls/control_segmented.vue
```diff
@@ -0,0 +1,81 @@
...
+  },
+});
+
+const emit = defineEmits([VALUE_UPDATE_EVENT]);
+
+const options = computed(() => {
+  return props.validValues?.map(v => ({
+    value: v,
+    label: props.generateLabel(v),
+  })) ?? [];
```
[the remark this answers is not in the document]
francisrupert: The null concern doesn't apply: null is synthetic (not in `member.values`), props hitting the heuristic (e.g. `size`) are effectively required numeric scales, and no segment selected _**is**_  the default.

packages/combinator/src/components/controls/control_segmented.vue
```diff
@@ -0,0 +1,81 @@
...
+});
+
+// DtSegmentedControl coerces values to strings — map back to original types on selection
+const valueMap = computed(() => {
+  const map = {};
+  props.validValues?.forEach(v => {
+    map[String(v)] = v;
+  });
```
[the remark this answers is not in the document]
francisrupert: The concern is that `1` (number) and `'1'` (string) would collide. This requires a Dialtone component to have _mixed_-type values with colliding string representations. That doesn't exist — they're either all numbers (`[100, 200, 300, 400, 500]`) or all strings. A component with both `1` and `'1'` as valid values for the same prop would be a broken API.

packages/combinator/src/lib/control.js
```diff
@@ -110,6 +119,11 @@ export function getControlByMemberType (type, args) {
   }
 }
 
+function shouldUseSegmented (values) {
+  return values.length <= MAX_SEGMENTED_COUNT &&
+    values.every(v => String(v).length <= MAX_SEGMENTED_LABEL_LENGTH);
+}
```
braddialpad, line 125: Is it intended to be checking the "values" for length here as opposed to the "label" which is what actually renders on screen?

francisrupert: Yes, that's kind of the point with this PR. I'm selectively only using Segmented Controls if they're within a length threshold and/or the number of values available are few enough to fit them all in a Segmented Control. Otherwise, they display a dropdown.

In this context, the values are what get displayed as labels, so checking the value length is checking the label length.

braddialpad: ah okay, I thought it was possible to have label be different than value

braddialpad (approved):
Looks good!

PR #1178, opened 2026-04-07
feat(dialtone-docs): DLT-3109 add markdown-to-JSON generator for AI docs

belumontoya:
# feat(dialtone-docs): DLT-3109 add markdown-to-JSON generator for AI docs

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-3109

## :book: Description

Adds the markdown-to-JSON build pipeline for the `dialtone-docs` package:

- **`src/generators/build-ai-docs.mjs`** — Reads all markdown files under `src/content/`, parses YAML frontmatter (type, category, keywords, ai_summary), strips markdown syntax, and compiles everything into `dist/ai-docs.json` — a flat JSON array of document entries for AI consumption.
- **`src/utils/strip-markdown.mjs`** — Utility that strips frontmatter, code blocks, HTML, links, headings, emphasis, and other markdown syntax to produce searchable plain text.
- **`package.json` / `project.json`** — Added `build` script and NX target so `pnpm nx run dialtone-docs:build` triggers the generator.
- **`tests/tests/build-output.test.js`** — 11 tests validating the JSON output schema (required fields, types, no markdown artifacts in content, file path integrity, no duplicate IDs).
- **`tests/tests/strip-markdown.test.js`** — Unit tests for the strip-markdown utility (headings, code blocks, links, emphasis, frontmatter removal).
- **`tests/helpers/markdownParser.js`** — Refactored to import `stripMarkdown`/`stripFrontmatter` from the new utility instead of bundling its own copy.

## :bulb: Context

The `dialtone-docs` package provides AI-discoverable documentation for the Dialtone monorepo. This PR adds Milestone 3: the build step that compiles markdown content into a structured JSON file (`ai-docs.json`). This JSON output will serve as the data source for MCP server and CLI search tools, enabling AI agents to search the entire documentation site programmatically.

Each JSON entry includes: `id`, `title`, `type`, `category`, `keywords`, `summary`, `content` (plain text), `filePath`, `lastUpdated`, and `relatedPackages`.

## :pencil: Checklist

## :crystal_ball: Next Steps

- Refactor existing test suite (consolidate 6 test files down to 3, remove hardcoded content assertions that do not scale)
- Integrate `ai-docs.json` into the MCP server and CLI as a search data source


packages/dialtone-docs/tests/tests/build-output.test.js
```diff
@@ -0,0 +1,92 @@
...
+        expect(typeof kw, `"${doc.id}" keyword "${kw}" not a string`).toBe('string');
+      }
+    }
+  });
+
+  test('type field uses allowed values', () => {
+    for (const doc of docs) {
+      expect(doc.type, `"${doc.id}" type is null`).not.toBeNull();
+      expect(ALLOWED_TYPES, `"${doc.id}" invalid type "${doc.type}"`).toContain(doc.type);
+    }
```
braddialpad: For all tests where we are looping through arrays like this, we should be using test.each instead of a for loop.

A single failure will mask all subsequent ones when doing it the current way.

packages/dialtone-docs/src/utils/strip-markdown.mjs
```diff
@@ -0,0 +1,64 @@
...
+  codeBlockFenced: /^`{3,}([^\n]*)\n([\s\S]*?)^`{3,}/gm,
+  codeInline: /`([^`]+)`/g,
+  link: /\[([^\]]*)\]\(([^)]+)\)/g,
+  image: /!\[([^\]]*)\]\(([^)]+)\)/g,
+  htmlComment: /<!--[\s\S]*?-->/g,
+  htmlTag: /<[^>]+>/g,
+  blockquote: /^>\s?.*/gm,
+  horizontalRule: /^(?:[-*_]){3,}\s*$/gm,
+  emphasis: /[*_]{1,2}([^*_]+)[*_]{1,2}/g,
+};
```
braddialpad, line 14: This PATTERNS is a duplicate of the one in `markdownParser.js`. Is that intentional? They could get out of sync in future changes if they are both being used.

packages/dialtone-docs/tests/tests/build-output.test.js
```diff
@@ -0,0 +1,92 @@
...
+      expect(Array.isArray(doc.keywords), `"${doc.id}" keywords not an array`).toBe(true);
+      for (const kw of doc.keywords) {
+        expect(typeof kw, `"${doc.id}" keyword "${kw}" not a string`).toBe('string');
+      }
+    }
+  });
+
+  test('type field uses allowed values', () => {
+    for (const doc of docs) {
+      expect(doc.type, `"${doc.id}" type is null`).not.toBeNull();
```
braddialpad: nit (optional): this first assertion is probably not necessary

braddialpad (commented):
Looks really good, perhaps my only concern would be the large amount of confusing regex to parse the markdown, however it is necessary for this change and also easier to understand in the age of AI.

Couple of small comments, nothing major.

packages/dialtone-docs/src/utils/strip-markdown.mjs
belumontoya: Nope, nice spotted, I'm actually thinking now to replace this with remove-markdown package to handle this instead of us doing it manually

packages/dialtone-docs/tests/tests/build-output.test.js
belumontoya: Agree, but one thing worth noting. docs is built in beforeAll, so test.each can't get it at test definition time. I could build synchronously at module scope or keep the loop, but use soft assertions. Any preference? Or a better idea?

braddialpad: I guess we can just leave docs alone, and do something like this at least:

```
test.each(REQUIRED_FIELDS)('all entries have field "%s"', (field) => {
    for (const doc of docs) {
      expect(doc).toHaveProperty(field);
    }
  });
```

Moving the file reads outside of test execution has worse side effects than the current method IMO

braddialpad (approved):
Thanks!

PR #1136, opened 2026-03-18
chore: NO-JIRA upgrade vite to version 8

braddialpad:
Upgrades vite and any related dependencies to the latest version.

PR #1188, opened 2026-04-08
chore: NO-JIRA fix Storybook dev mode JSX parse error after Vite 8 upgrade

francisrupert:
30892636f055bbf4a84bed6fd3d23bb3

- [x] Build system

## :book: Jira Ticket

NO-JIRA

## :book: Description

Add `@vitejs/plugin-react` to Storybook's Vite config to fix JSX parse errors in dev mode.

After the Vite 8 upgrade (#1136), Storybook local preview failed to work. `storybook dev` fails with `Parse error @:250:28` in `storybook:external-globals-plugin`. 

The root cause: `es-module-lexer` (used by Storybook's external-globals-plugin) cannot parse JSX syntax. In Vite 5-7, esbuild transformed `.jsx` files before this plugin ran. In Vite 8, the OXC transform doesn't pre-process `.jsx` in dev mode for Vue projects, so raw JSX reaches the plugin.

**Fix:** add `@vitejs/plugin-react` with `include: /\.jsx$/` and `jsxRuntime: 'classic'` to ensure JSX files are transformed before the external-globals-plugin sees them. Scoped to `.jsx` files only to avoid conflicts with Vue's SFC compiler.

> [!TIP]
> `storybook build` (CI) was unaffected — the issue is **dev-mode-only**.

## :bulb: Context

Storybook uses React for its docs addon (`preview.jsx`, `DialtoneDocsPage.jsx`) even in Vue projects. Without explicit React JSX handling, these files break in Vite 8 dev mode.

## :pencil: Checklist

For all PRs:

### DIFF
--- a/packages/dialtone-vue/.storybook/main.js
+++ b/packages/dialtone-vue/.storybook/main.js
@@ -2,6 +2,7 @@
 import { createRequire } from 'node:module';
 import { dirname, join } from 'node:path';
 import { mergeConfig } from 'vite';
+import react from '@vitejs/plugin-react';
 
 const require = createRequire(import.meta.url);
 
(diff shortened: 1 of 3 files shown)


ninarepetto (approved):

braddialpad (approved):
The other solution would be to get rid of the jsx and use `React.createElement` since we use very little actual jsx. But that would make it less readable, so I think this solution is fine.

PR #1207, opened 2026-04-17
docs(modal): NO-JIRA fix code examples

jeverhart-dialpad:
# docs(modal): NO-JIRA code example update

Obligatory GIF

- [ ] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [x] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Description
This PR aligns the code examples on the modal documentation page with UI examples shown on the page. The modal page now also auto-generates the HTML examples to reduce manual updates to that in the future.

### Items addressed

**Button order and text**
- Swapped Confirm and Cancel button order in all code examples to match the live demo (Confirm first)
- Fixed button text in HTML examples from "Save changes" to "Confirm"

**Code example accuracy**
- Replaced all static hardcoded HTML code blocks with auto-generated HTML captured from live Vue component refs, so examples stay accurate as the component changes
- Removed undefined variable `:kind="secondaryButtonKind"` from code examples
- Removed non-existent prop `:showFooter="true"` from the Fixed Header example
- Fixed `:bannerKind` (camelCase) to `banner-kind` (kebab-case) in the Has Banner example
- Fixed `kind` not being forwarded to `<dt-modal>` in ExampleModal (was only passed to the confirm button)

**ExampleModal improvements**
- Added `initiallyOpen` prop and `mounted()` hook to support the HTML capture pattern
- Removed unused `fixedHeaderFooter` prop (DtModal handles its own default)
- Added `class="d-mt0"` to the trigger button to prevent unwanted spacing from the `d-stack8` layout context
- Made modal-first, button-last template order consistent across all examples

**Accessibility and grammar**
- Fixed "announce its to accessible technology" → "announce it to assistive technology"
- Fixed "The value of aria-labelledby is to the `id` value" → "should match the `id` of its heading element"
- Added missing trigger button to the Custom Header `vueCode` example

### Note about revised HTML examples for modal
In order to get the HTML automatically, an extra example modal was added for each entry inside of a hidden `<div>` set to`inert`  to capture the proper HTML markup when the page loads. It renders an `<example-modal initially-open />` so that `DtLazyShow` (which only initializes the modal DOM when show=true) actually renders the full modal markup. The inert attribute prevents that hidden open modal from trapping focus or intercepting keyboard/pointer events on the page. The `d-h0 d-of-hidden d-vi-hidden d-pe-none` classes keep it visually hidden and out of the layout. Not sure if this is overkill, but it does seem to work.

## :bulb: Context
As reported in a Dialtone channel, the code examples provided on the modal documentation page did not match what was actually being displayed -- the buttons were in the wrong order.

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

There are a few margin fixes in this version because the modal component doesn't append to the body or an ancestor by default. This is something we may want to revisit. In the example of the docs page, it is grouped with a button inside of a stack, and so either the button or the modal end up inheriting a margin.

Another item to consider is whether we should use a stack in the example instead of margins on the buttons. Not sure if that was a purposeful decision.

### Before

Screenshot 2026-04-16 at 8 25 23 PM

### After

Screenshot 2026-04-16 at 8 27 19 PM

## :link: Sources

Corrects modal docs: fixes example button order, adds inert wrapper divs and an initiallyOpen prop to render full modal markup without trapping focus, and auto-generates HTML examples from Vue refs to reduce manual updates; also tightens copy and standardizes button styling.

### DIFF
--- a/apps/dialtone-documentation/docs/.vuepress/exampleComponents/ExampleModal.vue
+++ b/apps/dialtone-documentation/docs/.vuepress/exampleComponents/ExampleModal.vue
@@ -5,7 +5,6 @@
     :kind="kind"
     :banner-title="bannerTitle"
     :banner-kind="bannerKind"
-    :fixed-header-footer="fixedHeaderFooter"
     :size="size"
     :copy="copy"
     @update:show="isOpen = $event"
@@ -63,11 +62,6 @@ export default {
     bannerTitle: {
       type: String,
       default: '',
-    },
-
-    fixedHeaderFooter: {
-      type: Boolean,
-      default: true,
     },
 
     size: {
--- a/apps/dialtone-documentation/docs/.vuepress/exampleComponents/ExampleModal.vue
+++ b/apps/dialtone-documentation/docs/.vuepress/exampleComponents/ExampleModal.vue
@@ -1,9 +1,4 @@
 <template>
-  <dt-button
-    @click="isOpen = !isOpen"
-  >
-    Click to open
-  </dt-button>
   <dt-modal
     title="Example title"
     :show="isOpen"
@@ -13,7 +8,6 @@
     :fixed-header-footer="fixedHeaderFooter"
     :size="size"
     :copy="copy"
-    modal-class="d-mt0"
     @update:show="isOpen = $event"
   >
     <template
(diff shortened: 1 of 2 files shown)


ninarepetto (approved):

braddialpad (approved):
Thanks!

PR #1216, opened 2026-04-21
feat(components): DLT-3192 establish the slotclass convention rule

iropolo:
# Establish the slotclass convention rule

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-3192

## :book: Description

We are creating the convention rules in `Claude` for slotClass usage.


braddialpad (approved):
Thanks!

PR #1222, opened 2026-04-24
fix(avatar, presence): NO-JIRA update recipe class override to include uikits version

jeverhart-dialpad:
# fix(avatar, presence): NO-JIRA update recipe class override to include uikits version

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [ ] Refactor

## :book: Jira Ticket
NO-JIRA

## :book: Description

Updated the uses of `d-recipe-leftbar-row` in `avatar.less` and `presence.less` to include the class name that ui-kits uses since the recipes have been migrated to ui-kits.  I tested this locally using `pnpm pack`.

This fixes the immediate issue, but we may also want to make this update in the ui-kits repository for a more permanent fix. These classes were originally left in for some backwards-compatibility I believe, and we may not even need the ones in the presence.less file. However, it is worth considering how these rows will change when the new version of dialtone is introduced and whether that will be a better time to update it fully and remove the old border styles. Open to guidance there.

## :bulb: Context
This should fix the 'selected' left sidebar row state. Currently the border color is slightly off around the presence.

Screenshot 2026-04-24 at 8 37 14 AM

## :pencil: Checklist

For all PRs:

For all CSS changes:

Screenshot 2026-04-28 at 10 45 33 AM

### DIFF
--- a/packages/dialtone-css/lib/build/less/components/avatar.less
+++ b/packages/dialtone-css/lib/build/less/components/avatar.less
@@ -125,7 +125,9 @@
     box-shadow: 0 0 0 var(--dt-size-200) var(--avatar-count-color-shadow);
 
     .d-recipe-leftbar-row--selected &,
-    .d-recipe-leftbar-row__primary:hover & {
+    .d-recipe-leftbar-row__primary:hover &,
+    .dp-general-row--selected &,
+    .dp-general-row__primary:hover & {
       --avatar-count-color-shadow: var(--dt-shell-action-color-background-primary-hover);
     }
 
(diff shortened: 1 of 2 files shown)


braddialpad (changes requested):
I think these should instead just be transferred directly to ui-kits css if they don't already exist there. We shouldn't have any ui-kits related css in core dialtone period.

Should be a good task for your first contribution to ui-kits 😄.

jeverhart-dialpad:
> I think these should instead just be transferred directly to ui-kits css if they don't already exist there. We shouldn't have any ui-kits related css in core dialtone period.
> 
> Should be a good task for your first contribution to ui-kits 😄.

👍  Ok, see you over there.

PR #1241, opened 2026-04-30
fix(popover, collapsible): DP-185811 cancel transitions on unmount

braddialpad:

- [x] Fix

## :book: Jira Ticket

[DP-185811](https://dialpad.atlassian.net/browse/DP-185811)

## :book: Description

Fixes a race condition where a CSS leave transition (fade/collapse) could complete after `DtPopover` or `DtCollapsible` was already unmounted, causing Vue to attempt updates on a dead component instance.

## :bulb: Context

Reported from the product side — implementing components were unmounting while a popover/collapsible leave animation was in progress. The `after-leave` hook then fired on the already-torn-down component, producing warnings and potential errors about updating unmounted component instances.

## :pencil: Checklist

For all PRs:

For all Vue changes:

[DP-185811]: https://dialpad.atlassian.net/browse/DP-185811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dialtone-vue/components/collapsible/collapsible.vue
+++ b/packages/dialtone-vue/components/collapsible/collapsible.vue
@@ -249,6 +249,8 @@ export default {
 
   beforeUnmount () {
     this._isUnmounting = true;
+    // Prevent transition callbacks from calling into dead lifecycle methods
+    // after this component is torn down (DP-185811).
   },
 
   methods: {
--- a/packages/dialtone-vue/components/collapsible/collapsible.vue
+++ b/packages/dialtone-vue/components/collapsible/collapsible.vue
@@ -247,8 +247,13 @@ export default {
     this.validateProperAnchor();
   },
 
+  beforeUnmount () {
+    this._isUnmounting = true;
+  },
+
   methods: {
     onLeaveTransitionComplete () {
+      if (this._isUnmounting) return;
       this.$emit('opened', false);
       if (this.open !== null) {
         this.$emit('update:open', false);
@@ -256,6 +261,7 @@ export default {
     },
 
     onEnterTransitionComplete () {
+      if (this._isUnmounting) return;
       this.$emit('opened', true, this.$refs.content);
       if (this.open !== null) {
         this.$emit('update:open', true);
--- a/packages/dialtone-vue/components/popover/popover.test.js
+++ b/packages/dialtone-vue/components/popover/popover.test.js
@@ -305,6 +305,26 @@ describe('DtPopover Tests', () => {
     });
   });
 
+  describe('When component is unmounting', () => {
+    it('sets transition: none on content element to cancel in-flight transitions', () => {
+      const contentEl = wrapper.vm.popoverContentEl;
+      wrapper.unmount();
+      expect(contentEl.style.transition).toBe('none');
+    });
+
+    it('does not emit "opened" when leave transition completes', async () => {
+      wrapper.vm._isUnmounting = true;
+      await wrapper.vm.onLeaveTransitionComplete();
+      expect(wrapper.emitted('opened')).toBeUndefined();
+    });
+
+    it('does not emit "opened" when enter transition completes', async () => {
+      wrapper.vm._isUnmounting = true;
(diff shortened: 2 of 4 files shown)


packages/dialtone-vue/components/popover/popover.test.js
```diff
@@ -305,6 +305,36 @@ describe('DtPopover Tests', () => {
...
+    it('sets transition: none on content element to cancel in-flight transitions', () => {
+      const contentEl = wrapper.vm.popoverContentEl;
+      wrapper.unmount();
+      expect(contentEl.style.transition).toBe('none');
+    });
+
+    it('does not emit "opened" when leave transition completes, but does emit when not unmounting', async () => {
+      wrapper.vm._isUnmounting = true;
+      await wrapper.vm.onLeaveTransitionComplete();
+      expect(wrapper.emitted('opened')).toBeUndefined();
```
belumontoya, line 321: Testing rule §6 (one assertion per test)
[nit / Dialtone testing rule] These two new test blocks each cover both the guarded path and the happy path with 2-3 expects. Per .claude/rules/code-review.md §6 ("Do the tests have one assertion per test?"), Consider splitting each into two it blocks — one for the unmounting case, one for the normal case. Easier to read which path failed when a regression hits.

packages/dialtone-vue/components/popover/popover.test.js
```diff
@@ -305,6 +305,36 @@ describe('DtPopover Tests', () => {
...
   });
 
+  describe('When component is unmounting', () => {
+    it('sets transition: none on content element to cancel in-flight transitions', () => {
+      const contentEl = wrapper.vm.popoverContentEl;
+      wrapper.unmount();
+      expect(contentEl.style.transition).toBe('none');
+    });
+
+    it('does not emit "opened" when leave transition completes, but does emit when not unmounting', async () => {
```
belumontoya: "When component is unmounting" describe) — Testing rule §6 (beforeEach to cut repetition)

[nit / Dialtone testing rule] wrapper.vm._isUnmounting = true is set inline in both new tests. The collapsible suite already centralizes this in a beforeEach at the top of its equivalent describe block (collapsible.test.js:148-151) — mirroring that pattern here keeps the two suites consistent and matches .claude/rules/code-review.md §6 ("Do the tests take advantage of beforeEach...").

packages/dialtone-vue/components/collapsible/collapsible.vue
```diff
@@ -247,15 +247,21 @@ export default {
     this.validateProperAnchor();
   },
 
+  beforeUnmount () {
```
belumontoya, line 250: Code Quality rule §2 (comments for non-obvious code)
[nit / Dialtone code-quality rule] popover.vue:716-717 has a comment explaining why beforeUnmount sets the flag and cancels transitions. The same hook here in collapsible has no equivalent comment, so a reader unfamiliar with DP-185811 might think it's dead code and remove it. Mirroring the popover comment would prevent that.

packages/dialtone-vue/components/popover/popover.vue
```diff
@@ -712,6 +712,12 @@ export default {
   },
 
   beforeUnmount () {
+    this._isUnmounting = true;
```
belumontoya, line 715: Reuse rule §1 (consider once a third component lands)
[note / Dialtone reuse rule] The beforeUnmount flag + early-return guard pattern is now in two components. Not worth extracting yet — but per .claude/rules/code-review.md §1 ("Is there any repeated code that should be extracted into a function?"), if a third component (e.g., a future modal/drawer fix) gets the same pattern, it'd be a clean candidate for a useUnmountGuard() composable in the style of packages/dialtone-vue/components/resizable/composables/. Flagging now so it doesn't slip past three.

belumontoya (approved):
no mayor issues from the ones already flagged.

braddialpad:
Thanks! made the fixes

PR #1268, opened 2026-05-13
fix(modal): auto-detect shadow root and teleport dialog into it

ninarepetto:
## Summary

When `DtModal` is rendered inside a shadow root (e.g. in a Web Components / MFE context), `dialog.showModal()` promotes the native dialog to the document top layer, which escapes the shadow root's CSS scope. This causes the modal to render without styles.

### Fix

In `mounted()`, detect if the component is inside a shadow root via `this.$el?.getRootNode()`. If so, set `autoTeleportTarget` to the shadow root and defer `syncDialogState` via `$nextTick` so the teleport renders first.

The `<Teleport>` target falls back to `autoTeleportTarget` when `appendTo` is not set, keeping existing behavior unchanged for non-shadow-root contexts.

### Behavior

- Shadow root context: dialog teleports into the shadow root → CSS scope maintained
- Default (no shadow root, no `appendTo`): Teleport disabled → renders in place (unchanged)
- `appendTo` prop set: explicit prop takes precedence over auto-detection (unchanged)

### Tests

Added 3 tests to `Modal.test.js` covering shadow root detection, teleport into shadow root, and `appendTo` precedence. All 27 tests pass.

### DIFF
--- a/packages/dialtone-vue/components/Modal/Modal.vue
+++ b/packages/dialtone-vue/components/Modal/Modal.vue
@@ -424,7 +424,7 @@ export default {
       if (root instanceof ShadowRoot) {
         this.autoTeleportTarget = root;
         // Defer syncDialogState so the teleport renders inside the shadow root first.
-        if (this.open) this.$nextTick(() => { this.syncDialogState(true); });
+        if (this.open) this.$nextTick(() => { if (this.open) this.syncDialogState(true); });
         return;
       }
     }
--- a/packages/dialtone-vue/components/Modal/Modal.vue
+++ b/packages/dialtone-vue/components/Modal/Modal.vue
@@ -1,7 +1,7 @@
 <template>
   <teleport
-    :disabled="!appendTo"
-    :to="appendTo"
+    :disabled="!appendTo && !autoTeleportTarget"
+    :to="appendTo || autoTeleportTarget"
   >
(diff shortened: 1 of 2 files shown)


fabiangarga (approved):
LGTM, we can try this, few things:
- It only applies to the ones that we don't specify `appendTo`, we need to update `app-admin`
- we need to review the other usages, like if we can apply the same shadow detection at the `initTippyInstance` for a more generic implementation

packages/dialtone-vue/components/Modal/Modal.vue
```diff
@@ -394,6 +394,7 @@ export default {
       MODAL_BANNER_KINDS,
       hasSlotContent,
       i18n: new DialtoneLocalization(),
+      autoTeleportTarget: null,
```
iropolo, line 397: Cleanup `autoTeleportTarget` on a `beforeMount` to `null`

iropolo (approved):

packages/dialtone-vue/components/Modal/Modal.vue
ninarepetto: autoTeleportTarget just holds a reference to the ShadowRoot — a DOM object owned by the host element, not created by the modal. When the component is unmounted, the instance is destroyed and the reference goes away with it. No external listeners are attached, no state outside the component is modified.

PR #1270, opened 2026-05-13
fix(components): no-jira update tree-shaking export

iropolo:
# Update tree-shaking export

- [x] Fix

## :book: Description

This previous adjust https://github.com/dialpad/dialtone/pull/1255/changes/d1c369b379e94dfe57116643b4d9b6093496cc10
Was to tackle the first letter uppercase issue, but .toLowerCase() alone loses the hyphens.

Im updating the way we do it to achieve the expected result:

e.g.:
ComboboxMultiSelect/ComboboxMultiSelect -> combobox-multi-select
Avatar/Avatar -> avatar/avatar
InputGroup/InputGroup -> input-group/input-group
.....

### DIFF
--- a/packages/dialtone-vue/vite.config.js
+++ b/packages/dialtone-vue/vite.config.js
@@ -24,6 +24,7 @@ function _getEntries (pathPrefix, globRegex) {
       .replace(`${pathPrefix}/`, '')
       .replace(/\.(vue|js)/, '')
       .replaceAll('_', '-')
+      .replace(/([a-z0-9])([A-Z])/g, '$1-$2')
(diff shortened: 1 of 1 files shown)


iropolo:
I guess in the future we should have these as PascalCase like:
@dialpad/dialtone/vue3/lib/select-menu ❌
@dialpad/dialtone/vue3/lib/selectMenu ✅

But for now to avoid any breaking change we will still return it as snake-case.
This is specific for tree-shaking because the components exports are already correct on PascalCase (and it was before this refactor)

ninarepetto (approved):

PR #1279, opened 2026-05-18
feat(modal): NO-JIRA apply headerClass to header text element

jeverhart-dialpad:
# feat(modal): NO-JIRA apply headerClass to header text element

Obligatory GIF

- [ ] Fix
- [x] Feature
- [ ] Performance Improvement
- [ ] Refactor

- [ ] Documentation
- [ ] Build system
- [ ] CI
- [ ] Test
- [ ] Other (chore)

## :book: Jira Ticket

NO-JIRA

## :book: Description

Apply the `headerClass` prop on `DtModal` to the element with class `d-modal__header`, even when it is the default output with no slot override. Not 100% sure if this would be considered a fix or a feature.

### What

`headerClass` was declared as a prop on `DtModal` and applied correctly when using the `#header` slot, but was not bound on the `DtText` element used when `headerText` is supplied directly. The class was silently dropped in the common case.

### Why

The `v-if` (slot) and `v-else` (text) branches both render the same `d-modal__header` element — seems like both should accept the same consumer class.

### Changes

- `Modal.vue` — bind `headerClass` on the `DtText` `v-else` branch to match the existing binding on the `#header` slot `div`
- `Modal.test.js` — add test asserting the class is applied when the prop is set

## :bulb: Context

Noticed while using DtModal in Beacon that the `headerClass` wasn't applying as I expected.

However, I could see a case for omitting it if we want to reserve it only for when slots are provided as overrides. I also started this in order to fix the `headerText` in the docs examples, but it looks like that is already covered in https://github.com/dialpad/dialtone/pull/1277, so I think this could just be considered an optional follow up to that PR.

## :pencil: Checklist

For all PRs:

## :crystal_ball: Next Steps

Check other components for similar quirks.

### Before
Showing with `headerClass` prop in use, but no custom header slot.

Screenshot 2026-05-18 at 7 52 02 AM

### After

Screenshot 2026-05-18 at 7 50 52 AM

### DIFF
--- a/packages/dialtone-vue/components/Modal/Modal.test.js
+++ b/packages/dialtone-vue/components/Modal/Modal.test.js
@@ -232,6 +232,16 @@ describe('DtModal Tests', () => {
       expect(banner.classes(bannerClass)).toBe(true);
     });
 
+    it('Should apply header class to the header text element', async () => {
+      const headerClass = 'header-class';
+
+      expect(title.classes(headerClass)).toBe(false);
+
+      await wrapper.setProps({ headerClass });
+
+      expect(title.classes(headerClass)).toBe(true);
+    });
+
     it('Should apply banner kind', async () => {
       await wrapper.setProps({
         open: true,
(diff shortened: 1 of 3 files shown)


belumontoya (approved):
LGTM

PR #1297, opened 2026-05-28
feat(rich-text-editor): DP-190415 implement phone number linking

briandial:

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

[DP-190415](https://dialpad.atlassian.net/browse/DP-190415)

## :book: Description

Adds a `linkPhoneNumbers` prop to `DtRichTextEditor` and a companion `LinkPhoneNumbers` TipTap extension for linking phone numbers in received messages. When a user clicks a linked number the editor emits a `phone-click` event; consuming apps can wire this to a native dialer action.

### How it works

**`linkPhoneNumbers: string[] | null` prop** — pass a list of phone number strings (e.g. from backend `rich_media`) to mark exactly those numbers as clickable links. `null` (default) disables the feature entirely.

Marks are applied once after content is set (and re-applied whenever the prop changes), rather than live as the user types. This is intentional: the use case is read-only received messages where the backend already knows which numbers are present. No regex runs at all — only exact string matches against the provided list are linked.

**`phone-click` event** — emitted with `{ phoneNumber: string }` when a linked number is clicked. The phone number is resolved directly from the ProseMirror document (not the DOM) so it is always correct in test environments.

### Extension design

`LinkPhoneNumbers` is a standalone TipTap `Mark` extension separate from the existing `customLink` / `autolink` extensions. This keeps URL/email autolinking and phone linking as independent concerns with no shared state. The extension only contributes a ProseMirror `handleClick` plugin; `DtRichTextEditor` owns all mark application logic in `_applyPhoneMarks()`.

`_applyPhoneMarks()` clears all existing `LinkPhoneNumbers` marks before re-applying, so stale marks are removed when the prop changes (e.g. `[A]` → `[B]` correctly unlinks `A` and links `B`).

### `autolink.js` robustness fixes

Two edge-case fixes to the existing URL/email autolinker that are independent of the phone feature:

- **Full-doc scan fallback** — when `getChangedRanges` returns nothing (e.g. after a wholesale `setContent`), fall back to scanning the entire document instead of silently skipping.
- **Zero-width range guard** — skip ranges where `from === to` to prevent mark-only steps from deleting existing marks.

Also removes a dead `mounted()` dispatch that set a `dp-force-autolink` meta key that `appendTransaction` never read (the initial full scan already fires via the `!hasInitialized` branch on the first `setContent` transaction).

## :bulb: Context

The `linkPhoneNumbers` prop is intended for message display contexts where the backend has already identified which phone numbers are present in the message body. The exact-match approach avoids regex false positives (e.g. numeric URL path segments) and keeps the feature deterministic.

Known limitation: a phone number that spans a ProseMirror text-node boundary (e.g. partially bold) will not be matched by `_applyPhoneMarks()`. Backend-provided numbers in plain-text messages are not expected to hit this case in practice.

## :pencil: Checklist

For all PRs:

For all Vue changes:

[DP-190415]: https://dialpad.atlassian.net/browse/DP-190415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


belumontoya:
@briandial make sure test pass please

braddialpad:
Hey Brian, I don't think we are actually using the custom-link prop in MessageInput currently. We'll have to make sure everything is still working the same with that enabled. Not necessarily a problem, just a warning that some things with the existing link system have the potential to break.

belumontoya:
he took care of that too https://github.com/dialpad/dialpad-uikits/pull/247

braddialpad:
Cool, I'm just a bit apprehensive because I remember we made and tested this customlink extension before and then went back to the regular link extension because some things didn't work correctly. Maybe Brian's change has fixed that though.

briandial:
> Cool, I'm just a bit apprehensive because I remember we made and tested this customlink extension before and then went back to the regular link extension because some things didn't work correctly. Maybe Brian's change has fixed that though.

This change is for fixing the issue with phone numbers not rendering as links as in regular app without rich text. I tested everything locally. Unfortunately I cannot solve the issue in firespotter and I need to put code here, that implies it can impact in more people. Please help me to understand how I can deliver this and close the ticket.

braddialpad (changes requested):
Okay so the current customLink extension has issues with editing links and such. I think overall the whole customLink thing is not great architecture. (nothing to do with you, this was created like 4 years ago) It conflicts/rewrites the entire built in link extension so we're basically re-writing code that's already in the library to a large extent.

What I'd prefer is to keep the existing "link" active since we know it works well, and then have an additional extension specifically for linking phone numbers `allowLinkPhoneNumber`. Let's leave CustomLink alone and we'll possibly just remove it in the future. Let's re-architect it in this way and then go from there.

braddialpad:
After discussion with @briandial I will be taking this one over due to it's complexity

braddialpad:
What the hell happened on that merge...

braddialpad:
oh my bad.. I merged next instead of staging lol

braddialpad:
@coderabbitai please review this PR over again

PR #1299, opened 2026-06-02
fix(combinator): DLT-3171 update v-model bindings from rendered component events

paulojreis-dialpad:

Obligatory GIF

- [x] Fix

## :book: Jira Ticket

[DLT-3171](https://dialpad.atlassian.net/browse/DLT-3171)

## :book: Description

Adds an `@event` listener to `<dtc-renderer>` in `combinator.vue` and a handler (`onComponentEvent`) that writes `update:<prop>` events back into the reactive options model via the existing `computedModel` setter. For any `update:*` event (e.g. `update:modelValue`, `update:open`), the emitted value is written to the matching key in `options.props` or `options.attributes`, completing the v-model loop and keeping the live preview and generated code snippet in sync.

Also adds unit tests to `combinator.test.js` covering boolean (DtToggle) and string (DtInput) modelValue writebacks, and a guard asserting that non-`update:` events (e.g. `change`) do not mutate options.

## :bulb: Context

v-model–driven components rendered in the Combinator playground were controlled but had no writeback path — clicking them had no visible effect. The emitted `update:modelValue` event was captured by `renderer_target.vue`, bubbled through `renderer.vue`, and silently dropped at `combinator.vue` because no `@event` listener existed. `options.bindings` only exposed `.get()`, never a write path, so `modelValue` was frozen at its initial value.

The ticket attributed the bug to the event-name → handler-key conversion (`on${capitalize(name)}`), but investigation confirmed that conversion already produces the correct key for all event name forms (verified against Vue 3.5's `emit` resolution in `@vue/runtime-core`). The sole root cause was the missing listener and writeback.

## :camera: Before / After

**Before**

Changing Model Value in the Props panel updates the component, but the opposite doesn't happen (component interaction doesn't update Model Value).

https://github.com/user-attachments/assets/552c0fbe-4bc1-4d5c-aa58-6ae572176eb1

https://github.com/user-attachments/assets/faa45a39-8a87-4cbd-902b-43f44ae19c71

**After**

Both directions work (component interaction now updates Model Value).

https://github.com/user-attachments/assets/bd05f56d-46bc-450c-a12e-781916cb22b0

https://github.com/user-attachments/assets/473c9cab-e253-49d5-ac53-6929783e8623

## :pencil: Checklist

For all PRs:

For all Vue changes:

[DLT-3171]: https://dialpad.atlassian.net/browse/DLT-3171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/combinator/src/components/combinator.test.js
+++ b/packages/combinator/src/components/combinator.test.js
@@ -2,8 +2,14 @@ import DtcCombinator from './combinator.vue';
 
 import { assert } from 'chai';
 import { shallowMount } from '@vue/test-utils';
+import { nextTick } from 'vue';
 import { getSupportedComponents } from '@/src/lib/test/utils_test';
-import documentation from '@/node_modules/@dialpad/dialtone-vue/dist/component-documentation.json';
+import { DtInput, DtToggle } from '@dialpad/dialtone-vue';
+import allDocs from '@/node_modules/@dialpad/dialtone-vue/dist/component-documentation.json';
+
+const documentation = allDocs;
+const toggleDoc = allDocs.find(d => d.displayName === 'DtToggle');
+const inputDoc = allDocs.find(d => d.displayName === 'DtInput');
 
 describe('combinator.vue test', function () {
   const testComponents = getSupportedComponents();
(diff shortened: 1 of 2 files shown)


francisrupert (approved):
Awesome, just as I hoped for.

braddialpad (approved):

PR #1308, opened 2026-06-03
revert: NO-JIRA accidental shadows push to next

francisrupert:
## Summary

Reverts the two next-shadows commits that were accidentally pushed to next:

- e616b1a49b84bd204ed19957b2553dab2d1990ae (init next shadow visual refresh)
- 961532ebbd78a425f6eeaefe87363ce02517f059 (overlay surface variant)

The shadow work has been preserved on the separate next-shadows branch, which now exists on the remote and is configured as the local branch upstream.

## Notes

- This PR does not rewrite next history.
- It removes the accidental changes by revert commit only.
- Local next-shadows now tracks origin/next-shadows, not origin/next.

## Verification

- Pre-commit hook completed successfully during the revert commit.
- git status --short --branch is clean on revert-next-shadows-from-next.
- git branch -vv confirms next-shadows tracks origin/next-shadows.


braddialpad (changes requested):
I think you are better off just pushing this revert to the `next` branch directly instead of via PR. The problem is that when you merge via GitHub it always squashes the commits which git would then not classify these commits as a revert, and could cause weird problems with merge conflicts in the future.

francisrupert:
closed without merging.

PR #1195, opened 2026-04-12
feat(focustrap): DLT-3297 add v-dt-focustrap directive for declarative focus trapping

francisrupert:
undoubtedly-its-a-trap-ovlpnki9ajlr2unq

## :hammer_and_wrench: Type Of Change

- [x] Feature

## :book: Jira Ticket

[DLT-3297](https://dialpad.atlassian.net/browse/DLT-3297)

## :book: Description

New `v-dt-focustrap` directive. Traps Tab/Shift+Tab within a container element. Configurable initial focus (auto, CSS selector, element ref, or disabled), Tab boundary wrapping, and focus restoration on deactivation. Reactive — bind to open/close state.

Replaces the need to manually wire the Modal mixin for focus trapping. Follows the same directive pattern as `v-dt-focusgroup`.

**Does NOT:** handle Escape, click-outside, `aria-modal`, scroll lock, or migrate existing components (separate follow-up tickets).

## :bulb: Context

Focus trapping is scattered across 5+ components via the Modal mixin, each wiring it differently. No shared primitive for new components like Hovercard (DLT-2845). This directive provides one consistent implementation.

## :pencil: Checklist

For all PRs:

- [x] I have ensured no private Dialpad links or info are in the code or pull request description (Dialtone is a public repo!).
- [x] I have reviewed my changes.
- [x] I have added all relevant documentation.
- [x] I have considered the performance impact of my change.

For all Vue changes:

- [x] I have added / updated unit tests.
- [ ] I have validated components with a screen reader.
- [x] I have validated components keyboard navigation.

## :crystal_ball: Next Steps

- Migrate existing components to use `v-dt-focustrap`: Modal, Banner, Popover, ImageViewer, Hovercard (DLT-2845), FeedItemRow (separate tickets per component)

## :camera: Screenshots / GIFs / Video

image

https://github.com/user-attachments/assets/c7134548-91ac-412a-9a6b-a6c4f739ba5a

Adds v-dt-focustrap Vue directive to trap Tab/Shift+Tab inside a container with configurable initial focus (auto, selector, element, or disabled), optional focus restoration, boundary wrapping, reactive activation, plus utilities, Storybook docs, tests, and a blog post.

[DLT-3297]: https://dialpad.atlassian.net/browse/DLT-3297?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1313, opened 2026-06-04
fix(image-viewer): DLT-3456 trap focus within the modal via v-dt-focustrap directive

paulojreis-dialpad:

Obligatory GIF

- [x] Fix
- [ ] Feature
- [ ] Performance Improvement
- [x] Refactor

## :book: Jira Ticket

[DLT-3456](https://dialpad.atlassian.net/browse/DLT-3456)

## :book: Description

Migrates `DtImageViewer`'s Tab focus-trapping to the `v-dt-focustrap` directive — and in doing so **fixes a real bug**: focus could escape the open modal. Part of [DLT-3452](https://dialpad.atlassian.net/browse/DLT-3452).

**The bug:** the old `trapFocus` called the Modal mixin's `focusTrappedTabPress(e)` with no element, so it scoped to `this.$el` — the component's *outer root* `<div>`, whose only focusable is the preview button. But the modal content (`.d-modal` + close button) is **teleported to `body`**, so it was never in the trapped scope. Tab was therefore never prevented and **focus leaked out of the open modal** in production. The existing unit tests stub `<Teleport>`, which renders the content inline and masked the bug.

**The fix (trap-only):**
- **`ImageViewer.vue`** — add `v-dt-focustrap="{ active: isOpen, initialFocus: false, restoreFocus: false }"` on the teleported `.d-modal` element, so the trap actually scopes to the modal content. Remove the Tab `case` from `modalListeners.keydown` and the `trapFocus` method. Drop the now-unused `Modal` mixin (`focusTrappedTabPress` was its only consumer; `focusAfterOpen` uses `returnFirstEl`, not the mixin) — `ModeMixin` is retained. The component keeps its own initial focus (`focusAfterOpen` → close button) and focus restoration (the `isOpen` watch), unchanged.
- **`ImageViewer.test.js`** — register `DtFocustrapDirective` in the mount `global.plugins`, add `attachTo: document.body`, and add a focus-trap test (a Tab keydown on the open modal is `defaultPrevented`; verified it fails without the active directive). Also fixed a harmless pre-existing missing-`]` in a `data-qa` selector.

## :bulb: Context

Focus trapping was scattered across components via the Modal mixin; DLT-3452 consolidates it on one directive (`v-dt-focustrap`, shipped in #1195). For ImageViewer the migration doubled as a bug fix, since the mixin-based trap was bound to the wrong element relative to the teleport.

**Trap-only:** the directive owns only the Tab trap (`initialFocus: false, restoreFocus: false`); ImageViewer keeps its existing initial-focus and restoration, so the only behaviour change is that focus is now actually contained (the bug fix).

**Global registration:** `v-dt-focustrap` is a globally-installed plugin — used in the template with no local `directives:` registration. Confirmed consumer apps already call `app.use(DtFocustrapDirective)`, and Storybook registers it in `preview.jsx`.

## :pencil: Checklist

For all PRs:

For all Vue changes:

Current state, illustrating the bug. Notice that 1) focus moves to the viewer's close button (expected); and 2) tabbing then moves to the browser address bar, dev tools, etc (i.e. it's not trapped):

https://github.com/user-attachments/assets/245ab8be-d3dc-47c5-902e-ea528df042af

With this PR, focus is actually trapped (video doesn't show it, but I'm actually trying to tab away from the close button):

https://github.com/user-attachments/assets/06478a2d-5d4b-4d28-ac74-65e0070efa40

[DLT-3456]: https://dialpad.atlassian.net/browse/DLT-3456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DLT-3452]: https://dialpad.atlassian.net/browse/DLT-3452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iropolo (approved):
looks good, thanks!

belumontoya (approved):

PR #1318, opened 2026-06-05
chore: NO-JIRA upgrade playwright to 1.60.0 to fix install hang on node…

paulojreis-dialpad:
Upgrades `playwright` to `^1.60.0` and `@storybook/test-runner` to `^0.24.4` to work around a regression in Node.js 24.16.0 ([nodejs/node#63487](https://github.com/nodejs/node/issues/63487)) that causes `playwright install` to hang indefinitely after the Chrome zip download completes. **Do not merge** — this is a diagnostic PR to confirm the durable fix before closing [#1317](https://github.com/dialpad/dialtone/pull/1317).

### DIFF
--- a/package.json
+++ b/package.json
@@ -173,7 +173,7 @@
     "markdownlint-cli": "^0.45.0",
     "npm-run-all": "^4.1.5",
     "nx": "19.8.0",
-    "playwright": "^1.49.0",
+    "playwright": "^1.60.0",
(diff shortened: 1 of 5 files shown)


braddialpad:
This is the preferred way to fix it, I will go with this version instead of the other PR and remove the DO NOT MERGE tag

PR #1323, opened 2026-06-12
feat(combinator): DLT-3480 add searchable filter to selection dropdowns

paulojreis-dialpad:

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

[DLT-3480](https://dialpad.atlassian.net/browse/DLT-3480)

## :book: Description

Adds a live search field inside the Combinator option bar's selection dropdowns. The field appears only when a dropdown has more than 5 options. Typing filters the list in real time against both the option label and its resolved token value. Keyboard shortcuts: **Enter** selects the first enabled match, **↓** moves focus to the first list item (from where DtDropdown's arrow-key navigation takes over), **Escape** closes the dropdown. Focus is moved to the search field automatically when the dropdown opens.

## :bulb: Context

Token-backed props (e.g. `size`, `color`, `kind`) can surface dozens of valid values. Scrolling a long flat list to find a specific value slows down component exploration in the playground. A search field makes the option bar significantly faster to use for any prop with many options.

## :pencil: Checklist

For all PRs:

For all Vue changes:

https://github.com/user-attachments/assets/466656d3-3300-46c7-9298-fe8d655cdc51

[DLT-3480]: https://dialpad.atlassian.net/browse/DLT-3480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/combinator/src/components/controls/control_selection.vue
+++ b/packages/combinator/src/components/controls/control_selection.vue
@@ -78,6 +78,21 @@
           <template #startIcon="{ iconSize }">
             <dt-icon-search :size="iconSize" />
           </template>
+          <template #endIcon="{ clear }">
+            <dt-button
+              :style="!query.length ? { visibility: 'hidden', pointerEvents: 'none' } : undefined"
+              kind="muted"
+              importance="clear"
+              :size="100"
+              circle
+              aria-label="Clear search"
+              @click="clear"
+            >
+              <template #startIcon="{ iconSize }">
+                <dt-icon-close :size="iconSize" />
+              </template>
+            </dt-button>
+          </template>
         </dt-input>
       </dt-box>
       <div ref="listWrapper" class="d-of-y-auto d-hmx-400">
@@ -137,7 +152,7 @@
 
 <script setup>
 import { DtBox, DtButton, DtInput, DtStack, DtText } from '@dialpad/dialtone-vue';
-import { DtIconChevronsUpDown, DtIconSearch } from '@dialpad/dialtone-icons/vue';
+import { DtIconClose, DtIconChevronsUpDown, DtIconSearch } from '@dialpad/dialtone-icons/vue';
 
 import { VALUE_UPDATE_EVENT } from '@/src/lib/constants';
 import { resolveTokenValue } from '@/src/lib/tokens';
--- a/packages/combinator/src/components/controls/control_selection.vue
+++ b/packages/combinator/src/components/controls/control_selection.vue
@@ -12,6 +12,7 @@
     placement="bottom-start"
     content-width="anchor"
     content-class="d-hmx-400"
+    @opened="onOpened"
   >
     <template #anchor="{ attrs }">
       <dt-button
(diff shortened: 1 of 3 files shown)


francisrupert:
Still in draft I know, though capturing our Friday discussion:

> "... replacing the current Combobox I did for icon slots. That one is also awkward when a value is already selected, and one may not fully realize you need to clear it out to see all other options."
- Do in this PR or follow-up PR and Jira story? Your call. 

> Sticky-positioned Search container: "...use DtBox for that one rather than `class="d-ps-sticky d-ibs-0 d-zi-base1 d-bgc-primary d-pbe-50 d-bb d-bc-subtle"`"
- though you'll see need some of those `class` that DtBox doesn't provide.
- extra bottom border probably isn't needed.

> "Would be great if we could preview icons too". 
- That'd be so lovely. I wonder if performance would be a concern. We do have this rarely used [Scroller](https://dialtone.dialpad.com/next/components/scroller.html) that might help. 
- Do in this PR or follow-up PR and Jira story? Your call. 

> "...once I start mousing over dropdown, the input focus is lost."

Great stuff, great to see this incrementally improve!

francisrupert:
Another UI nit: Add a `x` button in end slot. You can see that in leftbar search field, and a few other locations. See `#search` on https://dialtone.dialpad.com/next/components/input.html#search

paulojreis-dialpad:
@francisrupert I've addressed most of your feedback, LMK what you think.

Performance-wise (with the icons) I think it's OK but I'm also conscious that I'm running this on a v12 500HP laptop… might be good enough for an internal tool, but let me know if you'd like to see the virtual scroller implemented.

https://github.com/user-attachments/assets/e017d49b-e0e7-4d57-adb7-bcefda917347

packages/combinator/src/components/controls/control_selection.vue
```diff
@@ -53,49 +61,107 @@
...
+            <dt-icon-search :size="iconSize" />
+          </template>
+          <template #endIcon="{ clear }">
+            <dt-button
+              :style="!query.length ? { visibility: 'hidden', pointerEvents: 'none' } : undefined"
+              kind="muted"
+              importance="clear"
+              :size="100"
+              circle
+              aria-label="Clear search"
```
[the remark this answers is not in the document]
francisrupert: ignore. doesn't apply to doc site.

packages/combinator/src/components/controls/control_selection.vue
```diff
@@ -53,49 +61,107 @@
...
-            kind="body"
-            :size="100"
-            tone="muted"
-            class="d-mis-auto"
+          <template #startIcon="{ iconSize }">
+            <dt-icon-search :size="iconSize" />
+          </template>
+          <template #endIcon="{ clear }">
+            <dt-button
+              :style="!query.length ? { visibility: 'hidden', pointerEvents: 'none' } : undefined"
```
francisrupert: Something more like...
```suggestion
              v-if="query.length !== 0"
```
See the sidebar's search field.

packages/combinator/src/components/controls/control_selection.vue
```diff
@@ -53,49 +61,107 @@
...
+              kind="muted"
+              importance="clear"
+              :size="100"
+              circle
```
francisrupert:
```suggestion
              class="d-bgc-transparent"
```
Remove `circle`, and add `d-bgc-transparent` to the button. It's not ideal, and we probably need a more official slot for Inputs (like DtButton has with `trailing`. I'm fine it it not have hover/active background color.

packages/combinator/src/components/controls/control_selection.vue
```diff
@@ -53,49 +61,107 @@
...
+      <dt-text
+        v-if="!filteredOptions.length"
+        kind="body"
+        :size="100"
+        tone="muted"
+        as="div"
+        class="d-px-200 d-py-150 d-ta-center"
+      >
+        No matches
+      </dt-text>
```
francisrupert: Use DtEmptyState. See sidebar search, something like...

```html
<dt-empty-state
  size="200"
  body-text="No matches"
/>
```

packages/combinator/src/components/controls/control_selection.vue
```diff
@@ -53,49 +61,107 @@
...
-        v-for="option in options"
-        :key="option.value"
-        role="menuitem"
-        navigation-type="arrow-keys"
-        :class="{ 'd-o50 d-pe-none': option.disabled, 'd-bgc-moderate-opaque': option.value === value }"
-        :aria-disabled="option.disabled || undefined"
-        @click="!option.disabled && (onInput(option.value), close())"
+      <dt-box
+        v-if="showSearch"
+        class="dtc-search-box"
```
francisrupert:
```suggestion
        class="d-ps-sticky d-ibs-0 d-zi-navigation-fixed"
```

packages/combinator/src/components/controls/control_selection.vue
```diff
@@ -53,49 +61,107 @@
...
+              class="d-ba d-bc-subtle d-bar-circle d-as-center"
+              :style="swatchStyle(option.resolved)"
+            />
+            <span>{{ option.label }}</span>
+            <component
+              v-if="option.previewComponent"
+              :is="option.previewComponent"
+              size="200"
+              class="d-mis-auto d-as-center d-fc-muted"
+            />
```
francisrupert, line 123: Let's have the icon render on the left. Definitely bump the size to at least `300`, maybe `400`. Whatever works well. Heck, I wouldn't necessarily mind if it were even bigger – just like you see in Figma. Though the selected one shouldn't increase the size of the unopened Dropdown.

Image

francisrupert: Muted color icon too dim. Consider `d-fc-tertiary` at least, perhaps `d-fc-secondary`.

packages/combinator/src/components/controls/control_selection.vue
```diff
@@ -168,6 +239,66 @@ const selectedOption = computed(() => {
...
+
+/**
+ * Handles keys in the search field. Arrow keys propagate to DtDropdown so its
+ * highlightIndex stays in sync (avoids the double-keystroke bug). All other
+ * keys are stopped to prevent DtDropdown's type-ahead from intercepting typing.
+ *
+ * @param {KeyboardEvent} e - The keydown event.
+ * @param {Function} close - Closes the dropdown.
+ */
+function onSearchKeydown (e, close) {
```
francisrupert, line 280: After I arrow down to the list, I should be able to arrow back up to the input field. See doc sidebar search. 

https://github.com/user-attachments/assets/6f3136f9-fc17-4a39-8d2b-2a2f29baa906

packages/combinator/src/components/controls/control_selection.vue
francisrupert, line 1: I don't have a strong judgement on the performance. Probably ok to go with this and keep an eye on it. 99% of users and traffic are internal on screaming machines with strong connection.

The DtScroller might be one potential would-be future refactor, even paired with DtLoader on search like this Figma equivalent.

https://github.com/user-attachments/assets/9a3467b7-b97d-4a2c-9d69-7dbd7d4fabbe

francisrupert (changes requested):
Love everything about this. Consider the coderabbit and other comments.

packages/combinator/src/components/controls/control_selection.vue
paulojreis-dialpad: This one is slightly annoying, but I have to admit I didn't like the solution either… This particular element is the tallest in the search container and therefore it determines its height. Adding/removing it from the DOM will cause a slight jump in height, which I wanted to avoid. Playing with the element's CSS visibility (rather than `display: none;`ing it or removing it from the DOM) ensures that the element is kind of always there (and taken into account when computing dimensions).

I'll try and find a nicer way to do this.

paulojreis-dialpad: I've managed to make the `v-if` work nicely by using a default-sized `DtInput`. Basically recreated the left sidebar search bar.

Just noting that with a `:size="100"` input, the content jumps a bit when using the `v-if` approach:

https://github.com/user-attachments/assets/778696c1-e24b-4494-9f66-1894f4b36a92

Should I raise a ticket?

packages/combinator/src/components/controls/control_selection.vue
paulojreis-dialpad: Done ✅

packages/combinator/src/components/controls/control_selection.vue
paulojreis-dialpad: Done ✅  (and thanks! Good to know this exists)

packages/combinator/src/components/controls/control_selection.vue
paulojreis-dialpad: Done ✅

packages/combinator/src/components/controls/control_selection.vue
paulojreis-dialpad: > Let's have the icon render on the left. Definitely bump the size to at least `300`, maybe `400`

400 works good I think

image

packages/combinator/src/components/controls/control_selection.vue
paulojreis-dialpad: Should be sorted now 👍

packages/combinator/src/components/controls/control_selection.vue
paulojreis-dialpad: I might give it a go, just to become familiar with the API

packages/combinator/src/components/controls/control_selection.vue
francisrupert: that looks like it's because the `×` button is too tall. No worries for now. A better future task is more likely to have a proper `trailing` slot for DtInput, DtSelect and adjust from there. Right now, placing an actionable item (button) in the endIcon slot is kind of a hack, but still not too horrible.

francisrupert: it works a dream.

francisrupert (approved):
💪 frickin awesome

braddialpad (approved):

paulojreis-dialpad:
Thanks @francisrupert and @braddialpad for the review. Just added a commit to fix a regression I've introduced… should be good after that.

iropolo (approved):
Looks good, thanks!!

PR #1327, opened 2026-06-12
fix(rich-text-editor): DP-188940 fix cursor jump after toggleCodeBlock

briandial:

fixing it

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DP-188940

## :book: Description

After `toggleCodeBlock` converts a multi-line code block back to paragraphs, the cursor jumped to the start of the next line instead of staying on the current line.

## :bulb: Context

ProseMirror's `replaceWith` uses right-biased step mapping (`assoc=1`) by default. When a multi-line code block is split into paragraphs, each `\n` becomes a paragraph boundary, and right-bias places the cursor at the start of the *next* paragraph.

Fix: track `prevAnchor` and `prevInCodeBlock` via the `selectionUpdate` event. On any doc-changing `transaction` that exits a code block, remap the anchor with left-bias (`assoc=-1`) and dispatch `setTextSelection` in `setTimeout(0)`, after ProseMirror's own step mapping has run.

The `setTimeout(0)` is necessary because setting selection inside a Tiptap chain command is overridden by the EditorView's DOM reconciliation after dispatch. The `isDestroyed` guard prevents the callback from running if the component unmounts in the interim.

## :pencil: Checklist

Verified locally via Playwright against Storybook. No visual changes — behaviour fix only.

### DIFF
--- a/packages/dialtone-vue/components/rich_text_editor/rich_text_editor.test.js
+++ b/packages/dialtone-vue/components/rich_text_editor/rich_text_editor.test.js
@@ -2152,6 +2152,42 @@ describe('DtRichTextEditor tests', () => {
       });
     });
 
+    describe('Code block cursor fix tests', function () {
+      beforeEach(async function () {
+        await wrapper.setProps({ allowCodeblock: true, modelValue: '' });
+        await wrapper.vm.$nextTick();
+      });
+
+      describe('When toggling a multi-line code block with a trailing newline back to paragraphs', function () {
+        it('cursor stays in a content paragraph instead of jumping to the empty trailing paragraph', async function () {
+          const ed = wrapper.vm.editor;
+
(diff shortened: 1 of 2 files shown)


braddialpad (changes requested):
couple of things I've noticed here.

1. If I put the cursor on an individual point within the codeblock it is fixed, however if all the text in the codeblock is highlighted it's the same problem as before

Image

2. If you insert a codeblock and then remove it, the cursor now goes to the start of the (prior) codeblock instead of the end. Not a huge problem but I think it feels more natural if it goes to the end (without going to the next line of course)

francisrupert:
Close? Revisit?

braddialpad:
@briandial did you see my comments?

PR #1330, opened 2026-06-17
feat(hovercard): DLT-2845 add focus trap to hovercard content

paulojreis-dialpad:
# feat(hovercard): DLT-2845 add focus trap to hovercard content

Obligatory GIF

- [x] Feature

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DLT-2845

## :book: Description

Binds `v-dt-focustrap` to the Hovercard content wrapper, trapping Tab/Shift+Tab within the card while it is open.

## :bulb: Context

~~This is a **draft for discussion**; a few open questions should be resolved before merging:~~

~~1. **Do we want to trap focus in this particular case?** Hovercard is explicitly non-modal (`:modal="false"`)~~

~~2. If we do want to trap focus, it's worth noting that **there is currently no keyboard entry path.** The hovercard only shows on `mouseenter`. WCAG seems to suggest that "Content which can be triggered via pointer hover should also be able to be triggered by keyboard focus". This implies that we should show the hovercard when trigger element is (keyboard) focused — are we happy with this?~~
~~- A related side-effect of this is that if we show the hovercard on `mouseenter` only, the user might have to _tab_ quite a lot until they get into the hovercard (depending on which element is focused when the card is shown). We may want to consider automatically focusing an element in the hovercard via `initialFocus`~~

Discussed with design engineering and agreed that:

- The hovercard is to be explicitly framed as a progressive enhancement: anything it surfaces must also be reachable another way, so keyboard/AT users are never blocked 
- Because of this, and because it has the potential to become noisy and disruptive, we will not be adding keyboard-driven ways of triggering the hovercard — there's no need to, content and actions are available elsewhere, and excessive "pop-ups" being triggered with keyboard navigation will affect the UX negatively
- Focus trapping is still desirable, but only within the scope of the hovercard. Once focus enters the hovercard (by e.g. mouse click), tabbing should be constrained to the hovercard focusable elements
- Escape should dismiss the hovercard; focus should be restored to where it came from
- Documentation needs to provide guidance and reflect our accessibility practices

This PR touches the Popover, which was out of scope. This is because the Hovercard essentially wraps the Popover component, and controlling the focustrapping at the Popover-level was the only way to ensure we covered all 3 slots (header, content, footer). More than happy to consider other approaches to this.

## :pencil: Checklist

Focus trapping:

https://github.com/user-attachments/assets/5f88724a-2e6f-48cc-8c7d-de2ddcfccb33

Notice how tabbing doesn't trigger the hovercard (illustrates lack of keyboard entry path):

https://github.com/user-attachments/assets/d7295312-2bb2-4844-8016-0b5292c1476f

Changes to the docs:

image

### DIFF
--- a/apps/dialtone-documentation/docs/components/hovercard.md
+++ b/apps/dialtone-documentation/docs/components/hovercard.md
@@ -33,6 +33,33 @@ The hovercard will appear upon the mouse entering the anchor, with a delay of 30
 
 <!-- <component-combinator component-name="DtHovercard" /> -->
 
+## Usage
+
+- Hovercard is a **progressive enhancement**. It surfaces supplementary information about an entity or provides convenience shortcuts on hover, without navigating away from the current view. Every action it exposes **must also be reachable by another route** (a dedicated page, a menu, or an inline button).
+- Keep content concise. Hovercards are transient overlays — deep interactions, multi-step flows, or extensive content belong in a dedicated UI, or in a [Modal](modal.md).
+- Use Hovercard for contextual content with structure or actions. Use [Tooltip](tooltip.md) for brief, text-only descriptions of a control. Use [Popover](popover.md) when the content is triggered by a deliberate click and should persist until dismissed.
+- `enterDelay` is used to prevent the card from triggering on accidental cursor passes. The default delay of 300 ms is recommended for most contexts.
+
+<dialtone-usage>
+<template #do>
+
+- Surface contextual details like a user's status, avatar, and quick-action shortcuts.
+- Treat hovercard actions as shortcuts — ensure the same actions are reachable elsewhere in the UI.
+
+</template>
+<template #dont>
+
+- Use Hovercard as the only way to reach content or perform an action.
+- Place critical or destructive actions exclusively inside a hovercard.
+- Use for deep interactions or lengthy content — prefer a dedicated UI or a [Modal](modal.md).
+
+</template>
+</dialtone-usage>
+
+## Accessibility
+
+When focus moves into an open hovercard, focus is trapped within it — the user can Tab between focusable elements inside the card without accidentally leaving. Pressing Escape dismisses the card and restores focus to the element that had focus before the card opened.
+
 ## Variants
 
 ### Many Hovercards
--- a/apps/dialtone-documentation/docs/components/hovercard.md
+++ b/apps/dialtone-documentation/docs/components/hovercard.md
@@ -36,7 +36,7 @@ The hovercard will appear upon the mouse entering the anchor, with a delay of 30
 ## Usage
 
 - Hovercard is a **progressive enhancement**. It surfaces supplementary information about an entity or provides convenience shortcuts on hover, without navigating away from the current view. Its content or functionality **must also be reachable by another route** (a dedicated page, a menu, or an inline button).
-- Keep content concise. Hovercards are transient overlays — deep interactions, multi-step flows, or extensive content belong in a dedicated UI, or in a [Modal](modal.md).
+- Keep content concise. Hovercards are transient overlays, and are not appropriate for deep interactions, multi-step flows, or extensive content.
 - Use Hovercard for contextual content with structure or actions. Use [Tooltip](tooltip.md) for brief, text-only descriptions of a control. Use [Popover](popover.md) when the content is triggered by a deliberate click and should persist until dismissed.
 - `enterDelay` is used to prevent the card from triggering on accidental cursor passes. The default delay of 300 ms is recommended for most contexts.
 
--- a/apps/dialtone-documentation/docs/components/hovercard.md
+++ b/apps/dialtone-documentation/docs/components/hovercard.md
@@ -58,7 +58,7 @@ The hovercard will appear upon the mouse entering the anchor, with a delay of 30
 
 ## Accessibility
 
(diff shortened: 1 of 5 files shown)


apps/dialtone-documentation/docs/components/hovercard.md
```diff
@@ -33,6 +33,33 @@ The hovercard will appear upon the mouse entering the anchor, with a delay of 30
...
+- Use Hovercard as the only way to reach content or perform an action.
+- Place critical or destructive actions exclusively inside a hovercard.
+- Use for deep interactions or lengthy content — prefer a dedicated UI or a [Modal](modal.md).
+
+</template>
+</dialtone-usage>
+
+## Accessibility
+
+When focus moves into an open hovercard, focus is trapped within it — the user can Tab between focusable elements inside the card without accidentally leaving. Pressing Escape dismisses the card and restores focus to the element that had focus before the card opened.
```
francisrupert:
```suggestion
When focus moves into an open hovercard, focus is trapped within. The user can Tab between focusable elements inside the card without accidentally leaving. Clicking outside the hovercard or keypress of `esc` will dismiss the card and restore focus to the element that had focus before the card opened.
```

apps/dialtone-documentation/docs/components/hovercard.md
```diff
@@ -33,6 +33,33 @@ The hovercard will appear upon the mouse entering the anchor, with a delay of 30
 
 <!-- <component-combinator component-name="DtHovercard" /> -->
 
+## Usage
+
+- Hovercard is a **progressive enhancement**. It surfaces supplementary information about an entity or provides convenience shortcuts on hover, without navigating away from the current view. Every action it exposes **must also be reachable by another route** (a dedicated page, a menu, or an inline button).
```
francisrupert:
```suggestion
- Hovercard is a **progressive enhancement**. It surfaces supplementary information about an entity or provides convenience shortcuts on hover, without navigating away from the current view. Its content or functionality **must also be reachable by another route** (a dedicated page, a menu, or an inline button).
```

apps/dialtone-documentation/docs/components/hovercard.md
```diff
@@ -33,6 +33,33 @@ The hovercard will appear upon the mouse entering the anchor, with a delay of 30
...
+- Keep content concise. Hovercards are transient overlays — deep interactions, multi-step flows, or extensive content belong in a dedicated UI, or in a [Modal](modal.md).
```
francisrupert:
```suggestion
- Keep content concise. Hovercards are transient overlays, and are not appropriate for deep interactions, multi-step flows, or extensive content.
```

packages/dialtone-vue/components/Hovercard/Hovercard.test.js
```diff
@@ -233,5 +236,22 @@ describe('DtHovercard Tests', () => {
...
+
+        const popover = wrapper.findComponent(DtPopover);
+        expect(popover.props('focustrap')).toBe(true);
+      });
+
+      it('passes a falsy focustrap to DtPopover when hovercard is closed', () => {
+        const popover = wrapper.findComponent(DtPopover);
+        expect(popover.props('focustrap')).toBeFalsy();
+      });
+    });
```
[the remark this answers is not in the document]
paulojreis-dialpad: Acknowledged; partially addressed.

Added behavioral Escape test (closes and restores focus to the anchor on Escape).

Couldn't add focus-trapping due to JSDOM constraints when mocking teleported DOM elements, so these remain prop-tests.

francisrupert (commented):
Initial minor content suggestions.

apps/dialtone-documentation/docs/components/hovercard.md
francisrupert: > "Pressing Escape dismisses the card and restores focus to the element that had focus before the card opened."

This is currently not true. In attached video, I...

1. Focus in middle of the page.
2. Hover over top nav, tab within trapped Hovercard.
3. Hit escape...
  1. At this point, I expect the focus to go to the last item I had focused before bringing up hovercard.
  2. It currently appears to go to the first focusable element at root.

Should we keep track of the last focused item once the Hovercard has focus within? 

https://github.com/user-attachments/assets/090980d7-4c61-4963-aa6e-0d341af66133

francisrupert: Beacon example.

https://github.com/user-attachments/assets/8218e569-a05f-4f7c-a7f5-521b0021c3a0

braddialpad (approved):
Looks good IMO

apps/dialtone-documentation/docs/components/hovercard.md
paulojreis-dialpad: Hmmm, this should've been fixed with https://github.com/dialpad/dialtone/pull/1330/changes/ddc769934e9fec48adde7887d629e49ebd3e463a 

https://github.com/user-attachments/assets/95ac7f84-428c-49c7-b517-ef59a3dd986c

Try again please and let me know if you're still seeing the issue

francisrupert: Confirmed. It does. Merge away!

https://github.com/user-attachments/assets/66f42de1-d0f8-46e4-8078-838bd318f82a

francisrupert (approved):

iropolo (approved):

PR #1334, opened 2026-06-22
feat(text): DLT-3487 expose full typography size scale

hynes-dialpad:

- [x] Feature

## :book: Jira Ticket

DLT-3487

## :book: Description

- Exposes the full typography font-size token scale through `DtText` `size` for raw font-size control, including the new 50, 75, 450, and 550-800 stops while keeping 125.
- Adds `variant`, `family`, and `italic` support to `DtText`, with `variant` used for complete typography composition and numeric `size` available as an override.
- Updates typography tokens, CSS font-size utilities, Vue docs, migration docs, generated docs data, and component thumbnails.
- Aligns ESLint migration rules so deprecated DtText t-shirt sizes migrate toward `variant` instead of the older numeric `size` suggestions.

## :bulb: Context

`DtText` previously coupled composition and size through `kind + size`, and the `size` prop did not expose all available font-size token stops. This keeps the legacy API backward compatible while giving new usage a clearer split: `variant` for text composition, `size` for raw font-size tokens, `family` for font-family selection, and `italic` for font style.

## :white_check_mark: Validation

Passed locally:

- `pnpm --dir packages/dialtone-vue exec vitest run components/Text/Text.test.js --test-timeout=10000`
- `pnpm --dir packages/eslint-plugin-dialtone exec mocha tests/lib/rules/deprecated-tshirt-sizes.js tests/lib/rules/deprecated-headline-sizes.js`
- `pnpm --dir packages/dialtone-css run lint` (passes with existing warnings)
- `pnpm --dir packages/dialtone-docs run build`
- `pnpm --dir packages/dialtone-docs run test`
- `pnpm --dir packages/dialtone-docs run check-freshness`
- `pnpm --dir packages/dialtone-tokens run test`
- `pnpm --dir apps/dialtone-documentation exec markdownlint docs/components/text.md docs/guides/migration/typography/index.md docs/utilities/typography/font-size.md`

Local caveats:

- `pnpm nx run eslint-plugin-dialtone:test` hit local Nx project graph/daemon issues and later hung; focused rule tests passed directly.
- `pnpm --dir apps/dialtone-documentation run lint` fails on existing unrelated markdownlint issues outside this PR's changed typography docs.

## :pencil: Checklist

For all PRs:

For all Vue changes:

For all CSS changes:

## :crystal_ball: Next Steps

- Verify the generated documentation preview and component thumbnails in CI/deploy preview.
- Confirm whether future ESLint work should remove deprecated t-shirt size support after migration adoption.

Not included. The PR updates generated docs thumbnails and should be reviewed in the docs deploy preview.

## :link: Sources

None.


belumontoya (commented):
i think most of coderabbits comments about test are correct specially for order and consistency so i would check them out

francisrupert:
Firstly, Codex and Coderabbit responses should definitely be addressed. 

Other than that, I like this a lot. I will do a follow-up PR to fix some of the Combinator awkwardness with this DtText change. Here's my main feedback and it's **important** and my strong opinion:

**DtText must not be used with `size` prop _by itself_.**

This **_should not_** be treated as valid usage:

```vue
<dt-text size="500"> ... </dt-text>
```

`variant` should be the heavily-favored typography composition. Thus, `size` must only be valid in one of two contexts:

1. With `variant`, in the rare case where it acts as an intentiona font-size override:

```vue
<dt-text variant="body-md" size="400"> ... </dt-text>
```

2. With `kind` (legacy), preserving existing `kind` / `size` composition:

```vue
<dt-text kind="body" size="300"> ... </dt-text>
```


I strongly cringe seeing **_just_** `size`. No different or better than a raw css uttility `<span class="d-fs-500">`

That is the pattern we are trying to move away from. The point of `variant` is to keep keep it composed, intentional typography styles first. Thus with `variant`, `size` should be considered rarely as an escape hatch: start from a composition like `body-md` and override just font size.

Update the implementation so `size` without either `variant` or legacy `kind` does not apply a raw font-size class. In `getFontSizeClass`, I think we need one more guard: if `size` is set but neither `variant` nor `kind` is set, return `null` and warn that `size` must be paired with `variant` or `kind`.

I think this should also be enforced in `eslint-plugin-dialtone`. If `DtText` has a `size` prop, it should also have either `variant` or legacy `kind` on the same component. This should be linted as an **_error_**.

Lint should steer toward `variant` first:

> `DtText`'s `size` must be paired with `variant` or legacy `kind`. Prefer `variant` for typography composition; use `size` only as a variant override.

Don't think it should be autofixable since something like `size="500"` alone isn't enough to know which `variant` or `kind` to use. An error with manual guidance is best approach.

Implementation-wise, this probably belongs near the existing `DtText` linting in `packages/eslint-plugin-dialtone`, alongside `deprecated-headline-sizes`. The rule should catch both `size="..."` and `:size="..."` when neither `variant` nor `kind` is present.

And lastly, consider how to update the guidance and do/don't to speak to this in `apps/dialtone-documentation/docs/components/text.md`

francisrupert (changes requested):
.

packages/dialtone-vue/components/Text/Text.vue
```diff
@@ -75,12 +76,15 @@ export default {
...
-     * Typography variant mapping to a complete token-backed composition.
+     * Typography variant mapping to a complete token-backed composition. Defaults to body-md unless legacy kind is present.
      * @values headline-3xl, headline-2xl, headline-xl, headline-lg, headline-md, headline-sm, headline-xs, body-lg, body-md, body-sm, body-xs, label-lg, label-md, label-sm, label-xs, code-lg, code-md, code-sm, code-xs
      */
     variant: {
       type: String,
-      default: null,
+      default(rawProps) {
+        return rawProps.kind ? null : DEFAULT_VARIANT;
+      },
```
francisrupert: Yikes, oh no. defaulting `<dt-text>...</dt-text>` to `variant="body-md"` is a significant **_breaking change_**, and I'm gonna say nope to that.

Inititally you had `variant` correctly default to `null`, which allowed `<dt-text>` to continue to inherit styles until you declare a `variant` or `kind/size`. DtText is not always about declaring font/line-height style, e.g. `<dt-text tone="muted">`, `<dt-text as="p" truncate>`, `<dt-text numeric>`, `<dt-text as="p" align="center">`. 

Keep `variant` defaulting to `null`, retaining previous DtText inheritance, require an explicit `variant` or `kind` when `size` is present. Use of DtText without `variant` or `kind` should remains valid, and can be used for other props that don't set font-size / line-height.

francisrupert (changes requested):
@hynes-dialpad I see you made some updates, though hadn't officially re-requested review – but this jumped out at me and must be resolved.

packages/dialtone-vue/components/Text/Text.vue
hynes-dialpad: Having to declare a variant/kind all the time seems arduous to me. I don't agree that we should be pushing people to always use type styles. They are preferred. They are suggested. But I don't think every DtText instance should need to declare a type style.

One reason is that it fights against CSS cascade. People should only declare values when needed. A `body-md` style is basically the universally declared type style (base style). Why do I need to re-declare it every time? It should cascade down into the DtText component. Secondly this makes DtText act different than DtStack or DtBox where props are additive. Both provide opinionated defaults out of the box. The difference with DtText is that an opinionated default should be almost nothing. I think almost every prop for DtText should be additive, none required.

I can revert the default `variant`, but I would not require `variant` or `kind` to be present for `size` to work. Having `<dt-text size="300">` is the same as `<span class="d-fs-300">`, which, while not something we would want to encourage, it should be possible when needed like in `DtBadge` and `DtChip` where the type styles are declared on the parent and size changes are nested within. 

Happy to discuss this further if it helps resolve items further before making anymore code changes.

packages/eslint-plugin-dialtone/tests/lib/rules/deprecated-tshirt-sizes.js
```diff
@@ -124,6 +184,16 @@ ruleTester.run('deprecated-tshirt-sizes', rule, {
       code: `<template><dt-button :size="'sm'" /></template>`,
       errors: [{ messageId: 'deprecatedSizeInBinding' }],
     },
+    // Dynamic binding on dt-text without variant/kind cannot infer composition
+    {
+      code: `<template><dt-text :size="'sm'" /></template>`,
+      errors: [{ messageId: 'unpairedDtTextSize' }],
```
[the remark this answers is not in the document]
francisrupert: @hynes-dialpad Agree with this review, and its "Proposed test hardening" suggestion.

apps/dialtone-documentation/docs/guides/migration/typography/index.md
```diff
@@ -345,16 +384,33 @@ npx eslint --fix "./src/**/*.vue"
...
+### Raw Font-Size Utilities
+
+Use the `size` prop for raw font-size token replacement only when it is paired with `variant` for a complete text composition, or with legacy `kind` while migrating older code. Do not migrate a raw font-size utility to `<dt-text size="…">` by itself.
+
+```html
+<!-- Before -->
+<p class="d-fs-200">Body text</p>
+
+<!-- After: composed style plus exact font-size token -->
+<dt-text as="p" variant="body-md" size="200">Body text</dt-text>
```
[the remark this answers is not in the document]
francisrupert: @hynes-dialpad This is an acceptable before/after since it falls under *manual migration*, and is a reasonable aim to migrate out raw CSS Utility use of `d-fs-*` But it requires a manual judgement call to retain use of `<el class="d-fs-*">` or update it to a DtText with `variant` (ideal) or `kind/size` pairing (less ideal and discouraged).

Suggestion: update the doc prose to reflect that distinction.

apps/dialtone-documentation/docs/components/text.md
```diff
@@ -17,79 +31,103 @@ keywords: ["dt-box", "DtBox", "typography", "size", "tone", "font", "font size",
...
-  <dt-text kind="headline" as="span">Headline</dt-text>
-  <dt-text kind="body">Body</dt-text>
-  <dt-text kind="label">Label</dt-text>
-  <dt-text kind="code">Code</dt-text>
+<dt-stack
+  gap="400"
+  :direction="{ default: 'column', md: 'row' }"
+  align="baseline"
+>
+  <dt-text variant="headline-md" as="span">Headline</dt-text>
```
francisrupert:
```suggestion
  <dt-text variant="headline-md" as="h3">Headline</dt-text>
```
Nit: change `as` to an `<h*>` element

apps/dialtone-documentation/docs/components/text.md
```diff
@@ -17,79 +31,103 @@ keywords: ["dt-box", "DtBox", "typography", "size", "tone", "font", "font size",
...
+  <dt-text variant="label-md">Label</dt-text>
+  <dt-text variant="code-md">Code</dt-text>
   <dt-text>Default (inherits)</dt-text>
 </dt-stack>
 ```
 
 ### Size
 
-All kinds support `size` prop, but not all sizes are available for each kind. When `kind` is set, size defaults to `300` if not specified.
+Use `size` with `variant` to apply a raw font-size token while keeping the rest of the composition. `size` should not be used by itself; pair it with `variant`, or with legacy `kind` while migrating older usages.
```
francisrupert:
```suggestion
Using `size` with `variant` acts as an override of the `font-size` over its composed default. It applies a raw font-size token while keeping the rest of the composition. `size` will not work by itself, as it requires a pairing of `variant`, or with legacy `kind` when migrating older usages.
```

apps/dialtone-documentation/docs/components/text.md
```diff
@@ -17,79 +31,103 @@ keywords: ["dt-box", "DtBox", "typography", "size", "tone", "font", "font size",
...
     </tr>
   </table>
 </dt-stack>
 
 ```vue code-only
-<dt-text kind="{kind}" :size="{size}">....</dt-text>
+<dt-text
+  variant="body-lg"
+  size="300"
+>Body composition with a 20px font-size override</dt-text>
```
francisrupert:
```suggestion
<dt-text
  variant="body-xs"
  size="125"
>Body composition with font-size override  to fall between "body-xs" and "body-sm" </dt-text>
```
`body-lg`+`300` doesn't make sense since that's equivalent to just `body-md`. 

This example demos a relevant use case of "start with `body-xs`, but override with `size="125"` to give me a size that falls between `body-xs` and `body-sm`.

apps/dialtone-documentation/docs/components/text.md
```diff
@@ -17,79 +31,103 @@ keywords: ["dt-box", "DtBox", "typography", "size", "tone", "font", "font size",
...
 ```
 
+### Legacy kind
+
+The `kind` prop remains supported for backward compatibility. When `kind` is present and `variant` is absent, `size` keeps its legacy composition behavior.
```
francisrupert:
```suggestion
> [!WARNING] Favor using `variant` over `kind/size` combinations. The `kind` prop remains supported for backward compatibility. 

When `kind` is present and `variant` is absent, `size` keeps its legacy composition behavior.

`kind` is ignored if `variant` is present. 
```

apps/dialtone-documentation/docs/components/text.md
```diff
@@ -466,15 +540,27 @@ Text box trim will only affect elements with block or inline-block styled contex
...
+  { size: 350, output: '22px' },
+  { size: 400, output: '26px' },
+  { size: 450, output: '29px' },
+  { size: 500, output: '32px' },
+  { size: 550, output: '36px' },
+  { size: 600, output: '42px' },
+  { size: 650, output: '46px' },
+  { size: 700, output: '52px' },
+  { size: 750, output: '58px' },
+  { size: 800, output: '66px' },
```
francisrupert: Let's use the existing `@data/type.json` source here instead of hard-coding the same scale

`fontSize.product` already has the stop + px mapping this table needs, so deriving `textSizeScale` from that would reduce (not solve) drift risk and keep this page aligned with the typography utility docs. 

It still doesn’t fully solve the underlying problem, since `type.json` itself is manually authored static data we should eventually automate from the token source too, among others. That's a different task; for this PR, using `@data/type.json` is acceptablestep of not adding another duplicate list.

apps/dialtone-documentation/docs/guides/migration/typography/index.md
```diff
@@ -9,13 +9,17 @@ status: ready
...
 - Override utilities like `d-fw-*`, `d-fc-*`, `d-lh-*`, `d-truncate`, and `d-ta-*` map to `strength`, `tone`, `density`, `truncate`, and `align` props.
 - One command does the migration: `npx dialtone-migrate-typography`.
 
 ## Overview
 
-[DtText](/components/text.html) is Dialtone's semantic text component. It exposes typography intent through props (`kind`, `size`, `density`, `strength`, `tone`, `truncate`, `align`) rather than utility classes, making typography decisions explicit and easier to validate at build time.
+[DtText](/components/text.html) is Dialtone's semantic text component. It exposes typography intent through props (`variant`, `size`, `density`, `strength`, `tone`, `truncate`, `align`) rather than utility classes, making typography decisions explicit and easier to validate at build time.
+
+For backward compatibility, `kind` remains supported and keeps the historical composed typography behavior when `variant` is not set. That means existing `<dt-text kind="body" size="300">` output from the migration tool is still valid. Prefer `variant` for new manual code and for cleanup you do after running the tool.
```
francisrupert, line 22: Can you log a backlog task for a future codemod-ish migration tool that migrations from `kind+size` combinations to `variant`? It should speculate what that migration script would look like, but it should probably be a future migration effort so we don't get in the way of the current migration tool which is doing a good job.

apps/dialtone-documentation/docs/guides/migration/typography/index.md
```diff
@@ -36,7 +40,7 @@ The tool is ideal for projects with many text elements using utility classes. Fo
 <div class="d-d-grid d-g-200 d-g-cols1 md:d-g-cols2">
 <div>
 
-**Before**
+<strong>Before</strong>
```
francisrupert: wat?

apps/dialtone-documentation/docs/components/text.md
```diff
@@ -1,10 +1,24 @@
 ---
 title: Text
-description: Consistent typography styling through semantic text kinds and sizes.
-status: new
+description: Consistent typography styling through text variants and raw font-size tokens.
+status: ready
```
francisrupert:
```suggestion
status: new
````

bring this back to `new`

francisrupert (approved):
Great update. I have updates for the Combinator in https://github.com/dialpad/dialtone/pull/1343 that will do a better job of exclusions: e.g. if `variant` is selected, the `kind` prop is disabled, among others.

Plus all the presets in `variants_text.js` will be updated to use `variant` instead of current `kind+size`. 

My comments below are purely documentation. Approving for the moment so you merge after doing those.

PR #1340, opened 2026-06-24
chore: NO-JIRA enforce 7-day minimum release age and migrate to pnpm 11

braddialpad:
## Summary

Mirrors the same supply-chain hardening done in [web-clients#1907](https://github.com/dialpad/web-clients/pull/1907).

- Enforces a 7-day minimum release age on all installed dependencies — freshly published (potentially compromised) versions will be blocked until they have aged out of the quarantine window
- All `@dialpad/*` packages are excluded so first-party releases install immediately
- Upgrades pnpm from `9.9.0` to `11.9.0` (required for `minimumReleaseAge` support)

## Changes

**`pnpm-workspace.yaml`**
- Migrate `overrides`, `peerDependencyRules`, `patchedDependencies` from `package.json` (pnpm 11 no longer reads the `pnpm` field there)
- Add `minimumReleaseAge: 10080` (7 days in minutes), `minimumReleaseAgeStrict: true`, `minimumReleaseAgeIgnoreMissingTime: false`
- Add `minimumReleaseAgeExclude: ['@dialpad/*']`
- Add `engineStrict: true`
- Add `allowBuilds` for all existing lifecycle script packages to retain parity with prior behaviour

**`package.json`**
- Bump `packageManager` to `pnpm@11.9.0`
- Add `engines.pnpm: '>=11.9.0'`
- Add `@eslint/js` as a direct devDependency (pnpm 11 strict mode requires explicit declaration for packages imported directly)
- Remove `pnpm` field (moved to `pnpm-workspace.yaml`)

**Docs**
- Update `README.md` and `getting-started/index.md` to recommend `corepack enable` over `npm install -g pnpm`
- Update stale Node `v16.x.x` requirement to use `.nvmrc` + `nvm install`

## Notes

- CI requires no changes — `pnpm/action-setup@v3` reads `packageManager` automatically, and Node 24 satisfies pnpm 11's `>=22.13` requirement
- `allowBuilds` entries retain existing behaviour and can be tightened in a follow-up

## JIRA

https://dialpad.atlassian.net/browse/DLT-194119

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,2 @@
 publish-branch=production
 save-prefix=""
-@dialpad:registry=https://npm.pkg.github.com
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,3 @@
 publish-branch=production
 save-prefix=""
+@dialpad:registry=https://npm.pkg.github.com
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -7,3 +7,56 @@ packages:
   - 'packages/language-server/*'
   # yeoman generator
   - 'generator-dialtone'
+
+engineStrict: true
+
+peerDependencyRules:
+  allowedVersions:
+    mem-fs: '^4.0.0'
+    postcss: '^7.0.39 || ^8.3.3'
+    react: '^18.3.1'
+    typescript: '^5.2'
+    vite: '^8.0.0'
+    vue: '^3.2'
+    vuepress: '^2.0.0-rc.24'
+
+overrides:
+  playwright-core: '>=1.60.0'
+  '@typescript-eslint/utils': '^8.44.1'
+  braces: '>=3.0.3'
+  cross-spawn: '>=7.0.5'
+  nanoid: '>=3.3.8'
+  sass: '^1.70.0'
+  sugarss: '^5.0.0'
+  semver-regex: '3.1.4'
+  stylelint: '^16.25.0'
+  tar-fs: '>=2.1.3'
+  test-exclude: '^7.0.1'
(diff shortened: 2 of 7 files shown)


.npmrc
```diff
@@ -1,2 +1,3 @@
 publish-branch=production
 save-prefix=""
+@dialpad:registry=https://npm.pkg.github.com
```
jawrey: hopefully everyone has their PATs set up!

pnpm-workspace.yaml
```diff
@@ -7,3 +7,56 @@ packages:
...
+  esbuild: true
+  fsevents: true
+  keytar: true
+  nodent-runtime: true
+  nx: true
+  spawn-sync: true
+  style-dictionary: true
+  unrs-resolver: true
+  vue-demi: true
+  yo: true
```
jawrey, line 62: I've got the same allow list set up in web-clients and I do think its probably better than just blanket allowing scripts to be ran from every package... but I also worry that we're just gonna end up blindly adding everything anyways.

Idk if anyone is really inspecting the scripts of every package they download, I'm certainly not lol.  At the very least its still nice to have a known list of packages that are running shit though, so thats good

jawrey (approved):
woot woot

ninarepetto (approved):

.npmrc
braddialpad: Actually.... I shouldn't set that since Dialtone is a public repo

pnpm-workspace.yaml
braddialpad: yeah.. unfortunately I don't really think theres any better alternative

PR #1342, opened 2026-06-25
fix(dialtone-css): DLT-3492 DLT-3493 fix codemod incorrectly mapping large spacing values to wrong tokens

braddialpad:
- [x] Bug fix

## :book: Jira Ticket
[DLT-3492](https://dialpad.atlassian.net/browse/DLT-3492) | [DLT-3493](https://dialpad.atlassian.net/browse/DLT-3493)

### `utility-class-to-token-stops` — silent spacing regression on large pixel values

The codemod was silently producing incorrect output for margin, padding, and position classes with pixel values ≥ 96px (e.g. `d-pt96`, `d-mt128`).

**Root cause:** The codemod mapped these via `SPACING_LAYOUT_MAP` to token-stop names like `d-pt-150`. However, token-stop padding/margin/position classes (`d-pt-{stop}`, `d-mt-{stop}`, `d-t-{stop}`) are generated exclusively from `SPACING_STOPS` using **spacing tokens** — not layout tokens. So `d-pt-150` resolves to `--dt-spacing-150` (~12px), not `--dt-layout-150` (96px). This is a significant and silent spacing regression.

**Fix:** Remove `SPACING_LAYOUT_MAP` from the margin, padding, and position expressions. Classes like `d-pt96` now pass through unchanged, consistent with how unrecognized sizing values like `d-h72` are handled — left for manual review. `SPACING_LAYOUT_MAP` is still correctly applied to sizing classes (`d-h`, `d-w`, etc.) where layout-token-backed classes do exist.

---

### `stack-gap-to-spacing` — double-transformation in convergence loop

The `stack-gap-to-spacing` codemod was double-transforming gap values. `gap="500"` would become `gap="25"` instead of the correct `gap="200"`.

**Root cause:** `runConfigMigration` in the master migration script uses a convergence loop (`while (changed)`) that re-runs all expressions until a full pass produces no changes. This was added for configs like `size-to-layout`, where a `border:` property with two `var(--dt-size-*)` tokens can only match the second token after the first pass reshapes the line prefix — so multi-pass is genuinely needed there.

For `stack-gap-to-spacing`, however, the GAP_MAP output values overlap with the input set: `500 → 200`, and `200` is itself a valid old input that maps to `25`. So the loop would transform `gap="500"` → `gap="200"` on pass one, then pick it up again and transform `gap="200"` → `gap="25"` on pass two. The same cascading problem exists at other stops: `600 → 400 → 100`, `650 → 600 → 400 → 100`, etc.

**Fix:** Add `singlePass: true` to `stack-gap-to-spacing.mjs` (with a comment explaining the invariant), and change the convergence loop in `runConfigMigration` from `while (changed)` to `do { … } while (changed && !configData.singlePass)`. Configs that need multi-pass convergence are unaffected.

## :pencil: Checklist

For all PRs:

For all CSS changes:

[DLT-3492]: https://dialpad.atlassian.net/browse/DLT-3492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DLT-3493]: https://dialpad.atlassian.net/browse/DLT-3493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/dialtone-css/lib/build/js/dialtone_migrate/index.mjs
+++ b/packages/dialtone-css/lib/build/js/dialtone_migrate/index.mjs
@@ -743,8 +743,8 @@ async function runConfigMigration (migration, opts) {
   // Loop until convergence: some config regexes only match one token per
   // property declaration per pass (e.g. border-width with two var(--dt-size-*)).
   for (const entry of matched) {
-    let changed = true;
-    while (changed) {
+    let changed;
+    do {
       changed = false;
       for (const expr of configData.expressions) {
         const before = entry.data;
@@ -758,7 +758,7 @@ async function runConfigMigration (migration, opts) {
           changed = true;
         }
       }
-    }
+    } while (changed && !configData.singlePass);
     if (entry.matches > 0) {
       await fs.writeFile(entry.file, entry.data, 'utf8');
       const shortname = path.relative(opts.cwd, entry.file);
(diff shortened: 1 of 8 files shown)


braddialpad:
@coderabbitai re-run the review, the base branch was incorrect

iropolo (approved):

PR #1349, opened 2026-07-02
feat(migration-helper): v0.3.57 — in-place token substitution + tooltip fix

belumontoya:
## What

Lands the `packages/dialtone-migration-helper` Chrome extension into the monorepo (initial commit — the package existed on disk but was untracked).

### Replace mode (was broken)
The previous implementation crammed the full stable framework CSS into the first `dialtone-css-*` tag and blanked the other two, destroying Next's runtime theming API slots and leaving tokens undefined. Rebuilt as **in-place value substitution**:
- Each tag keeps its structure, mode blocks (`[data-dt-mode]`), `@layer` wrappers, and token names byte-for-byte
- Only `--dt-*` declaration values are swapped to the last stable release
- Mechanically renamed families resolved via reverse-rename tables in the new `token-renames.js` (`spacing←space`, `layout←size`, `positive←success`, border/radius)
- `setMode()`/`setBrand()` keep working natively — no MutationObserver hack needed
- Disable restores every tag to byte-identical original content

### Token inspector tooltip
Utility class section now shows custom-property declarations one per line (capped at 6 with `…+N more`) instead of dumping the raw `cssText` blob.

### ESLint
Added browser + Chrome extension globals override in `eslint.config.js` for the migration helper package.

## How to test
1. Load `packages/dialtone-migration-helper` unpacked in `chrome://extensions`
2. Open a live dialtone-next page
3. Toggle **Replace stable tokens** — token values flip in place, page layout intact
4. Toggle off — page restores exactly
5. Toggle **Enable inspector** — hover an element with `d-*` classes, utility declarations show one per line

```sh
cd packages/dialtone-migration-helper && node e2e.mjs  # 9 tests
```


francisrupert:
close without merge? revisit?

PR #1373, opened 2026-07-17
chore(android): NO-JIRA upgrade Gradle/AGP/Kotlin/Compose

braddialpad:
## Summary
- Gradle 7.5.1 could not configure the build under newer JDKs (e.g. JDK 23), breaking the VS Code Gradle integration for dialtone-tokens and dialtone-icons/android. The android-dialtone app that consumes both published artifacts is also already on a newer Kotlin/AGP/Compose stack than either package, so this brings both fully in line with that consumer rather than just unblocking local tooling.

- Bumps Gradle wrapper 7.5.1 -> 9.0.0, AGP 7.1.1 -> 8.13.2, Kotlin 1.7.20 -> 2.3.0, and Compose BOM 2022.10.00 -> 2026.01.01 for both Android Gradle projects, matching the versions pinned in android-dialtone's buildSrc/Dependencies.kt exactly.

- Adds the now-required namespace in each android {} block (AGP 8 no longer derives it from the manifest package attribute) and removes the redundant package= attribute from both AndroidManifest.xml files.

- Raises compileSdk 31 -> 36, targetSdk 31 -> 35, and minSdk 21 -> 28 to match android-dialtone's config. The minSdk raise drops support for API 21-27 consumers.

- Replaces composeOptions { kotlinCompilerExtensionVersion } with the org.jetbrains.kotlin.plugin.compose plugin in both build.gradle files. As of Kotlin 2.0 the Compose compiler moved into the Kotlin repo and ships in lockstep with the Kotlin version, so the old composeOptions mechanism no longer applies.

- Switches Maven publishing to AGP's singleVariant('release') { withSourcesJar() } DSL, removing the hand-rolled androidSourcesJar task whose output wasn't declared as an input to generateMetadataFileForReleasePublication - Gradle's stricter task validation was rejecting it, breaking ./gradlew publish for both packages.

- Fixes a Gradle 9 breaking change surfaced by the wrapper bump - WriteProperties.outputFile was removed in favor of destinationFile

- Keeps compiled bytecode targeting Java 8 (explicit compileOptions/kotlinOptions) to match prior default behavior and avoid breaking downstream AAR consumers on older toolchains.

- Bumps kotlin.daemon.jvmargs heap in both packages' gradle.properties - the newer Kotlin compiler OOM'd at the previous default heap.

- Extracts a shared Gradle version catalog (packages/gradle/libs.versions.toml) referenced by both projects' settings.gradle via dependencyResolutionManagement.versionCatalogs, so AGP/Kotlin/Compose BOM/SDK versions are defined once instead of duplicated across dialtone-tokens/build.gradle and dialtone-icons/android/build.gradle.

No impact to the JS/Vue/Nx build pipeline - these Gradle projects are isolated to Android platform artifact generation. CI runs on JDK 17, which remains compatible with Gradle 9 and AGP 8.13.

## Test plan
- [x] `./gradlew clean assembleRelease` succeeds for `dialtone-tokens` and `dialtone-icons/android` under JDK 17/18
- [x] Same builds verified under JDK 23 (the version that originally triggered the VS Code Gradle error)
- [x] Confirmed no change to published namespace/package or resource content vs. prior build output


braddialpad:
mostly for @johnp-dialpad to review

iropolo (approved):
Im approving it as codeowner to unblock the merge.
Waiting for @johnp-dialpad to review.
Thanks,

johnp-dialpad:
Hi @braddialpad thanks for doing this. There are a few other things that would be good to update now e.g. compose bom version which require more changes and bumping some of these versions even higher. Can I push those to this PR when I'm done?

braddialpad:
Yeah please do!

johnp-dialpad (approved):

PR #1312, opened 2026-06-04
refactor(popover): DLT-3455 trap focus with v-dt-focustrap directive

paulojreis-dialpad:
## Obligatory GIF (super important!)

Obligatory GIF

## :hammer_and_wrench: Type Of Change

- [ ] Fix
- [ ] Feature
- [ ] Performance Improvement
- [x] Refactor

## :book: Jira Ticket

[DLT-3455](https://dialpad.atlassian.net/browse/DLT-3455)

## :book: Description

Migrates `DtPopover`'s Tab focus-trapping from the legacy `common/mixins/modal.js` mixin to the `v-dt-focustrap` directive — a **trap-only** swap. The directive owns *only* the Tab trap; Popover keeps its existing initial-focus and anchor-restoration (and the Modal mixin, still used by those). Behaviour is preserved. Part of [DLT-3452](https://dialpad.atlassian.net/browse/DLT-3452).

- **`Popover.vue`** — add `v-dt-focustrap="{ active: modal && isOpen, initialFocus: false, restoreFocus: false }"` on the `<dt-lazy-show>` content (the exact element the old trap scoped to — Tippy relocates this node but the directive's listener moves with it). Remove the Tab branch from `onKeydown` (Escape handling + the `keydown` emit are kept). The Modal mixin stays — `focusInitialElement` / `focusFirstElementIfNeeded` / `_getFocusableElements` still drive Popover's own initial focus and restoration.
- **`Popover.test.js`** — register `DtFocustrapDirective` in the mount `global.plugins`; add a "Focus trapping" suite (modal wraps both directions; non-modal doesn't trap).
- **Dropdown** (the other half of DLT-3455) needs **no standalone change**: `Dropdown.vue` has no trap of its own — it wraps `<dt-popover :modal="modal" role="menu">` with `modal` defaulting `true`, so its trapping is fully delegated to Popover. This migration already covers it; a modal Dropdown traps via the directive on its Popover.

## :bulb: Context

Focus trapping was scattered across components via the Modal mixin; DLT-3452 consolidates it on one directive (`v-dt-focustrap`, shipped in #1195).

[...]

PR #1392, opened 2026-08-13
fix(popover): NO-JIRA don't hide dialog from assistive tech while it still holds focus on close

belumontoya:

- [x] Fix

## :book: Jira Ticket

NO-JIRA

## :book: Description

`DtPopover`'s dialog binds `aria-hidden` directly to `!isOpen`, so it flips to `true` the instant the popover starts closing. For modal popovers (used by `DtSelectMenu`/dropdowns), focus isn't restored to the anchor until `onLeaveTransitionComplete` runs *after* the leave transition finishes — so for the duration of that transition, the dialog is `aria-hidden="true"` while a descendant (e.g. the selected list item) still holds focus. Chrome blocks this and logs:

```
Blocked aria-hidden on an element because its descendant retained focus...
```

This fix decouples `aria-hidden` from `isOpen`. A new `isDialogAriaHidden` data property starts `true` and:
- is set `false` immediately when the popover opens (no risk there),
- is set `true` only inside `onLeaveTransitionComplete`, after focus has already been moved back to the anchor for modal popovers (or the leave transition has otherwise finished for non-modal ones).

`aria-hidden` in the template now binds to `isDialogAriaHidden` instead of `!isOpen`.

## :bulb: Context

Introduced by #1312 (DLT-3455), which added the `v-dt-focustrap` directive to `DtPopover` with `restoreFocus: false` for modal popovers, relying on `onLeaveTransitionComplete` for focus restoration instead — but nothing delayed `aria-hidden` to match. Confirmed pre-Next Dialtone (9.x) didn't have this issue since it had no focus-trapping at all, so nothing retained focus inside the dialog for aria-hidden to race against.

## :pencil: Checklist

Added two tests under `Focus trapping > When modal and open`:
- `closePopover()` alone must not mark the dialog `aria-hidden`.
- The dialog is marked `aria-hidden` only after `onLeaveTransitionComplete` has restored focus to the anchor.

Verified both fail against the pre-fix component (confirmed via `git stash`) and the full `Popover.test.js` suite passes (50 passed, 2 pre-existing skips) with the fix applied.


braddialpad (approved):
Yep was exactly the issue I expected, thanks!

francisrupert (approved):

PR #1394, opened 2026-08-13
fix(notice, toast): DP-203134 render message paragraph only when there is a message

belumontoya:

- [x] Fix

## :book: Jira Ticket

https://dialpad.atlassian.net/browse/DP-203134

## :book: Description

A title-only toast (e.g. "Copied to clipboard") rendered an empty `<p class="d-notice__message">`, so the close icon centred against the title *plus* that invisible line instead of the title alone.

`DtNoticeContent` already guarded the message with `hasSlotContent($slots.default)`, but that guard could never be false via `DtToast`: both `DtToast` and `ToastLayoutDefault` forwarded the default slot **with a `{{ message }}` fallback**, so the slot always carried a vnode even when there was no message.

The header never had this problem, because `headerText` reaches `DtNoticeContent` as a **prop** and its fallback lives *inside* `DtNoticeContent` — so an absent value leaves the header slot genuinely empty and `v-if="headerText || hasSlotContent($slots.header)"` correctly evaluates false.

This gives `message` the same shape as `headerText`:

- `DtNoticeContent` takes a `message` prop, hosts the `{{ message }}` fallback, and guards on `message || hasSlotContent($slots.default)`
- `ToastLayoutDefault` passes `:message` through and forwards the default slot bare (`<slot />`)
- `DtToast` forwards the default slot bare, since `message` now reaches `DtNoticeContent` as a prop

`ToastLayoutAlternate` is deliberately untouched — it renders the slot in its own `div` rather than through `DtNoticeContent`, so it never produced the empty paragraph.

## :bulb: Context

Follow-up to #1390, which added the `hasSlotContent` guard. That guard is correct and works when `DtNoticeContent` is used directly, but it could not fire through `DtToast` because of the fallbacks described above — so the reported case was still broken. Verified against the released `10.0.0-next.13` in a running app before writing this fix, by A/B-ing the two paths: removing `:header-text` from the consumer hid the title paragraph, while removing `:message` did not hide the message paragraph.

## :pencil: Checklist

Tests added to `Toast.test.js`:
- `headerText` only, no message and no default slot → no message paragraph, title still rendered
- `message` provided → message paragraph rendered with the message text

Both fail before this change and pass after. Full `dialtone-vue` suite green: **100 files, 3243 passed, 2 pre-existing skips, 0 failures.** Lint clean on all changed files.


braddialpad (approved):
Thanks

PR #1408, opened 2026-09-01
NO-JIRA: fix links inside rich text editor message open new tab

iropolo:
Alternative fix for https://github.com/dialpad/firespotter/pull/81693

## What I added to DtRichTextEditor
A new link-click event that lets consumers intercept link clicks (the root-cause fix — the target="_blank" new-tab behavior originates in this component's TipTap Link extension).

## Three changes in RichTextEditor.vue:

Imports — getMarkRange from @tiptap/core, Plugin/PluginKey from @tiptap/pm/state (verified resolvable in this repo's versions).

A ProseMirror click plugin on the built-in Link extension (addProseMirrorPlugins) — on clicking a link mark, it emits link-click with { href, text, event }. It returns event.defaultPrevented, so if the consumer calls event.preventDefault(), ProseMirror treats the click as handled and the browser does not follow the anchor's target="_blank". Crucially, it preserves any parent plugins ([...this.parent?.(), linkClickPlugin]).

Event bridge + declaration — editor.on('link-click') → this.$emit('link-click'), plus a documented entry in the emits array (mirroring the existing phone-click convention).

## Tests added (RichTextEditor.test.js):

emits link-click with the href/text/event when a link is clicked

does NOT emit when the click isn't on a link mark

reports the click as handled when the consumer prevents default

## How ubervoice will consume it (once this ships)

After a Dialtone release + version bump, the ubervoice fix simplifies to just listening to the new event on RichMediaInline:

```
<dt-rich-text-editor ... @link-click="onLinkClick" />

onLinkClick ({ href, event }) {
  const appPrefix = `${config.domain}/app/`;
  if (href?.startsWith(appPrefix)) {
    event.preventDefault();
    this.$navigate(href.substring(appPrefix.length));
  }
}
```
— replacing the delegated DOM listener (my 8d291b588d7 workaround) with the proper component API.

### DIFF
--- a/packages/dialtone-vue/components/RichTextEditor/RichTextEditor.vue
+++ b/packages/dialtone-vue/components/RichTextEditor/RichTextEditor.vue
@@ -56,7 +56,8 @@
 /* eslint-disable max-lines */
 import { Editor, EditorContent } from '@tiptap/vue-3';
 import { BubbleMenu } from '@tiptap/vue-3/menus';
-import { Extension } from '@tiptap/core';
+import { Extension, getMarkRange } from '@tiptap/core';
+import { Plugin, PluginKey } from '@tiptap/pm/state';
 import { DtButton } from '../Button';
 import { DtStack } from '../Stack';
 import Blockquote from '@tiptap/extension-blockquote';
@@ -617,6 +618,18 @@ export default {
      * @type {Object}
      */
     'phone-click',
+
+    /**
+     * Event fired when a link is clicked.
+     * Payload: { href: string, text: string, event: MouseEvent } — the link's
+     * href, its text content, and the originating click event. Call
+     * event.preventDefault() to suppress the default navigation (e.g. to route
+     * same-origin URLs in-app instead of opening a new tab).
+     * Requires the `link` prop to be set.
+     * @event link-click
+     * @type {Object}
+     */
+    'link-click',
   ],
 
   data () {
@@ -722,6 +735,30 @@ export default {
                 return true;
               },
             };
+          },
+          addProseMirrorPlugins () {
+            const markType = this.type;
+            // Emit `link-click` so consumers can intercept links (e.g. to
+            // navigate same-origin URLs in-app instead of opening a new tab).
+            // The listener may call event.preventDefault() to suppress the
+            // default navigation; otherwise the anchor behaves normally.
+            const linkClickPlugin = new Plugin({
+              key: new PluginKey('linkClick'),
+              props: {
+                handleClick (view, pos, event) {
+                  const { state } = view;
+                  const $pos = state.doc.resolve(pos);
+                  const linkMark = $pos.marks().find(m => m.type === markType);
+                  if (!linkMark) return false;
+                  const range = getMarkRange($pos, markType);
+                  const text = range ? state.doc.textBetween(range.from, range.to) : '';
+                  self.editor.emit('link-click', { href: linkMark.attrs.href, text, event });
+                  return event.defaultPrevented;
+                },
+              },
+            });
+
+            return [...(this.parent?.() ?? []), linkClickPlugin];
           },
         }).configure({
           HTMLAttributes: {
@@ -1324,6 +1361,11 @@ export default {
       // Phone number link is clicked — only emitted by the LinkPhoneNumbers extension
       this.editor.on('phone-click', (phoneData) => {
         this.$emit('phone-click', phoneData);
+      });
+
+      // A link is clicked — only emitted when the built-in `link` extension is active
+      this.editor.on('link-click', (linkData) => {
+        this.$emit('link-click', linkData);
       });
     },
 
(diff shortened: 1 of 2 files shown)


packages/dialtone-vue/components/RichTextEditor/RichTextEditor.vue
```diff
@@ -723,6 +736,30 @@ export default {
...
+                  const range = getMarkRange($pos, markType);
+                  const text = range ? state.doc.textBetween(range.from, range.to) : '';
+                  self.editor.emit('link-click', { href: linkMark.attrs.href, text, event });
+                  return event.defaultPrevented;
+                },
+              },
+            });
+
+            return [...(this.parent?.() ?? []), linkClickPlugin];
+          },
```
braddialpad: Can we extend the TipTapLink extension instead of doing this via addProseMirror plugins? Might be a lot cleaner

https://tiptap.dev/docs/editor/extensions/marks/link

braddialpad (changes requested):
This is the right way to do it, allow the consumer to do what they need with a link event instead of handling it internally.

1. Is it possible to do this in TipTap extension instead of falling back to prosemirror?
2. Can we make this opt in via a prop so it isn't a breaking change? `defaultLink = true` or something like that

iropolo:
> This is the right way to do it, allow the consumer to do what they need with a link event instead of handling it internally.
> 
> 1. Is it possible to do this in TipTap extension instead of falling back to prosemirror?
> 2. Can we make this opt in via a prop so it isn't a breaking change? `defaultLink = true` or something like that

Okay, this is the update:

## AI explanation: 

-------
1. TipTap extension vs raw ProseMirror plugin
I checked the installed @tiptap/extension-link source. This version has no onClick option/callback — its built-in clickHandler only does window.open when openOnClick: true, with no consumer hook. So there's no pure "extension option" to use.

But the change is now idiomatic TipTap: it lives in the extension's addProseMirrorPlugins override (the same mechanism TipTap itself uses to register clickHandler), and I rewrote the handler to mirror TipTap's own approach — read the clicked anchor via event.target.closest('a') and its href, instead of the manual getMarkRange/resolve/textBetween gymnastics. That let me drop the getMarkRange import and the body is much cleaner/more robust (works for a click anywhere in the anchor, not just text positions).

So: it's not a built-in option (this version doesn't offer one), but it's now the clean, extension-native way.

2. Opt-in prop (no breaking change)
Added emitLinkClick prop (default false). When off, the plugin isn't registered at all — behavior is byte-identical to before. When a consumer sets :emit-link-click="true" and listens to @link-click, they get { href, text, event } and can preventDefault() to route in-app. Named it emitLinkClick rather than defaultLink since it precisely describes what it does (defaultLink was ambiguous).

Final state (branch fix-rich-text-new-tab)
RichTextEditor.vue: new emitLinkClick prop, link-click in emits, gated + anchor-based addProseMirrorPlugins, event bridge, getMarkRange import removed.


-------
## Not AI: 

I update https://github.com/dialpad/firespotter/pull/81693 with the necessary change to apply the fix.
I need to deploy this version of Dialtone and then I will update it in package.json in the same PR.

braddialpad:
I pushed a small change to move this from `addProseMirrorPlugins` to `handleClick`. Works the same but the code is a bit cleaner and more the "tiptap" way.

iropolo:
> I pushed a small change to move this from `addProseMirrorPlugins` to `handleClick`. Works the same but the code is a bit cleaner and more the "tiptap" way.

Okay one sec Im testing it with pnpm pack on firespotter and doesn't work.

braddialpad:
> Opt-in prop (no breaking change)
Added emitLinkClick prop (default false). When off, the plugin isn't registered at all — behavior is byte-identical to before. When a consumer sets :emit-link-click="true" and listens to [tiffany (@link-click)](https://github.com/link-click), they get { href, text, event } and can preventDefault() to route in-app. Named it emitLinkClick rather than defaultLink since it precisely describes what it does (defaultLink was ambiguous).

Hmmm I am thinking we actually don't need this prop now.. Originally I thought the new event would remove the default link behaviour, but that's not the case. The default behaviour remains and it can be turned off via `event.preventDefault()`

braddialpad:
> Okay one sec Im testing it with pnpm pack on firespotter and doesn't work.

Is it because of the new prop?

iropolo:
Okay now I have it working :


https://github.com/user-attachments/assets/122b90a5-a16c-41bd-890d-81bce1f616b6

braddialpad (approved):
Nice thanks, this looks good.

iropolo:
> Nice thanks, this looks good.

I will continue with the deploy of this and update https://github.com/dialpad/firespotter/pull/81693 with the latest version. 
I will double check it if it works as expected and send it 👍🏼

iropolo:
@coderabbitai review

iropolo:
@coderabbitai full review

PR #1419, opened 2026-09-11
fix(rich-text-editor): DLT-3610 autolink URLs followed by an unmatched closing bracket

braddialpad:
## Summary
- Typing a URL immediately followed by a single unmatched closing bracket (e.g. "(like https://dialpad.design/) and ...") silently fails to autolink at all, even though `linkifyjs` correctly identifies the URL boundary on its own.
- Root cause is in `@tiptap/extension-link`'s `isValidLinkStructure()`: it only accepts a single link token, or a link token symmetrically wrapped in matching `()`/`[]`. A lone trailing bracket (a 2-token split of link + leftover punctuation) fails both checks, so the whole match is discarded.
- `isValidLinkStructure()` and the `autolink()` plugin that calls it aren't exported by `@tiptap/extension-link`, and a `pnpm patch` on the dependency only applies within this workspace — the published `@dialpad/dialtone-vue` package ships only `dist`, declares the unpatched dependency in `packages/dialtone-vue/package.json`, and externalizes `@tiptap/*` in `packages/dialtone-vue/vite.config.js`, so the patch never reaches consumers.
- Instead, `packages/dialtone-vue/components/RichTextEditor/Extensions/Link/Autolink.js` is a Dialtone-owned reimplementation of upstream's `autolink()` plugin (built only from public `@tiptap/core`, `@tiptap/pm/state`, and `linkifyjs` APIs) with the `isValidLinkStructure()` fix applied. `RichTextEditor.vue` disables the built-in extension's `autolink` option and pushes this fixed plugin instead via `addProseMirrorPlugins()`, preserving the existing customLink conflict-avoidance behavior.
- This ships as real source in the built bundle, so it works identically for consumers of the published package. The `@tiptap/extension-link` patch and its `patchedDependencies` entry are removed; the unrelated `@tiptap/vue-3` patch is untouched.

Reported by Josh Hynes and Francis Rupert in #launchathon-rich-text-editor.

Tracked in [DLT-3610](https://dialpad.atlassian.net/browse/DLT-3610). Filing an upstream issue against `ueberdosis/tiptap` separately; this Dialtone-owned plugin can be dropped in favor of `autolink: true` once that's fixed and the dependency is bumped.

## Test plan
- [x] Added `Autolink.test.js` covering `isValidLinkStructure()` against real `linkifyjs` tokens: the bracket-trailing/leading cases, the pre-existing symmetric-paren case, and the upstream safety case (`example.com1`) which still correctly does not autolink.
- [x] Verified with real keystrokes in a real Chromium browser (not jsdom): Josh's exact sentence now correctly linkifies, symmetric-paren case still works, and the safety case still correctly does not autolink.
- [x] Full `dialtone-vue` test suite passes (500 tests).
- [x] `dialtone-vue` lint passes with 0 errors.

[DLT-3610]: https://dialpad.atlassian.net/browse/DLT-3610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


braddialpad:
Not sure if I want to do this.. moving back to draft for now.

PR #1428, opened 2026-09-16
fix(ci): DLT-2979 strip quotes from blog post heading in send-blog-communications

braddialpad:
## Summary
- Fixes send-blog-communications.yml: the workflow ran "successfully" but silently failed to send an email or SMS for the v10 announcement post.
- Root cause: the blog post's frontmatter quotes its heading (heading: "Introducing Dialtone 10"). The sed extraction only stripped the "heading: " prefix, leaving the literal quotes in BLOG_POST_HEADING. That value was then substituted directly into the run: block via ${{ env.BLOG_POST_HEADING }} — a raw GitHub Actions expression, not a shell-safe reference — so the embedded quotes broke bash's quoting and shifted jq's arguments. jq failed to compile ("Dialtone/0 is not defined"), curl posted an empty/malformed body, and the Cloud Function correctly replied "Error: No recipients defined." The step still showed green because the pipeline had no pipefail, so only curl's exit code was checked.
- This would recur for every future post with a quoted frontmatter value (i.e. all of them, going by the current posts directory).

## Changes
- Strip surrounding quotes when extracting BLOG_POST_HEADING from frontmatter.
- Reference BLOG_POST_HEADING / POST_URL / TO_MAIL / API_URL via shell env vars ($VAR) inside the send-email and send-sms run: blocks instead of raw ${{ env.* }} interpolation, so embedded shell metacharacters can't break the command (matches the anti-pattern already called out in .claude/rules/github-workflows.md for PR titles).
- Add set -euo pipefail and curl --fail-with-body so a jq/curl failure actually fails the step instead of reporting false success.

## Test plan
- [ ] Confirm `echo 'heading: "Introducing Dialtone 10"' | sed -n 's/^heading: *"\{0,1\}\(.*[^"]\)"\{0,1\} *$/\1/p'` outputs `Introducing Dialtone 10` (no quotes)
- [ ] Re-run send-blog-communications.yml via workflow_dispatch with a test_email and confirm the email actually arrives with a clean subject line
- [ ] Confirm the workflow now fails loudly (not silently) if jq/curl fail


iropolo (approved):

braddialpad:
Going to close this. The real problem was incorrect syntax in the blog post. We COULD make these changes to support the invalid quotes but the solution here is a bit messy, and I think the correct solution is to totally rework our blog system

PR #1434, opened 2026-09-21
fix(popover): DLT-3607 return a rect fallback and guard tippy init against a null anchor

braddialpad:
## Summary
- `Popover.vue`'s `getReferenceClientRect` returned `undefined` when `anchorEl` was null. That callback stays wired into the tippy/popper instance for its whole lifetime, so popper's own debounced `forceUpdate()` (`@popperjs/core`) can invoke it well after creation — dereferencing `state.rects.reference.x` on `undefined` and throwing an unhandled `TypeError: Cannot read properties of undefined (reading 'x')`.
- The `isOpen` watcher could also call `initTippyInstance()` before `mounted()` had resolved an `anchorEl` (e.g. an immediate `open` prop watcher firing pre-mount), creating a tippy instance with no real reference element.
- `DtDropdown` composes `DtPopover` internally, so this fixes the crash reported against both components in [DLT-3607](https://dialpad.atlassian.net/browse/DLT-3607).

## Changes
- `getReferenceClientRect` now falls back to a zero rect instead of `undefined`.
- `isOpen` watcher now bails out of tippy initialization when `anchorEl` isn't resolved yet.
- Added two regression tests in `Popover.test.js`, each independently verified to fail when its corresponding fix is reverted.

## Test plan
- [x] `pnpm nx run dialtone-vue:test -- popover dropdown tooltip` — 175 passed, 2 skipped
- [x] Verified each new test fails when its fix is reverted (confirms they exercise the bug)
- [x] `npx eslint` on changed files — 0 errors, only pre-existing warnings (one complexity warning ticked 20→21 due to the added fallback)

[DLT-3607]: https://dialpad.atlassian.net/browse/DLT-3607?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


braddialpad:
@coderabbitai approve

